PR #59, opened 2021-10-22
MySql driver w/test bed (requires starting docker container db)

slawlor:
Start the test db with ```docker compose up``` assuming docker is installed

The test bench will determine if the container is running via a system call to check the docker output of the running containers. See ```MySqlDatabase::test_guard()```


IYANGO6:
God bless the poor

PR #77, opened 2021-11-08
Downgrading mysql_async from 0.28 to 0.23.1

slawlor:
mysql_async needs to be 0.23.1 for other dependencies within FB infra so this eases the import pain. Also matched Tokio with the required version internally (0.2.1)

Additionally removed the dev dependency on artix_rt for testing (using tokio::test directly)


kevinlewi:
Awesome, thanks!

PR #78, opened 2021-11-09
Performance adjustments + In-memory caching in MySQL reads

slawlor:
Added basic temporal caching within the MySQL database implementation. Additionally added flamegraph and notes about performance & performance measurements


Jasleen1 (approved):
Thank you!

PR #80, opened 2021-11-16
Bug fixes and adjustments for cache management

slawlor:
Currently performance is a batch of 100,000 takes approximately 40 minutes on a relatively fresh tree. That is expected to degrade as the tree grows, since compression will be minimized.

Also found the race condition with locks which was causing us to deadlock on cache cleaning

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ mysql_async = "0.23.1"
 async-trait = "0.1"
 tokio = { version = "0.2", features = ["full"] }
 async-recursion = "0.3"
+log = { version = "0.4.8", features = ["kv_unstable"] }
 
 [dev-dependencies]
 criterion = "0.3"
--- a/poc/Cargo.toml
+++ b/poc/Cargo.toml
@@ -14,6 +14,7 @@ colored = "2.0.0"
 hex = "0.4.3"
 structopt = "0.3.13"
 rand = "0.8"
+log = { version = "0.4.8", features = ["kv_unstable"] }
 
 [dev-dependencies]
 serial_test = "*"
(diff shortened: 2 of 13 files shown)


Cargo.toml
```diff
@@ -24,6 +24,7 @@ mysql_async = "0.23.1"
 async-trait = "0.1"
 tokio = { version = "0.2", features = ["full"] }
 async-recursion = "0.3"
+log = { version = "0.4.8", features = ["kv_unstable"] }
```
kevinlewi, line 27: Is this something that can be included as a dev-dependency, or does it need to be a main dependency?

kevinlewi:
LGTM.

To clarify, this doesn't change anything functional and doesn't affect performance, right? Since the cache behavior isn't actually changing? Sorry if I missed the code which does make the change.

kevinlewi (approved):

Cargo.toml
slawlor: So this matches what's in FB Infra, it's a better logging infrastructure over "write to console always", especially as we might migrate to other external loggers (e.g. Scuba) in the future. But I checked the version & args used internally, and this is what's there, so we shouldn't have any import conflicts.

It's not a dev dependency since the log calls will be done in prod. In the POC I added a basic "ConsoleLogger" implementation so you can see what it does. It lets us manage log-levels

slawlor:
@kevinlewi Actually this does affect perf a bit, since we no longer deadlock after 30s :p Also I did some profiling and adjusted the cache so we only "clean" it ever 15s rather than on every read & write. This saved us like 2x timing on larger data sets, since "most" clean operations, especially in a tight loop, won't result in any items to delete, yet will induce a lock & full scan of the items in the cache.

I also removed serialization in the cache & in-mem transaction implementations, so that saved a decent amount

PR #81, opened 2021-11-19
Benchmark timing information and future support for batch retrieval of data from MySQL

slawlor:
1. Added a lot of timing & debug info to various parts of the program for doing benchmark calculations of code performance. Example output from POC app
```
======= Benchmark operation requested ======= 
Beginning PUBLISH benchmark of 100 users with 1 updates/user
INFO - Starting the verifiable directory host
INFO - Cache hit since last: 4874, cached size: 3566 items
INFO - Transaction writes: 9771, Transaction reads: 9885
INFO - MySQL writes: 2511, MySQL reads: 3366, Time spent in MySQL op: 11.482 s
PUBLISHED 100 records in 12.164723323 s
Benchmark output: Inserted 100 users with 1 updates/user
Execution time: 12164 ms
Time-per-user (avg): 121649 µs
Time-per-op (avg): 121649 µs
INFO - AKD host shutting down
```
2. Added ```batch_get``` support to the storage layer for future support of node retrieval in batches (rather than 1-by-1)

### DIFF
--- a/poc/src/directory_host.rs
+++ b/poc/src/directory_host.rs
@@ -13,6 +13,7 @@ use log::{debug, error, info};
 use std::marker::{Send, Sync};
 use tokio::sync::mpsc::*;
 use winter_crypto::Hasher;
+use tokio::time::Instant;
 
 pub(crate) struct Rpc(
     pub(crate) DirectoryCommand,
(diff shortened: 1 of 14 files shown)


kevinlewi (approved):
LGTM!

PR #82, opened 2021-11-23
Batch user information retrieval + disabling cache cleaning during a transaction

slawlor:
This PR contains more IO improvements where we disable cache-cleaning during a transaction. The logic behind this is that we're doing high IO work and will likely be re-reading a lot of nodes frequently (especially at the higher levels of the tree), that we shouldn't clear the cache until the transaction is completed.

+ Migrated to a pure v2 storage impl in-memory database.

+ Batch retrieval of value state version information so with a single db query we can retrieve all user versions in a publish operation. The large the batch size, the more performant this will be (i.e. for 100K users it's 1 op vs 100K ops).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ Cargo.lock
 
 /poc/src/.DS_Store
 /poc/target
+/poc/log.txt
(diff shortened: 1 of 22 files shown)


kevinlewi (approved):
LGTM!

PR #83, opened 2021-11-24
Added preliminary preloads to publish operations

Jasleen1:


src/append_only_zks.rs
```diff
@@ -130,7 +161,8 @@ impl Azks {
         append_only_usage: bool,
     ) -> Result<(), AkdError> {
         self.increment_epoch();
-
+        self.preload_nodes_for_insertion::<S, H>(storage, insertion_set.clone())
```
slawlor, line 164: Can we borrow the insertion set rather than cloning it? This set might be 100K+ items

src/history_tree_node.rs
```diff
@@ -141,6 +141,25 @@ impl HistoryTreeNode {
         }
     }
 
+    pub(crate) async fn batch_get_from_storage<S: V2Storage + Send + Sync>(
+        storage: &S,
+        keys: Vec<NodeKey>,
```
slawlor, line 146: I'll refactor this later, but maybe I'll borrow the keys as well rather than clone them. It'll require a change on the storage trait (todo for me :p )

src/append_only_zks.rs
```diff
@@ -120,6 +120,37 @@ impl Azks {
             .await
     }
 
+    async fn preload_nodes_for_insertion<S: V2Storage + Sync + Send, H: Hasher>(
+        &self,
+        storage: &S,
+        _insertion_set: Vec<(NodeLabel, H::Digest)>,
+    ) -> Result<(), AkdError> {
+        let preload_depth = 10;
+        println!("-------------- preload depth = {} -------", preload_depth);
```
slawlor, line 129: ```debug!``` or ```trace!``` so it'll log to the proper loggers. We should be avoiding println! since we might want to redirect output to another source (i.e. external log sink eventually)

src/append_only_zks.rs
```diff
@@ -120,6 +120,37 @@ impl Azks {
             .await
     }
 
+    async fn preload_nodes_for_insertion<S: V2Storage + Sync + Send, H: Hasher>(
+        &self,
+        storage: &S,
+        _insertion_set: Vec<(NodeLabel, H::Digest)>,
```
slawlor, line 126: Unused? For the next-optimization, we'll be doing a runtime breadth-first-search for prefetching (greedily), so I don't think we'd use the optimization set for some kind of preload.

slawlor:
Thanks so much for this! A few nits, but otherwise it LGTM. I'm excited to bench it 👍

src/append_only_zks.rs
Jasleen1: Oh, we meant for this function to do all the necessary preloads, this is a starter draft version :)

slawlor: In that case we should also borrow it and not clone it :) (since it too is the size of the publish op (100K+))

PR #84, opened 2021-11-24
Move to multi-row statements in MySQL module

slawlor:
Update the prepared statements used at the data-layer in batch retrieval operations to favor multi-row statements when possible.

Multi-row statements are much more performant (i.e. 100K inserts of usernames to temp-table was taking ~2minutes and is now 6 seconds).

Future: We should also process set_batch(...) with the same methodology, but there's a lot more args...

see
```rust
async fn internal_batch_set(...)
```

in ```src/storage/mysql/mod.rs```

Note: Also dropped most of the logging from ```trace!``` to ```debug!``` since Tokio seems to be using the same lib and blows up the trace logger with output. We shouldn't go any lower than debug, ever.


kevinlewi (approved):
Awesome!

PR #87, opened 2021-11-30
Multi-row insert performance with MySQL & storage layer cleanups

slawlor:
Changes: 
1. Remove unused storage-layer operations
2. Performance enhancements with batch-insert in MySQL
3. Db tuning POC command, where the multi-row parallel insert number is adjustable.
4. Parallelize serialization operations in batches with ```rayon``` packages (version matches internal infra version)


kevinlewi (approved):
Thanks!

PR #89, opened 2021-12-03
Reorganize the repository structure

slawlor:
In order to better manage dependencies and minimal importing, this PR breaks apart the repository into 2 library crates and 1 binary crate linked in a single workspace.

Crates in project:
1. The base "AKD" library in the ```akd``` folder.
2. The MySQL specific data-layer implementation for an AKD in the ```akd_mysql``` folder, becoming a new crate "akd_mysql"
3. The POC application ("akd_app") in the ```poc``` folder. 

By using a common workspace, we can simplify the commands needed to be run by cargo to interactively test with the POC app (i.e. we can run ```cargo run``` from the directory root without having to go into the ```poc``` folder). Since it's the only _binary_, cargo will always run the POC without need for additional arguments.

```akd_mysql``` and ```akd_app``` both reference ```akd``` in their ```Cargo.toml``` files by filepath reference, so they're always up to date. Additionally Rust will detect breaks across all crates without having to navigate to different folders/solutions/etc. ```cargo fmt``` and ```cargo clippy``` will run across all crates with a single call as well to make sure we're up-to date. All unit tests, in all crates will be run with a single ```cargo test``` call.

With this breakapart, we've removed ```mysql_async```, ```evmap```, ```lazy_static```, and ```rayon``` as dependencies of the base ```akd``` crate so external users import less code that may not be relevant. Anyone who wants a MySQL connection can simply import ```akd_mysql``` on top of ```akd``` and will have a full implementation of a MySQL database connector.

Also updated docs & added debugging launch.json for VSCode with use of lldb.

Lastly, removed V1Storage in favor of a single Storage trait (which is the renamed trait from V2Storage) since no one had yet implemented V1, we didn't need the backwards compatibility. This required rewriting the cached in-memory database, which is better anyways since now it doesn't rely on a shared static instance of anything. Meaning we can instantiate multiple cache'd databases in a single process safely.


Jasleen1 (approved):
Thank you for this! Looks good!

PR #90, opened 2021-12-03
Integration tests

slawlor:
Added integration tests against 2 backing database structures

1. MySQL
2. In-memory database

These tests test the full end-to-end functionality of the directory of AKD from 

1. Publish
2. Lookup proof & verify
3. History proof & verify
4. Audit proof & verify

MySQL tests however won't run in the CI without adding docker support, but they will be auto-skipped if the docker container is not running or can't be found


kevinlewi (approved):
Excellent!

PR #93, opened 2021-12-05
Update directory publish API to return epoch and hash

afterdusk:
Update the directory publish API to return an epoch and hash on success.

The POC app is consequently updated to display the result to the CLI.

## Test Run
```console
$ cargo run        
   Compiling akd_app v0.1.0 (/Users/evanau/Projects/akd/poc)
    Finished dev [unoptimized + debuginfo] target(s) in 6.38s
     Running `/Users/evanau/Projects/akd/target/debug/akd_app`
Please enter a command
> [00:00:00.032] (7000023d7000) INFO   Starting the verifiable directory host (directory_host:57)
PUBLISH test test
[00:00:10.662] (7000009b0000) INFO   Retrieved 1 previous user versions of 1 requested (directory:92)
[00:00:10.662] (7000009b0000) INFO   Starting database insertion (directory:143)
[00:00:10.764] (700001bcb000) INFO   Preload of tree (6 objects loaded), took 0.101717335 s (append_only_zks:195)
[00:00:10.954] (7000009b0000) INFO   Cache hit since last: 66, cached size: 17 items (timed_cache:52)
[00:00:10.954] (7000009b0000) INFO   Transaction writes: 0, Transaction reads: 0 (transaction:78)
[00:00:10.962] (7000009b0000) INFO   MySQL writes: 40, MySQL reads: 6, Time read: 0.132675114 s, Time write: 0.174001216 s
        Tree size: 9
        Node state count: 12
        Value state count: 3 (mysql:704)
Response: PUBLISHED 'test' = 'test' in 0.329670055 s (epoch: 2, root hash: 8074b1f261ce707fb3cb8927d5ba60fd98162addbb3592b7e50805683729d157)
```

### DIFF
--- a/akd/src/storage/types.rs
+++ b/akd/src/storage/types.rs
@@ -12,8 +12,6 @@ use crate::storage::Storable;
 use serde::{Deserialize, Serialize};
 use std::convert::TryInto;
 
-use winter_crypto::Hasher;
-
 /// Various elements that can be stored
 #[derive(PartialEq, Eq, Debug, Hash, Clone, Copy)]
 pub enum StorageType {
@@ -39,10 +37,6 @@ pub struct Values(pub String);
 /// State for a value at a given version for that key
 #[derive(Debug, Clone, Hash, Eq, PartialEq, Serialize, Deserialize)]
 pub struct ValueStateKey(pub String, pub u64);
-
-/// Root hash of the tree and its associated epoch
-#[derive(Debug, Clone, Hash, Eq, PartialEq, Serialize, Deserialize)]
-pub struct EpochHash<H: Hasher>(pub u64, pub H::Digest);
 
 /// The state of the value for a given key, starting at a particular epoch.
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Hash)]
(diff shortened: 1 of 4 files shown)


akd/src/storage/types.rs
```diff
@@ -38,6 +40,10 @@ pub struct Values(pub String);
 #[derive(Debug, Clone, Hash, Eq, PartialEq, Serialize, Deserialize)]
 pub struct ValueStateKey(pub String, pub u64);
 
+/// Root hash of the tree and its associated epoch
+#[derive(Debug, Clone, Hash, Eq, PartialEq, Serialize, Deserialize)]
+pub struct EpochHash<H: Hasher>(pub u64, pub H::Digest);
+
```
slawlor, line 46: Since ```EpochHash<H>``` isn't actually a "storage" type, can you move this to directory.rs? Just since the storage layer has no interaction with this type.

slawlor (approved):
2 tiny things, but otherwise it'll be good to go! Thanks!

akd/src/storage/types.rs
```diff
@@ -38,6 +40,10 @@ pub struct Values(pub String);
 #[derive(Debug, Clone, Hash, Eq, PartialEq, Serialize, Deserialize)]
 pub struct ValueStateKey(pub String, pub u64);
 
+/// Root hash of the tree and its associated epoch
```
slawlor, line 43: This is no longer used right? Can you clean revert this file, then it should be good to merge!

PR #94, opened 2021-12-07
Upgrade `mysql_async` and `tokio`

eozturk1:
This branch upgrades `mysql_async` to version 0.28.1 and `tokio` to 1.10.2. Outputs for the following have been tested:

Build:
`cargo build`

Formatting:
`cargo fmt`
`cargo clippy `

Test Plan:
`cargo test`

```
cd poc
cargo run
> publish test test
```


slawlor:
Hmm the integration tests are failing in the CI. Are they failing on your side as well?

slawlor:
Everything in the PR looks fine to me, so I'd just want to see why the integration tests are failing (make sure you have docker up and running before starting the test suite). Additionally you should be able to take a look at the integration tests logfile for more debug info if needed. Let me know if you need any help!

slawlor (changes requested):
We'll need to address the test infra failures before merging. Just a suggestion, but make sure the docker container is up and running before running ```cargo test``` with

```
docker compose up -d
```

slawlor (approved):

eozturk1:
Missed upgrading tokio in one of the Cargo files. (Interestingly, all tests were passing on my machine.) All tests pass now.

In the future, would it be better to combine these Cargo files -- unless we need separate ones for tests, akd, poc etc.?

slawlor:
Yeah each package needs it's own Cargo file. We have a common one which is referred to as a "workspace" however we don't want a single global one since it impacts the necessary dependencies when published to [crates.io]. If we do that, then someone who wants just an in-memory AKD will need all the mysql dependencies, etc. This keeps it as lean as possible in the published crates, and users pull just what they need.

PR #97, opened 2021-12-07
[akd_mysql] Relative path dependency requires a specific version

slawlor:
Just a tweak to allow publish to reference the right version when the package is on crates.io

### DIFF
--- a/akd_mysql/README.md
+++ b/akd_mysql/README.md
@@ -1,0 +1,45 @@
+## akd ![Build Status](https://github.com/novifinancial/akd/workflows/CI/badge.svg)
+
+An implementation of a MySQL storage layer for the auditable key directory (also known as a verifiable registry).
+
+Auditable key directories can be used to help provide key transparency for end-to-end encrypted
+messaging.
+
+This implementation is based off of the protocol described in
+[SEEMless: Secure End-to-End Encrypted Messaging with less trust](https://eprint.iacr.org/2018/607).
+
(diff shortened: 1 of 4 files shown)


akd_mysql/README.md
```diff
@@ -0,0 +1,45 @@
...
+Rust **1.51** or higher.
+
+Contributors
+------------
+
+The authors of this code are
+Jasleen Malvai ([@jasleen1](https://github.com/jasleen1)),
+Kevin Lewi ([@kevinlewi](https://github.com/kevinlewi)), and
+Sean Lawlor ([@slawlor](https://github.com/slawlor)).
+To learn more about contributing to this project, [see this document](./CONTRIBUTING.md).
```
kevinlewi, line 40: This needs to be `../CONTRIBUTING.md` right?

akd_mysql/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+License
+-------
+
+This project is [licensed](./LICENSE) under either Apache 2.0 or MIT, at your option.
```
kevinlewi, line 45: This needs to be `../LICENSE.md` right?

kevinlewi (approved):

akd_mysql/README.md
slawlor: There is no actual License.md file, there are LICENSE-MIT and LICENSE-APACHE. I've updated it to reflect both through different versioned links.

PR #100, opened 2021-12-08
Minor fixes for miscellaneous issues

afterdusk:
Spotted a few areas of improvement as I was reading the code.

Please let me know if any of the changes shouldn't be applied, the diffs are isolated so they will be easy to revert :)

### DIFF
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -29,13 +29,14 @@ use keyed_priority_queue::{Entry, KeyedPriorityQueue};
 
 /// The default azks key
 pub const DEFAULT_AZKS_KEY: u8 = 1u8;
+/// The default location of the azks root
+pub const DEFAULT_AZKS_ROOT: u64 = 0;
+
 /// An append-only zero knowledge set, the data structure used to efficiently implement
 /// a auditable key directory.
 #[derive(Debug, Serialize, Deserialize, Eq, PartialEq)]
 #[serde(bound = "")]
 pub struct Azks {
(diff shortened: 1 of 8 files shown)


akd/src/node_state.rs
```diff
@@ -124,7 +124,7 @@ impl NodeLabel {
 /// Hashes a label of type NodeLabel using the hash function provided by
 /// the generic type H.
 pub fn hash_label<H: Hasher>(label: NodeLabel) -> H::Digest {
-    let byte_label_len = H::hash(&label.get_len().to_ne_bytes());
+    let byte_label_len = H::hash(&label.get_len().to_be_bytes());
```
afterdusk, line 127: For consistency across architectures.

akd/src/append_only_zks.rs
```diff
@@ -29,13 +29,14 @@ use keyed_priority_queue::{Entry, KeyedPriorityQueue};
 
 /// The default azks key
 pub const DEFAULT_AZKS_KEY: u8 = 1u8;
+/// The default location of the azks root
+pub const DEFAULT_AZKS_ROOT: u64 = 0;
```
afterdusk, line 33: The same value is being committed to the database repeatedly since the root location is determined on the first initialization. But if the flexibility of a potentially dynamic root location is worth it, I can revert this change.

slawlor (approved):
Thanks for this @afterdusk! We definitely don't need to be serializing u64's across the wire for no reason when they're effectively immutable.

And the test cleanups are always helpful for better engineering!

Issue #101, opened 2021-12-08
Upgrade winter_* to 0.2.0


slawlor:
```winter_crypto``` and associated libs (```_math```, ```_utils```) seem to have released a new version, with better handling of bytes and big namespace cleanups. It'd be worthwhile to update to the latest version for these improvements

slawlor:
Rust docs [link](https://docs.rs/winter-crypto/0.2.0/winter_crypto/index.html)

PR #105, opened 2021-12-09
Winter

eozturk1:
Completes #101. 

Upgraded to a specific build (0.2.0) since the issue specified it.

### DIFF
--- a/akd/Cargo.toml
+++ b/akd/Cargo.toml
@@ -14,9 +14,9 @@ bench = []
 
 [dependencies]
 bincode = "1"
-winter-crypto = "0.2.0"
-winter-utils = "0.2.0"
-winter-math = "0.2.0"
+winter-crypto = "0.2"
+winter-utils = "0.2"
+winter-math = "0.2"
 rand = "0.8"
 keyed_priority_queue = "0.3"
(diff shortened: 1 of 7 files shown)


slawlor (approved):
LGTM, thanks!

akd/Cargo.toml
```diff
@@ -14,9 +14,9 @@ bench = []
 
 [dependencies]
 bincode = "1"
-winter-crypto = "0.1"
-winter-utils = "0.1"
-winter-math = "0.1"
+winter-crypto = "0.2.0"
```
slawlor, line 17: One small thing, can you change these to just be "0.2". Specifying the _full_ version doesn't allow us to get minor updates without manually patching. 0.0.X build changes are almost universally safe to upgrade and just contain small bug fixes, so ideally we would like the latest minor revision.

slawlor (changes requested):

akd/Cargo.toml
slawlor: Additionally this in all cargo files.

slawlor (approved):
Merge with /main and you should be good to go. Thanks!

PR #106, opened 2021-12-09
Adjusting visibility of publicly exposed records to all pub properties

slawlor:
We need all of the types which are publicly exposed to have completely ```pub``` properties (not ```pub(crate)```) because they will eventually be serialized over the wire and generally serializers will need to read the properties of these structs.

### DIFF
--- a/akd/Cargo.toml
+++ b/akd/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "akd"
-version = "0.3.3"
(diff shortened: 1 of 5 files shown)


slawlor:
So this is kind of the reverse problem of the issue, but it's adjusting the visibility of various records to be public so they're externally serializable

kevinlewi (approved):

PR #110, opened 2021-12-12
Add drop command and bench-lookup implementation to POC app

afterdusk:
## Drop Command
Not explicitly the purpose of the POC app, but a drop command is useful from a dev perspective after a schema change.

```console
evanau@evanau-mbp akd % cargo run -- drop
   Compiling akd v0.3.4 (/Users/evanau/Projects/akd/akd)
   Compiling akd_mysql v0.3.4 (/Users/evanau/Projects/akd/akd_mysql)
   Compiling akd_app v0.0.0 (/Users/evanau/Projects/akd/poc)
    Finished dev [unoptimized + debuginfo] target(s) in 8.91s
     Running `target/debug/akd_app drop`
======= Dropping database ======= 
[00:00:00.106] (110d8be00) INFO   Database dropped. (akd_app:191)
```

## Bench-Lookup Command
Added an implementation to the bench-lookup stub. 

It accepts a number of users and inserts all of them at once into the tree before starting the clock. It then looks up and verifies all users in the tree, with the whole process being repeated based on a second parameter `num_lookups_per_user`.

Example looking up 3 times per user for 100 users (300 lookups in total):
```console
evanau@evanau-mbp akd % cargo run -- bench-lookup 100 3
    Finished dev [unoptimized + debuginfo] target(s) in 0.42s
     Running `target/debug/akd_app bench-lookup 100 3`
======= Benchmark operation requested ======= 
Beginning LOOKUP benchmark of 100 users with 3 lookups/user
[00:00:00.103] (700009c83000) INFO   Starting the verifiable directory host (directory_host:59)
[00:00:00.152] (107f37e00) INFO   Inserting 100 users (akd_app:354)
[00:00:00.175] (70000987d000) INFO   Retrieved 0 previous user versions of 100 requested (directory:98)
[00:00:00.175] (70000987d000) INFO   Starting database insertion (directory:149)
[00:00:00.183] (70000987d000) INFO   Preload of tree (2 objects loaded), took 0.007699475 s (append_only_zks:202)
[00:00:00.418] (70000845f000) INFO   Cache hit since last: 14, cached size: 500 items (timed_cache:52)
[00:00:00.418] (70000845f000) INFO   Transaction writes: 1993, Transaction reads: 3791 (transaction:78)
[00:00:00.425] (70000845f000) INFO   MySQL writes: 502, MySQL reads: 2, Time read: 0.025402979 s, Time write: 0.151962079 s
        Tree size: 199
        Node state count: 200
        Value state count: 100 (mysql:716)
[00:00:00.821] (107f37e00) INFO   LOOKUP of 100 users complete (iteration 1) (akd_app:386)
[00:00:01.271] (107f37e00) INFO   LOOKUP of 100 users complete (iteration 2) (akd_app:386)
[00:00:01.675] (107f37e00) INFO   LOOKUP of 100 users complete (iteration 3) (akd_app:386)
Benchmark output: Looked up and verified 100 users with 3 lookups/user
Execution time: 1523 ms
Time-per-user (avg): 15234 µs
Time-per-op (avg): 5078 µs
[00:00:01.675] (70000a089000) INFO   AKD host shutting down (directory_host:192)
```


akd_mysql/src/mysql.rs
```diff
@@ -385,6 +385,28 @@ impl<'a> AsyncMySqlDatabase {
         Ok(())
     }
 
+    /// Drop all the tables
+    pub async fn drop_tables(&self) -> core::result::Result<(), MySqlError> {
```
slawlor, line 389: Isn't this just the same as the ```test_cleanup``` method?

Can we drop ```test_cleanup``` then and just make this a public method "drop_tables"? In the unit tests and integration tests we'll want to change the call to ```test_cleanup```

poc/src/main.rs
```diff
@@ -140,6 +145,9 @@ async fn main() {
     if cli.memory_db {
         let db = akd::storage::memory::AsyncInMemoryDatabase::new();
         let mut directory = Directory::<_>::new::<Blake3>(&db).await.unwrap();
+        if let Some(()) = pre_process_input(&cli, &tx, None).await {
```
slawlor, line 148: Is this the right position? This is specifically when a user has selected "in-memory database"

poc/src/main.rs
```diff
@@ -177,9 +209,9 @@ async fn process_input(
                 println!("Beginning DB INSERT benchmark of {} users", num_users);
 
                 let mut values: Vec<String> = vec![];
-                for _ in 0..*num_users {
+                for i in 0..*num_users {
                     values.push(
-                        thread_rng()
+                        StdRng::seed_from_u64(i)
```
slawlor, line 214: It's fine to change this, but why are we doing it?

slawlor (changes requested):
It's almost there, but I think there's a few adjustments to make. Thanks!

poc/src/main.rs
afterdusk: Ah the idea was to have both methods call the same pre-processing function. At this time there's only logic relevant to mysql, but in future we could have some logic shared between both or even memory specific logic. Happy to remove for now if that's what we prefer!

poc/src/main.rs
afterdusk: Ah so the reason is for repeatability and fairer comparisons when running the benchmarks between diffs. Without a fixed seed, the usernames generated between runs differ and can lead to a different node state count despite the same run parameters. Does that make sense?

slawlor: Yeah that's all good. Thanks

slawlor (approved):

slawlor:
LGTM thanks!

PR #113, opened 2021-12-13
Remove the epochs vector from ```HistoryTreeNode```

### LINKED ISSUE #112: Unbounded ```HistoryTreeNode``` field
HistoryTreeNode has an epochs field that is used to identify the epochs which an update took place.
As it stands now, the field can grow to an unbounded size. This will pose an issue when the corresponding DB record hits either soft or hard limits imposed by the DB schema or architecture.

## ESTIMATED LIMITS

Epochs are u64 = 8 bytes.

* For mysql, with the current VARBINARY(2000) definition, 2KB / 8B = ~ 250 values (we could always raise the max value in the definition, this is just for illustration)
* For some alternate key-value db (KVDB) with say 100KB recommended soft limit for record size, ignoring other overhead, 100KB / 8B = ~ 12500 values

## SOLUTION?

The associated epochs for a HistoryTreeNode can be obtained directly from the HistoryNodeState table given the node label, since each update corresponds to a HistoryNodeState.

* For mysql, this can be done through one call with a join, or a separate, concurrent call to the DB.
* For the alternate KVDB, one call is also sufficient. Instead of a "get" operation to fetch a singular HistoryTreeNode record, we can use a "scan" operation to fetch both the HistoryTreeNode record AND associated HistoryNodeState records at one go, provided we have a good sharding setup.

Accumulating the data from separate tables will incur some computation time on the database server and we will need to profile and assess the impact. However, I think this is necessary to avoid the unbounded record size issue. Are there any other potential solutions I've missed?

slawlor:
This PR removes the the unbounded vector of epochs from ```HistoryTreeNode``` in favor of just 2 u64's referring to the birth and death epochs.

Resolves #112

Still need a merge with main to resolve merge failures.

### DIFF
--- a/akd/src/storage/memory.rs
+++ b/akd/src/storage/memory.rs
@@ -266,6 +266,43 @@ impl Storage for AsyncInMemoryDatabase {
         }
         Ok(map)
     }
+
+    async fn get_epoch_lte_epoch(
+        &self,
+        node_label: crate::node_state::NodeLabel,
+        epoch_in_question: u64,
+    ) -> Result<u64, StorageError> {
+        let ids = (0..=epoch_in_question)
+            .map(|epoch| crate::node_state::NodeStateKey(node_label, epoch))
+            .collect::<Vec<_>>();
+        let data = self
+            .batch_get::<crate::node_state::HistoryNodeState>(ids)
+            .await?;
+        let mut epochs = data
+            .into_iter()
+            .map(|item| {
+                if let DbRecord::HistoryNodeState(state) = item {
+                    state.key.1
+                } else {
+                    0u64
+                }
+            })
+            .collect::<Vec<u64>>();
+        // reverse sort
+        epochs.sort_unstable_by(|a, b| b.cmp(a));
+
+        // move through the epochs from largest to smallest, first one that's <= ```epoch_in_question```
+        // and Bob's your uncle
+        for item in epochs {
+            if item <= epoch_in_question {
+                return Ok(item);
+            }
+        }
+        Err(StorageError::GetError(format!(
+            "Node (val: {}, len: {}) did not exist <= epoch {}",
+            node_label.val, node_label.len, epoch_in_question
+        )))
+    }
 }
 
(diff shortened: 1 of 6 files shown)


afterdusk:
I’m amazed at how fast you crunched this out. This all looks reasonable to me and is an elegant solution to the problem described in #112 🙂 Let's rebase this and let @kevinlewi have a chance to review before merging!

akd/src/storage/memory.rs
```diff
@@ -266,6 +266,43 @@ impl Storage for AsyncInMemoryDatabase {
         }
         Ok(map)
     }
+
+    async fn get_epoch_lte_epoch(
+        &self,
+        node_label: crate::node_state::NodeLabel,
+        epoch_in_question: u64,
+    ) -> Result<u64, StorageError> {
```
kevinlewi, line 274: So I guess the cool thing about doing this "linear search" for the correct epoch is that it can be done efficiently with a MySQL query, but may be less efficient if we are just relying on a simple set/get interface. Would it be better to find a middle-ground here with some pointer-based solution? Or perhaps we are not too worried about the performance impact of this function because it is not going to be called in lookup or publish?

slawlor: Yeah so the key history and audit pero queries do make this call, but we could probably figure out a nifty way to cache it but I think this is a good starting point at least.

I agree that it's not perfect but it completely mitigates the unbounded storage problems.

kevinlewi: Agreed, good point. Let's merge this for now then, and if we run into issues with performance on key history / audit, we can always revisit.

Thanks!!

kevinlewi (approved):

akd/src/storage/memory.rs
afterdusk: I believe most key-value storage systems provide some sort of key scan or match API that would make regex matching or prefix searching possible. 

It would then be up to the storage layer implementer to use an encoding scheme that would allow for the most efficient searching for the specific database :)

PR #117, opened 2021-12-16
Error cleanup

eozturk1:
- Error names and descriptions updated
- Unused or duplicate errors removed
- Errors reused when needed

### DIFF
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -479,13 +479,23 @@ impl Azks {
             let mut labels = [NodeLabel::root(); ARITY - 1];
             let mut hashes = [H::hash(&[0u8]); ARITY - 1];
             let mut count = 0;
-            let direction = dir.ok_or(AkdError::NoDirectionError)?;
+            let direction = dir.ok_or_else(|| {
+                AkdError::HistoryTreeNode(HistoryTreeNodeError::NoDirection(
+                    curr_node.label.get_val(),
+                    None,
+                ))
+            })?;
             let next_state = curr_state.get_child_state_in_dir(direction);
             if next_state == None {
                 break;
             }
             for i in 0..ARITY {
-                if i != dir.ok_or(AkdError::NoDirectionError)? {
+                if i != dir.ok_or_else(|| {
+                    AkdError::HistoryTreeNode(HistoryTreeNodeError::NoDirection(
+                        curr_node.label.get_val(),
+                        None,
+                    ))
+                })? {
                     labels[count] =
                         optional_history_child_state_to_label(&curr_state.child_states[i]);
                     hashes[count] = to_digest::<H>(&optional_history_child_state_to_hash::<H>(
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -490,12 +490,10 @@ impl Azks {
                 break;
             }
             for i in 0..ARITY {
-                if i != dir.ok_or_else(|| {
-                    AkdError::HistoryTreeNode(HistoryTreeNodeError::NoDirection(
-                        curr_node.label.get_val(),
-                        None,
-                    ))
-                })? {
+                let no_direction_error = AkdError::HistoryTreeNode(
+                    HistoryTreeNodeError::NoDirection(curr_node.label.get_val(), None),
+                );
+                if i != dir.ok_or(no_direction_error)? {
                     labels[count] =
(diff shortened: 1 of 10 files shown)


akd/src/append_only_zks.rs
```diff
@@ -241,9 +241,7 @@ impl Azks {
         }
 
         while !hash_q.is_empty() {
-            let (next_node_label, _) = hash_q
-                .pop()
-                .ok_or(AzksError::PopFromEmptyPriorityQueue(self.latest_epoch))?;
+            let (next_node_label, _) = hash_q.pop().unwrap();
```
slawlor, line 244: unwrap() is going to panic if it fails, rather than bubble up the error in an expected format. i.e. panic will terminate the app completely

akd/src/errors.rs
```diff
@@ -13,33 +13,33 @@ use crate::node_state::NodeLabel;
 /// Symbolizes a AkdError, thrown by the akd.
 #[derive(Debug)]
 pub enum AkdError {
-    /// Error propogation
-    HistoryTreeNodeErr(HistoryTreeNodeError),
-    /// Error propogation
-    DirectoryErr(DirectoryError),
-    /// Error propogation
+    /// Error propagation
```
slawlor, line 16: I feel like the comments, which are exported in the crate publicly could be cleaned up too over "error propogation"

akd/src/errors.rs
```diff
@@ -51,112 +51,78 @@ impl From<AzksError> for AkdError {
...
     /// The state of a node did not exist at a given epoch
-    NodeDidNotHaveExistingStateAtEp(NodeLabel, u64),
-    /// Error propogation
-    StorageError(StorageError),
-    /// Error propogation
+    NoStateAtEpoch(NodeLabel, u64),
+    /// Error propagation
+    Storage(StorageError),
+    /// Error propagation
     SerializationError,
```
slawlor, line 84: Is this error ever called? Storage errors also have a serialization error, so I'm curious if this one is utilized anywhere

akd/src/append_only_zks.rs
slawlor: Also @Jasleen1 do we need the priority queue in the end here? Dropping it would let us also drop a crate dependency :)

slawlor (changes requested):
A lot cleaner for sure, thanks! We need to definitely address the ```unwrap()``` call before merging, but it's definitely close. Few minor things and we can close this out 👍

akd/src/append_only_zks.rs
eozturk1: `PopFromEmptyPriorityQueue` implies the queue is empty but there is already an `isEmpty` check right before. 

Could `unwrap` fail here?

akd/src/errors.rs
eozturk1: Are they not auto-synced? I can look into it if not.

akd/src/errors.rs
eozturk1: The other `SerializationError` was not used; I removed it.

I guess the propagation comment is unnecessary now...

akd/src/append_only_zks.rs
slawlor: So maybe we could just rewrite as a ```while let Ok((next_node_label, _)) = ...``` statement, an avoid the is empty check all together.

slawlor: But yes I believe it can't fail, there's just better syntax to be used I think

akd/src/errors.rs
slawlor: I believe they're hand typed... no? Maybe there's a rust feature I'm unaware of? Check with intellisense maybe and see what it says for a description on that case.

slawlor (approved):

akd/src/append_only_zks.rs
eozturk1: I like the `while let` approach, updated!

akd/src/errors.rs
eozturk1: `cargo doc` generates the docs and puts them into `target/doc `. We can specify a website that serves these docs. (See `[documentation](https://doc.rust-lang.org/cargo/reference/manifest.html#the-documentation-field)`). 

Alternatively, if we don't, `docs.rs` automatically builds the docs and serves them. So doc updates in the code should be reflected in the published repo.

PR #120, opened 2021-12-22
Add (and run) the script that waits for akd to be published

eozturk1:
- The script first finds the expected version of AKD in the AKD_MYSQL's Cargo.toml file and queries the published crates index to see whether the expected version has been published.
- If not, the script waits with an increasing wait time for each retry.
- After the expected version is published, the publish process continues.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -27,7 +27,7 @@ jobs:
       env:
         CARGO_REGISTRY_TOKEN: ${{ secrets.crates_io_token }}
     - name: Run the script that waits for AKD to be published
-      run: ./.github/workflows/wait-for-akd-publish.sh
+      run: bash ./.github/workflows/wait-for-akd-publish.sh
     - name: Dry run publish AKD_MYSQL
       run: cargo publish --dry-run --manifest-path Cargo.toml -p akd_mysql
     - name: Publish crate akd_mysql
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -26,6 +26,8 @@ jobs:
       run: cargo publish --manifest-path Cargo.toml -p akd
       env:
         CARGO_REGISTRY_TOKEN: ${{ secrets.crates_io_token }}
+    - name: Run the script that waits for AKD to be published
+      run: ./.github/workflows/wait-for-akd-publish.sh
     - name: Dry run publish AKD_MYSQL
       run: cargo publish --dry-run --manifest-path Cargo.toml -p akd_mysql
     - name: Publish crate akd_mysql
--- a/.github/workflows/wait-for-akd-publish.sh
+++ b/.github/workflows/wait-for-akd-publish.sh
@@ -1,0 +1,45 @@
+# Finds expected version of a crate in another crate's Cargo.toml file
+get_crate_expected_version_number()
+{
+    local crate_name=$1
+    local crate_to_look_for_version=$2
+
+    local cargo_toml_file="$crate_name/Cargo.toml"
+    local expected_version=$(grep "$crate_to_look_for_version = " $cargo_toml_file | tr -d '{^}' | awk -F '[,:=]' '{print $5}')
+    echo $expected_version
+}
+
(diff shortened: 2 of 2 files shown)


.github/workflows/wait-for-akd-publish.sh
```diff
@@ -0,0 +1,46 @@
...
+    )
+    echo "AKD published versions:" $akd_versions_published
+    # Check whether published versions contain the expected version
+    if [[ $akd_versions_published == *"$akd_version_expected"* ]]; then
+        echo "Expected version has been published."
+        break
+    fi
+    echo "Expected version has not been published. Retrying after a wait."
+    wait_time=$((wait_time+1))
+    if [[ $wait_time == 42 ]]; then
```
slawlor, line 42: Uhm... why 42? lol

.github/workflows/wait-for-akd-publish.sh
```diff
@@ -0,0 +1,46 @@
+# Finds expected version of a crate in another crate's Cargo.toml file
```
slawlor, line 1: You should have a shell identifier here or we're going to run into all sorts of problems if we change the CI pipeline ever.

Something like:
```
#!/bin/bash
```
if you expect bash

.github/workflows/wait-for-akd-publish.sh
```diff
@@ -0,0 +1,46 @@
...
+
+# Wait for a specific akd version to be published to crates.io.
+# See https://github.com/novifinancial/akd/issues/116.
+# Must be run in the project root folder.
+akd_version_expected=$(get_crate_expected_version_number "akd_mysql" "akd")
+echo "AKD expected version in AKD_MYSQL:" $akd_version_expected
+wait_time=1
+while sleep $wait_time;
+do
+    akd_versions_published=$(
```
slawlor, line 31: If this _needs_ to be run from the repo root, you could ```pushd``` and ```popd``` it when you're done. We know the path of the script being run, so it should be do-able with relative paths. Something like
```bash
SCRIPTPATH=`dirname $SCRIPT` # Optionally you can do `dirname $0` since $0 is always the script path
pushd $SCRIPTPATH/../../
...do some work...
popd
```

example:
```bash
seanlawlor-mbp:SEEMless seanlawlor$ echo "$(dirname `pwd`)/../../"
/Users/seanlawlor/Code/../../
```

.github/workflows/publish.yml
```diff
@@ -26,6 +26,8 @@ jobs:
       run: cargo publish --manifest-path Cargo.toml -p akd
       env:
         CARGO_REGISTRY_TOKEN: ${{ secrets.crates_io_token }}
+    - name: Run the script that waits for AKD to be published
+      run: ./.github/workflows/wait-for-akd-publish.sh
```
slawlor, line 30: (Just checking) you're sure the run-directory of the CI pipeline is the repo root and not something wonky like ```.github/workflows/```?

slawlor (changes requested):

.github/workflows/wait-for-akd-publish.sh
eozturk1: 30 seconds gives ~8 minutes for AKD to be published before giving up. Publishing AKD took ~6 minutes (to my surprise) in one instance.

For 42, ~15 minutes should be enough to make sure AKD is published (I think... but I'm not sure of the variation here). Why exactly 42? Somewhat arbitrary and [possibly](https://www.scientificamerican.com/article/for-math-fans-a-hitchhikers-guide-to-the-number-42/) not...

.github/workflows/publish.yml
eozturk1: Yes, verified that CI uses the repo root as the run-directory.

.github/workflows/wait-for-akd-publish.sh
eozturk1: CI already runs the script from the root repo. This is more of a comment on explicit use of the script, i.e., publishing a version that is not triggered with a release.

slawlor (approved):

PR #121, opened 2021-12-23
Proof structs prefer vec of structs instead of paired vectors

davidjattfb:
Previously, `MembershipProof` and `NonMembershipProof` held two vectors of equal lengths, one holding `NodeLabel` and one holding `Digest`. This change introduces a `Node` struct that pairs these together, allowing the proofs to hold a single vector of these pairs.

The `AppendOnlyProof` was holding a single vector of tuples and has likewise been converted to prefer the Node struct.

In general, all `NodeLabel`+`Digest` pairings are now using the `Node` struct. The only one remaining is `HistoryChildState::new` which appears only to be used in tests. As such, I've let it be for now, just to keep this work internally contained to the proofs.

Re: https://github.com/novifinancial/akd/issues/118

*Testing*

```
cargo test
```

```
cd poc
cargo run
> PUBLISH test test
# and a few other commands
```

```
cargo bench --features bench --no-run
```

*Other*

```
cargo fmt
cargo clippy
```

*Notes*

This is a bit busier than I'd like. I had a version that left a lot of the internal helper messages alone but it appears I squashed those commits together out of habit before putting this up. Well, it is what it is :)

_Updated to add `cargo bench` call, plus `fmt` and `clippy`._

### DIFF
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -91,16 +91,15 @@ impl Azks {
     pub async fn insert_leaf<S: Storage + Sync + Send, H: Hasher>(
         &mut self,
         storage: &S,
-        label: NodeLabel,
-        value: H::Digest,
+        node: Node<H>,
     ) -> Result<(), AkdError> {
         // Calls insert_single_leaf on the root node and updates the root and tree_nodes
         self.increment_epoch();
 
         let new_leaf = get_leaf_node::<H, S>(
             storage,
-            label,
-            value.as_bytes().as_ref(),
+            node.label,
+            node.hash.as_bytes().as_ref(),
             NodeLabel::root(),
             self.latest_epoch,
         )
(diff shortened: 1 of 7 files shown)


aped:
This is great! Thanks, davidjattfb!

slawlor:
Hi @davidjattfb! Thanks for working on this issue! The code change is very well written and clean, and I think it's a good improvement on the codebase to cleanup undocumented tuples.

One small thing, which I forgot to mention earlier, are the "benches". We have some benchmarks which can be run to benchmark AZKS operations, and they aren't part of the normal compilation pipeline. You can run the benchmarks with 
```bash
cargo bench --features bench --no-run
```

It seems this change does impact that benchmark, so there is a minor change to do and then the CI pipeline will finish clean. [here's the CI run](https://github.com/novifinancial/akd/runs/4613597251?check_suite_focus=true) for reference

davidjattfb:
Oh, bah! I should have caught that while I was grepping around but I think I was looking only for the generic digest, not a specific. This should be a simple enough fix and I'll cast about a little for any other danglers.

davidjattfb:
Turns out that reference in `benches` appears to have been the only stray. Easy enough to get it in line.

I'm not sure of the preference, here, whether I should submit the update as a second commit or amend the previous commit and force-push it into place. (I chose the latter out of convention from another project.)

slawlor:
> Turns out that reference in `benches` appears to have been the only stray. Easy enough to get it in line.
> 
> I'm not sure of the preference, here, whether I should submit the update as a second commit or amend the previous commit and force-push it into place. (I chose the latter out of convention from another project.)

Yeah an additional commit is fine, we'll squash when we merge anyways

slawlor (approved):
Looks great! Thanks :)

Issue #119, opened 2021-12-16
Fixing membership and non-membership verifications


Jasleen1:
Currently, the membership and non-membership proof generation and verification both include membership proof operations as sub-routines. However, this is never actually used in the non-membership verification (see: https://github.com/novifinancial/akd/blob/b0a10e8e9cd17199a890686b36696f5da8f9d820/akd/src/client.rs#L82). This needs to be fixed and will likely require fixing `get_membership_proof_and_node` in `append_only_zks.rs`.

Issue #34, opened 2021-08-17
Adding VRF-based computation of labels


Jasleen1:
We want to use a VRF to compute labels in `seemless_directory.rs` and then verify these VRFs in `lookup_verify`, `key_history_verify` etc. 

See: https://crates.io/crates/vrf for one option. Below is a starting checklist of things to consider:

- [x] We will need to convert the output of the VRF into `NodeLabel` type. We may want to change the NodeLabel type as a result, or implement a trait instead.
- [ ] VRF keys need to be stored in some sort of `config` file, so that the server can generate VRF values.
- [ ] In the Java implementation of SEEMless, we used BouncyCastle, and at the time, we had to hash to the curve from strings/bytearrays. We need to hash to the correct space in this case as well.
- [ ] If we hash to a curve and later hash the output value, we should be careful about the security guarantees offered by the various parameter sizes.  Would be helpful to add documentation for this.

slawlor:
Completed, merged

kevinlewi:
I believe this should be done and this can be closed (finally!), but @Jasleen1 to confirm

PR #122, opened 2021-12-27
Verifiable random function

Jasleen1:
Added the VRF for NodeLabel computation, as well as verification.
This PR includes:
- A new dependency: the Rust vrf crate. 
- Updates to the function signatures for various proof structs.
- New functions for VRF label computations, proofs and verifications.

NOTE: At the moment, this code includes the VRF secret key from the VRF library and corresponding public key generation. We would want to change this later.

Addresses #34 and #119.


akd/src/client.rs
```diff
@@ -91,58 +96,106 @@ pub fn verify_nonmembership<H: Hasher>(
...
+    let mut stale_bytes = &[1u8];
+    if stale {
+        stale_bytes = &[0u8];
+    }
+
+    let hashed_label = H::merge(&[
+        name_hash_bytes,
+        H::merge_with_int(H::hash(stale_bytes), version),
+    ]);
+    // let label_slice = hashed_label.as_bytes();
```
slawlor, line 123: These unused fn's should all be cleaned up, they're effectively stale comments.

akd/src/directory.rs
```diff
@@ -14,15 +14,19 @@ use crate::proof_structs::*;
...
 use crate::errors::{AkdError, DirectoryError, HistoryTreeNodeError, StorageError};
 
+use crate::serialization::from_digest;
 use crate::storage::types::{AkdKey, DbRecord, ValueState, ValueStateRetrievalFlag, Values};
 use crate::storage::Storage;
 
 use log::{debug, error, info};
 use rand::{CryptoRng, RngCore};
 
+// use core::slice::SlicePattern;
```
slawlor, line 24: unused

akd/src/directory.rs
```diff
@@ -306,6 +313,48 @@ impl<S: Storage + Sync + Send> Directory<S> {
     /// or a version that we are retiring.
     pub(crate) fn get_nodelabel<H: Hasher>(uname: &AkdKey, stale: bool, version: u64) -> NodeLabel {
         // this function will need to read the VRF key using some function
+        // Initialization of VRF context by providing a curve
+        let mut vrf = ECVRF::from_suite(CipherSuite::SECP256K1_SHA256_TAI).unwrap();
+        // Inputs: Secret Key, Public Key (derived) & Message
+        let secret_key =
+            hex::decode("c9afa9d845ba75166b5c215767b1d6934e50c3db36e89b127b8a622b120f6721")
+                .unwrap();
+        // let public_key = vrf.derive_public_key(&secret_key).unwrap();
```
slawlor, line 322: unused

akd/src/directory.rs
```diff
@@ -306,6 +313,48 @@ impl<S: Storage + Sync + Send> Directory<S> {
     /// or a version that we are retiring.
     pub(crate) fn get_nodelabel<H: Hasher>(uname: &AkdKey, stale: bool, version: u64) -> NodeLabel {
         // this function will need to read the VRF key using some function
+        // Initialization of VRF context by providing a curve
+        let mut vrf = ECVRF::from_suite(CipherSuite::SECP256K1_SHA256_TAI).unwrap();
+        // Inputs: Secret Key, Public Key (derived) & Message
+        let secret_key =
+            hex::decode("c9afa9d845ba75166b5c215767b1d6934e50c3db36e89b127b8a622b120f6721")
```
slawlor, line 320: We should probably create a trait which allows getting & settings the public key of the VRF. That way it can be outsourced to the implementing library down-the-road at some point. We can have a test-implementation which is just a memory location, but the trait should be like the storage layer & injectable.

Additionally, if there is any hard-coded values (i.e. key values), they should map to a constant, so we don't suffer from copy-paste risks.

akd/src/directory.rs
```diff
@@ -306,6 +313,48 @@ impl<S: Storage + Sync + Send> Directory<S> {
...
+    pub(crate) fn get_label_proof<H: Hasher>(
+        &self,
+        uname: &AkdKey,
+        stale: bool,
+        version: u64,
+    ) -> Vec<u8> {
+        let mut vrf = ECVRF::from_suite(CipherSuite::SECP256K1_SHA256_TAI).unwrap();
+        // Inputs: Secret Key, Public Key (derived) & Message
+        let secret_key =
+            hex::decode("c9afa9d845ba75166b5c215767b1d6934e50c3db36e89b127b8a622b120f6721")
```
slawlor, line 354: map value to const

akd/src/directory.rs
```diff
@@ -306,6 +313,48 @@ impl<S: Storage + Sync + Send> Directory<S> {
...
+                .unwrap();
+        // let public_key = vrf.derive_public_key(&secret_key).unwrap();
```
slawlor, line 356: unused

akd/src/directory.rs
```diff
@@ -316,9 +365,23 @@ impl<S: Storage + Sync + Send> Directory<S> {
             name_hash_bytes,
             H::merge_with_int(H::hash(stale_bytes), version),
         ]);
-        let label_slice = hashed_label.as_bytes();
-        let hashed_label_bytes = convert_byte_slice_to_array(&label_slice);
-        NodeLabel::new(u64::from_ne_bytes(hashed_label_bytes), 64u32)
+        // let label_slice = hashed_label.as_bytes();
```
slawlor, line 368: unused

akd/src/directory.rs
```diff
@@ -316,9 +365,23 @@ impl<S: Storage + Sync + Send> Directory<S> {
...
+        // Initialization of VRF context by providing a curve
+        let mut vrf = ECVRF::from_suite(CipherSuite::SECP256K1_SHA256_TAI).unwrap();
+        // Inputs: Secret Key, Public Key (derived) & Message
+        let secret_key =
+            hex::decode("c9afa9d845ba75166b5c215767b1d6934e50c3db36e89b127b8a622b120f6721")
+                .unwrap();
+        vrf.derive_public_key(&secret_key).unwrap()
     }
 
     // FIXME: Make a real commitment here, alongwith a blinding factor.
```
slawlor, line 387: Probably add a ref here to the issue you opened related to commitments? So we can link it back

akd/src/directory.rs
```diff
@@ -486,6 +546,12 @@ pub async fn get_key_history_hashes<S: Storage + Sync + Send, H: Hasher>(
     Ok((root_hashes, previous_root_hashes))
 }
 
+fn vec_to_u8_arr(vector_u8: Vec<u8>) -> [u8; 32] {
```
slawlor, line 549: Definitely add a comment here that this TRUNCATES, so if anyone uses this in the future they know the behavior

akd/src/errors.rs
```diff
@@ -88,10 +90,10 @@ impl fmt::Display for HistoryTreeNodeError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             Self::NoDirection(node_label, child_label) => {
-                let mut to_print = format!("no direction provided for the node {}", node_label);
+                let mut to_print = format!("no direction provided for the node {:?}", node_label);
```
slawlor, line 93: maybe we should implement```std::fmt::Display``` for ```NodeLabel``` so it'd look something like 
```rust
write!(stream, "({}, 0x{})", label.len, hex::encode(&label.val))
```

e.g. outputing ```(123, 0x123978acf12)```

slawlor (changes requested):
Approved in-principal, but some small things need to be addressed. The biggest is with the public key of the VRF

akd/src/node_state.rs
```diff
@@ -237,8 +251,8 @@ impl Storable for HistoryNodeState {
             result.push(*byte);
         }
 
-        let parts: [[u8; 8]; 2] = [key.0.val.to_be_bytes(), key.1.to_be_bytes()];
-        for iarray in &parts {
+        let parts: [&[u8]; 2] = [&key.0.val, &key.1.to_be_bytes()];
+        for iarray in parts {
             for byte in iarray {
                 result.push(*byte);
             }
```
afterdusk, line 258: Nit: We can merge this whole chunk:
```rust
let parts: [&[u8]; 3] = [&key.0.len.to_be_bytes(), &key.0.val, &key.1.to_be_bytes()];
```

akd/src/storage/tests.rs
```diff
@@ -176,7 +176,7 @@ async fn test_batch_get_items<Ns: Storage>(storage: &Ns) {
                 plaintext_val: Values(value.clone()),
                 version: epoch,
                 label: NodeLabel {
-                    val: 1u64,
+                    val: byte_arr_from_u64(1u64),
```
afterdusk, line 179: Nit: We can remove the `u64` for consistency with the other calls to `byte_arr_from_u64` in this file :)

afterdusk (commented):
I'm not as familiar with the use of VRF in the protocol, so I was only able to add some non-blocking nits. Thanks for working on this!

akd/src/node_state.rs
```diff
@@ -169,8 +182,9 @@ impl NodeLabel {
 /// Hashes a label of type NodeLabel using the hash function provided by
 /// the generic type H.
 pub fn hash_label<H: Hasher>(label: NodeLabel) -> H::Digest {
-    let byte_label_len = H::hash(&label.get_len().to_be_bytes());
-    H::merge_with_int(byte_label_len, label.get_val())
+    let byte_label_len = H::hash(&label.get_len().to_ne_bytes());
```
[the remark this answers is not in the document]
Jasleen1: Yes, great catch! Changed everywhere there was `to_ne_bytes()`.

akd/src/client.rs
```diff
@@ -91,58 +96,106 @@ pub fn verify_nonmembership<H: Hasher>(
...
+    label: NodeLabel,
+) -> Result<(), AkdError> {
+    // Initialization of VRF context by providing a curve
+    let mut vrf = ECVRF::from_suite(CipherSuite::SECP256K1_SHA256_TAI).unwrap();
+
+    let name_hash_bytes = H::hash(uname.0.as_bytes());
+    let mut stale_bytes = &[1u8];
+    if stale {
+        stale_bytes = &[0u8];
+    }
```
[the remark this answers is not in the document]
Jasleen1: Done.

akd/src/directory.rs
Jasleen1: Makes sense and commented in both places this is used.

slawlor:
@Jasleen1 do you think we should land this as-is or should we add the VRF crypto trait first? Personally I think we probably shouldn't embed the VRF key material in code directly in the /main branch. I'd like to have this changed prior to merging

slawlor (approved):

slawlor:
As a note, this PR adds full support for VRFs across the various components of the AKD structure (client, lean client, etc). However there is another issue open (#144) which is still pending, hence the commenting out of 2 portions of tests for the moment.

Issue #125, opened 2022-01-07
AKD Quorum distributed change auditor


slawlor:
## Overview

To prevent a split-view attack the root hash of every epoch needs to have a signature signed by a private key which cannot be leaked from the quorum (via shared-secret methodology). This quorum participates to independently validate the append-only proof of the key directory and each participant provides their partial shard of the quorum signing key and when enough participants agree, the changes are signed off on and stored in persistent storage. 

That way a proof only needs to give the root hash, and the signature on it to ascertain the quorum has agreed on the changes, and the AKD (or any other 3rd party) cannot generate its own signatures.

Eventually these auditors can be participants from external entities who can participate in the quorum vote. 

## Requirements

1. The Quorum Key (QK) is long-lived and securely managed such that individual (or a few) participants cannot generate their own signatures and invalidate the trust of the system
2. The quorum participants can evolve over time (i.e. removal in the face of failures, and additions for future collaborative growth or upgrades).
3. Communication channels are secure, such that 3rd parties cannot sniff the key shares over the wire and compromise the QK.
4. The public key of the QK is publicly distributed such that it's easy to client-side validate the validity of the epochs without additional I/O operations to a potentially malicious acting key directory.

## Initial design overview

A Quorum Key is generated at the beginning of the life of the quorum, and the private key is broken into "shards" via [Shamir secret sharing](https://github.com/dsprenkels/sss-rs).
 These shards are transmitted to the quorum participants who hold them in secure storage. The shards are generated with the following properties

 1. There are 3 _f_ + 1 nodes in the quorum
 2. 2 _f_ + 1 nodes need to agree to reconstruct the signing key (quorum key)
 3. The public key is given publicly to every client which will need to verify the signature

 ## Communcation with the quorum

 The collection of nodes in this quorum receive commands from an external 3rd party (key directory epoch notification or admin interface for example).
 The messages they can receive are the following

 1. New epoch publication - The AKD has published a new epoch and the quorum should validate the changes and generate a signature
 2. Quorum member enrollment - The quorum should attempt to add the new member to the set. A series of independent tests will be
 performed against the new member and if they pass, a new shard-set will be generated and the node enrolled
 3. Quorum member removal - The quorum has a member which should be removed. If 2 _f_ + 1 nodes agree, the Quorum Key is reconstructed and new shards are generated and
 distributed to the remaining membership.

[...]

PR #126, opened 2022-01-07
PLACEHOLDER: AKD Quorum voter initial logic

slawlor:
This is an initial skeleton structure for what a AKD Quorum node can look like.

## Overview

This PR adds preliminary support for a distributed quorum of nodes which inter-communicate to validate and generate a distributed signature on the changes coming from the AKD key directory (i.e. an auditor of the changes in the AKD directory between epochs).

By design, this would be a separately hosted system which the process generating a new epoch of the AKD (key directory process publishing new records) communicates with and pushes a new *append-only proof* which validates that the changes are not mutating previous data and are appending changes only. 

Anything operated by the ```QuorumCryptographer``` should be done within a secure storage area to assert that key components cannot leak. Each node needs its own public/private key-pair to encrypt node-to-node messages as well as an ability to securely store their "shard" of the quorum signing private key. Meanwhile the ```QuorumStorage``` operator is assumed to be less-secure storage as trust isn't needed at that boundary for cryptographic secret storage.

## Still to be solved:
1. We still need to generate a valid "test" for other nodes. I.e. a full AKD tree which is either a valid or an invalid proof construction.
2. We will need test implementations and test coverage of most of the FSM logic. Presently there is none in this PR, but it's not prod ready yet.

Addresses #125


slawlor:
Had to fix the CI pipeline (needed protoc installed)

alexme22:
Overall I'd suggest to split huge PRs into small ones.

akd_quorum/src/comms/nonces.rs
```diff
@@ -0,0 +1,85 @@
...
+
+        if nonce == expected {
+            self.incoming.write().await.insert(from, expected + 1);
+            Ok(())
+        } else {
+            let err = crate::comms::CommunicationError::NonceError(
+                from,
+                nonce,
+                format!(
+                    "Nonce mismatch in raft inter-messages: Node {}, Nonce: {}, Expected Nonce: {}",
```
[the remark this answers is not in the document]
slawlor: Ah yes, raft was the "first" attempt at inter-node comms, however we determined it was more complex than necessary and therefore moved to something simpler. There's probably some artifacts laying around in this branch because of it.

akd_quorum/src/comms/nonces.rs
```diff
@@ -0,0 +1,85 @@
...
+        Self {
+            outgoing: Arc::new(RwLock::new(HashMap::new())),
+            incoming: Arc::new(RwLock::new(HashMap::new())),
+        }
+    }
+
+    /// Retrieve the "next" nonce for message send, will auto-increment the nonce
+    /// mapping
+    pub(crate) async fn get_next_outgoing_nonce(&self, to: NodeId) -> Nonce {
+        let mut guard = self.incoming.write().await;
```
slawlor, line 35: Ah this is a typo, outgoing should be used here.

slawlor:
> Overall I'd suggest to split huge PRs into small ones.

In general, I agree. I needed a prototyping playground however to put all the pieces together and generate the FSM logic for the node. I can easily split it with say boiler-plate, messages, and then the FSM components rounding out the feature.

Until the full FSM state had been realized, the messaging structure changed. And when we removed Raft, it changed how we process messages, etc. So there was a lot of mutability in this PR during it's development hence the single branch.

akd_quorum/src/node/mod.rs
```diff
@@ -0,0 +1,2217 @@
...
+        };
+        reply.send(result).map_err(|_| {
+            QuorumOperationError::Communication(crate::comms::CommunicationError::SendError(
+                "Failed to send reply in RPC call (likely channel closed)".to_string(),
+            ))
+        })?;
+        Ok(())
+    }
+
+    async fn encrypt_message(
```
alexme22, line 458: Let's move it to separate "encryption" mod.

akd_quorum/src/node/mod.rs
```diff
@@ -0,0 +1,2217 @@
...
+        Ok(Message {
+            from: message.from,
+            to: message.to,
+            nonce,
+            serialized_message: data,
+        })
+    }
+
+    /// Generate a test for a new node
+    async fn generate_test(&self) -> Result<(bool, NewNodeTest<H>), QuorumOperationError> {
```
alexme22, line 529: How about to extract testing feature into it's own mod?

akd_quorum/src/node/mod.rs
```diff
@@ -0,0 +1,2217 @@
...
+        let mut guard = self.state.status.write().await;
+        *guard = new_state;
+    }
+
+    async fn get_state(&self) -> NodeStatus<H> {
+        let guard = self.state.status.read().await;
+        guard.clone()
+    }
+
+    async fn public_verify_impl(
```
alexme22, line 643: I'd extract verification logic into separate mod.

akd_quorum/src/node/mod.rs
```diff
@@ -0,0 +1,2217 @@
...
+
+                // wait on the self-review of the changes
+                self_handle.await?;
+            }
+        }
+
+        Ok(())
+    }
+
+    async fn public_enroll_impl(
```
alexme22, line 774: Same for quorum management feature - it could be moved to separate mod.

alexme22:
> > I can easily split it with say boiler-plate, messages, and then the FSM components rounding out the feature.

That would help a lot, please do! 
The split of node\mod.rs into submodules could be also benefitial. verification, membership management (add/remove/test), encryption/decryption components could be extracted to new mods (and probably PRs).

akd_quorum/src/node/mod.rs
```diff
@@ -0,0 +1,2217 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+//
+// This source code is licensed under both the MIT license found in the
+// LICENSE-MIT file in the root directory of this source tree and the Apache
+// License, Version 2.0 found in the LICENSE-APACHE file in the root directory
+// of this source tree.
+
+//! This module handles the node's primary message handling logic and
```
alexme22, line 8: Looks like it handles much more functionality than described here. I'd suggest splitting the mod into smaller pieces each focusing on (ideally) single function with clear mod description - it will make review and future maintainance simpler. 
I've commented down the file on some possible submodules following function naming but this is a just suggestion. What do you think?

akd_quorum/src/crypto.rs
```diff
@@ -0,0 +1,421 @@
...
+                        .to_string(),
+                ));
+            }
+        }
+        Ok(potential_result)
+    }
+}
+
+#[cfg(test)]
+mod crypto_tests {
```
artemkorneev, line 294: I'd propose to move tests into separate files. It is not such a big difference from the code organization point of view, but it is easier to review code if the tests are separated - if the change is only in crypto_tests.rs, than it is clear that the production code is not changed.

akd_quorum/src/node/mod.rs
```diff
@@ -0,0 +1,2217 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
```
artemkorneev, line 1: Well.. I think, some code walk-through Zoom call would be useful.

akd_quorum/src/crypto.rs
```diff
@@ -0,0 +1,421 @@
...
+/// signing key. A single shard cannot be utilized to reconstruct the
+/// full quorum key.
+///
+/// Due to limitations of the Shamir's Secret Sharing lib, we are constrained
+/// to break the secret information into batches of DATA_SIZE _exactly_ to generate
+/// the shards. This means that to support a key bigger than DATA_SIZE, we need to
+/// have multiple shards for each slice of the secret information.
+#[derive(PartialEq, Debug)]
+pub struct QuorumKeyShard {
+    pub(crate) components: [[u8; SHARE_SIZE]; QUORUM_KEY_NUM_PARTS],
```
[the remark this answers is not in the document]
slawlor: It's just an array of array's. The Shamir secret sharing crate only supports breaking a byte array of size ```DATA_SIZE``` into components of ```SHARE_SIZE```. ```DATA_SIZE``` is set to 64 bytes, so assuming we'd want a key larger than that, we need to break it into parts. I.e. a "shard" is an array of "shards" which is of length ```QUORUM_KEY_NUM_PARTS```. It does mean the QK needs to be a modulo of the ```DATA_SIZE``` but that should be OK.

akd_quorum/src/node/mod.rs
slawlor: Agreed, like I mentioned earlier this was to serve as a jumping off point and be ultimately broken into separate PRs of smaller scale. It proved painful to not try and design all the traits/messages/crypto together in a single branch, but now that we have a relatively decent idea, we should be able to break it apart reasonably.

akd_quorum/src/node/mod.rs
```diff
@@ -0,0 +1,2217 @@
...
+            .handle_message(NodeMessage::Internal(
+                node_id,
+                InterNodeMessage::<H>::TimerTick,
+            ))
+            .await?;
+
+        Ok(())
+    }
+
+    async fn handle_inter_node_message(
```
alexme22, line 405: I'd suggest to rename `handle_inter_node_message` into something like `process_inter_node_message` to not confuse with others `inter_node_message_handler` and `handle_inter_node_message_helper`

akd_quorum/src/node/mod.rs
```diff
@@ -0,0 +1,2217 @@
...
+                    // no-op, message handled with no reply
+                }
+                (Err(err), _) => {
+                    error!("Error handling backlogged message: {}", err);
+                }
+            }
+        }
+    }
+
+    async fn inter_node_message_handler(
```
alexme22, line 336: This method handles public messages read from `reciever` channel and inter_node messages. What do you think on splitting it into two?

akd_quorum/src/node/mod.rs
```diff
@@ -0,0 +1,2217 @@
...
+        &self,
+        receiver: &mut tokio::sync::mpsc::Receiver<PublicNodeMessage<H>>,
+    ) -> Result<(), QuorumOperationError> {
+        loop {
+            tokio::select! {
```
alexme22, line 341: This macro cancels other threads once first one is completed. I'd guess this may lead to dropped messages or interrupted executions in the cancelled thread. Does node tolerate that?

akd_quorum/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "akd_quorum"
+version = "0.3.7"
+authors = ["Sean Lawlor <seanlawlor@fb.com>", "Kevin Lewi <klewi@fb.com>"]
+description = "Validation of a key directory's latest epoch using a quorum of members which validate the append-only nature of the AKD"
+license = "MIT OR Apache-2.0"
+edition = "2018"
```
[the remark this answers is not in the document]
slawlor: We're using 2018 for downstream usage of this crate. We're always building with the latest Rust but with 2018 compat.

kevinlewi:
We can reopen this if we revisit it in the future

PR #128, opened 2022-01-12
Standardize AKD label/value partners.

davidjattfb:
This renames:
* AkdKey => AkdLabel
* Values => AkdValue

This hews closer to the SEEMless terms "Label" and "Value" without treading too close to something like `NodeLabel`.

https://github.com/novifinancial/akd/issues/104

This commit represents essentially just a rename of the structs, dropped in place. This is the first part of issue 104 and is pretty low stakes. I want to consider it separate from the forthcoming `String` to `Vec<u8>` conversion which is a much hairier thread to pull.

**Testing**

```
% cargo test 
% cargo fmt
% cargo clippy
% cargo bench --features bench --no-run
```

```
% cd poc
% cargo run
Please enter a command
> [00:00:00.041] (70000fe41000) INFO   Starting the verifiable directory host (directory_host:59)
publish test1 test2
[00:00:06.137] (70000f432000) INFO   Retrieved 0 previous user versions of 1 requested (directory:98)
[00:00:06.137] (70000f432000) INFO   Starting database insertion (directory:155)
[00:00:06.229] (70000f432000) INFO   Preload of tree (14 objects loaded), took 0.091703745 s (append_only_zks:198)
[00:00:06.338] (70000e61d000) INFO   Cache hit since last: 84, cached size: 23 items (timed_cache:52)
[00:00:06.338] (70000e61d000) INFO   Transaction writes: 0, Transaction reads: 0 (transaction:78)
[00:00:06.343] (70000e61d000) INFO   MySQL writes: 39, MySQL reads: 10, Time read: 0.111147572 s, Time write: 0.098209911 s
        Tree size: 21
        Node state count: 35
        Value state count: 7 (mysql:693)
Response: PUBLISHED 'test1' = 'test2' in 0.224341848 s (epoch: 7, root hash: cacb8325c7dabdd0e697fcb1088a91be30473b961137ba2aa6dfbd7d0e938f5a)
Please enter a command
>
```

### DIFF
--- a/akd/src/storage/types.rs
+++ b/akd/src/storage/types.rs
@@ -27,12 +27,12 @@ pub enum StorageType {
 
 /// The keys for this key-value store
 #[derive(Debug, Clone, Hash, Eq, PartialEq, Serialize, Deserialize)]
-pub struct AkdKey(pub String);
+pub struct AkdLabel(pub String);
 
-/// The types of values used in the key-value pairs of a AKD
+/// The types of value used in the key-value pairs of a AKD
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Hash)]
 #[serde(bound = "")]
-pub struct Values(pub String);
+pub struct AkdValue(pub String);
 
 /// State for a value at a given version for that key
 #[derive(Debug, Clone, Hash, Eq, PartialEq, Serialize, Deserialize)]
@@ -43,7 +43,7 @@ pub struct ValueStateKey(pub String, pub u64);
 #[serde(bound = "")]
 pub struct ValueState {
     /// The plaintext value of the user information in the directory
-    pub plaintext_val: Values, // This needs to be the plaintext value, to discuss
+    pub plaintext_val: AkdValue, // This needs to be the plaintext value, to discuss
     /// The version of the user's value-state
     pub version: u64, // to discuss
     /// The Node Label
@@ -51,7 +51,7 @@ pub struct ValueState {
     /// The epoch this value state was published in
     pub epoch: u64,
     /// The username associated to this value state (username + epoch is the record key)
-    pub username: AkdKey,
+    pub username: AkdLabel,
 }
 
 impl crate::storage::Storable for ValueState {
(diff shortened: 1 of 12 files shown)


akd/src/storage/types.rs
```diff
@@ -27,12 +27,12 @@ pub enum StorageType {
...
 #[derive(Debug, Clone, Hash, Eq, PartialEq, Serialize, Deserialize)]
-pub struct AkdKey(pub String);
+pub struct AkdLabel(pub String);
 
-/// The types of values used in the key-value pairs of a AKD
+/// The types of value used in the key-value pairs of a AKD
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Hash)]
 #[serde(bound = "")]
-pub struct Values(pub String);
+pub struct AkdValue(pub String);
```
slawlor, line 35: I thought we were changing this to a ```Vec<u8>```. Looking at how it's utilized down-stream as well, we will want the label in binary I think as well...so

```rust
pub struct AkdLabel(pub Vec<u8>)
pub struct AkdValue(pub Vec<u8>)
```

slawlor (changes requested):
So in general it looks fine, but we want to change the label and value to binary instead of strings will will likely induce other changes down-stream (hashing, poc, etc). Thanks!

akd/src/storage/types.rs
davidjattfb: Yes, I'm working on that part separately after the name change, hence this comment in the PR summary above:

> This commit represents essentially just a rename of the structs, dropped in place. This is the first part of issue 104 and is pretty low stakes. I want to consider it separate from the forthcoming `String` to `Vec<u8>` conversion which is a much hairier thread to pull.

My intent here is to consider the two steps separately for ease of review, but I'm happy to combine them into a single PR if you'd prefer?

slawlor: Ah I'm sorry I missed that. 2 PRs is fine. Approving to unblock. Thanks!

slawlor (approved):

PR #129, opened 2022-01-14
Change the types of AkdLabel and AkdValue to be an array of bytes.

davidjattfb:
https://github.com/novifinancial/akd/issues/104

Previously, these values were traded around as `String` values for convenience of readability. This change moves these values to `Vec<u8>` to reflect that these are intended as arbitrary blobs of bytes.

The POC continues to traffic in strings since it's a demonstration meant for the command line, where the readability is more convenient.

**Testing**

```
% cargo test 
% cargo fmt
% cargo clippy
% cargo bench --features bench --no-run
```

```
% cd poc
% cargo run
Please enter a command
> [00:00:00.036] (70000fa7d000) INFO   Starting the verifiable directory host (directory_host:59)
publish banana green
[00:00:08.006] (70000f87a000) INFO   Retrieved 0 previous user versions of 1 requested (directory:98)
[00:00:08.006] (70000f87a000) INFO   Starting database insertion (directory:155)
[00:00:08.100] (70000f87a000) INFO   Preload of tree (10 objects loaded), took 0.093771548 s (append_only_zks:198)
[00:00:08.232] (70000e056000) INFO   Cache hit since last: 63, cached size: 18 items (timed_cache:52)
[00:00:08.232] (70000e056000) INFO   Transaction writes: 0, Transaction reads: 0 (transaction:78)
[00:00:08.239] (70000e056000) INFO   MySQL writes: 31, MySQL reads: 8, Time read: 0.118470512 s, Time write: 0.119331427 s
        Tree size: 7
        Node state count: 11
        Value state count: 3 (mysql:693)
Response: PUBLISHED 'banana' = 'green' in 0.256827742 s (epoch: 3, root hash: 487aa4a60b398ee51b90374a0865aac769d1222cfe6a023933b51edc732ca864)
Please enter a command
> publish banana yellow
[00:00:12.656] (70000f87a000) INFO   Retrieved 1 previous user versions of 1 requested (directory:98)
[00:00:12.656] (70000f87a000) INFO   Starting database insertion (directory:155)
[00:00:12.657] (70000f87a000) INFO   Preload of tree (6 objects loaded), took 0.000360711 s (append_only_zks:198)
[00:00:12.809] (70000e056000) INFO   Cache hit since last: 75, cached size: 28 items (timed_cache:52)
[00:00:12.810] (70000e056000) INFO   Transaction writes: 0, Transaction reads: 0 (transaction:78)
[00:00:12.815] (70000e056000) INFO   MySQL writes: 40, MySQL reads: 1, Time read: 0.01543748 s, Time write: 0.138823311 s
        Tree size: 11
        Node state count: 16
        Value state count: 4 (mysql:693)
Response: PUBLISHED 'banana' = 'yellow' in 0.174664147 s (epoch: 4, root hash: 6526a193ba85f40617cb2805f87df33d34405ffde8c3b84df8fe03067f86dc83)
Please enter a command
> lookup banana
Response: Lookup proof verified for user 'banana'
Please enter a command
> history banana
Response: GOT KEY HISTORY FOR 'banana'
Please enter a command
> x
[00:00:23.132] (10e138600) INFO   Exiting... (akd_app:439)
```

### DIFF
--- a/akd/src/directory.rs
+++ b/akd/src/directory.rs
@@ -32,14 +32,14 @@ pub struct EpochHash<H: Hasher>(pub u64, pub H::Digest);
 impl AkdValue {
     /// Gets a random value for a AKD
     pub fn random<R: RngCore + CryptoRng>(rng: &mut R) -> Self {
-        Self(get_random_str(rng))
+        Self(get_random_bytes(rng))
     }
 }
 
 impl AkdLabel {
-    /// Creates a random key for a AKD
+    /// Creates a random label for a AKD
     pub fn random<R: RngCore + CryptoRng>(rng: &mut R) -> Self {
-        Self(get_random_str(rng))
+        Self(get_random_bytes(rng))
     }
 }
 
@@ -242,7 +242,6 @@ impl<S: Storage + Sync + Send> Directory<S> {
         &self,
         uname: &AkdLabel,
     ) -> Result<HistoryProof<H>, AkdError> {
-        let username = uname.0.to_string();
         if let Ok(this_user_data) = self.storage.get_user_data(uname).await {
             let mut proofs = Vec::<UpdateProof<H>>::new();
             for user_state in &this_user_data.states {
@@ -252,7 +251,7 @@ impl<S: Storage + Sync + Send> Directory<S> {
             Ok(HistoryProof { proofs })
         } else {
             Err(AkdError::Directory(DirectoryError::NonExistentUser(
-                username,
+                uname.0.clone(),
                 self.current_epoch,
             )))
         }
@@ -310,7 +309,7 @@ impl<S: Storage + Sync + Send> Directory<S> {
         version: u64,
     ) -> NodeLabel {
         // this function will need to read the VRF key using some function
-        let name_hash_bytes = H::hash(uname.0.as_bytes());
+        let name_hash_bytes = H::hash(uname.0.as_slice());
         let mut stale_bytes = &[1u8];
         if stale {
             stale_bytes = &[0u8];
@@ -444,10 +443,11 @@ fn convert_byte_slice_to_array(slice: &[u8]) -> [u8; 8] {
     out_arr
 }
 
-fn get_random_str<R: RngCore + CryptoRng>(rng: &mut R) -> String {
-    let mut byte_str = [0u8; 32];
-    rng.fill_bytes(&mut byte_str);
-    format!("{:?}", &byte_str)
+/// get 32 random bytes in a Vec
+fn get_random_bytes<R: RngCore + CryptoRng>(rng: &mut R) -> Vec<u8> {
+    let mut buf = [0u8; 32];
+    rng.fill_bytes(&mut buf);
+    Vec::from(buf)
 }
 
 type KeyHistoryHelper<D> = (Vec<D>, Vec<Option<D>>);
@@ -510,7 +510,7 @@ mod tests {
         let mut akd = Directory::<_>::new::<Blake3>(&db).await?;
 
         akd.publish::<Blake3>(
-            vec![(AkdLabel("hello".to_string()), AkdValue("world".to_string()))],
+            vec![(AkdLabel(Vec::from("hello")), AkdValue(Vec::from("world")))],
             false,
         )
         .await?;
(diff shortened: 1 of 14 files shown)


akd/src/storage/types.rs
```diff
@@ -85,11 +82,8 @@ impl crate::storage::Storable for ValueState {
         }
         let epoch_bytes: [u8; 8] = bin[1..=8].try_into().expect("Slice with incorrect length");
         let epoch = u64::from_be_bytes(epoch_bytes);
-        if let Ok(username) = std::str::from_utf8(&bin[9..]) {
-            Ok(ValueStateKey(username.to_string(), epoch))
-        } else {
-            Err("Invalid string format".to_string())
-        }
+        let username = Vec::from(&bin[9..]);
+        Ok(ValueStateKey(username, epoch))
```
davidjattfb, line 86: Notable here, both before and after, is that there's no guard that the parameter `bin` is long enough.

integration_tests/src/test_util.rs
```diff
@@ -125,22 +125,22 @@ impl log::Log for FileLogger {
...
-                .collect(),
-        );
+    let mut users: Vec<Vec<u8>> = vec![];
+
+    // ensure that one user is string bytes and the other is bytes that are not utf-8
+    users.push(Vec::from("a username that is valid utf-8"));
+    users.push(vec![0x80]); // a username that is NOT valid utf-8
+
+    for _ in 2..num_users {
+        users.push((&mut rng).sample_iter(&Standard).take(30).collect());
```
davidjattfb, line 143: For this particular test, I was having an issue wherein I'd updated the wrong database such that the test was still trying to put this value to a `VARCHAR` when I thought I'd already updated it to a `VARBINARY`. The MySQL parameter replacement would quietly try to coerce the value to a string.

However, this test and a bunch of other tests in this project are generating random data. This meant that sometimes this test would _work_ because the random bytes it happened to have generated were legal UTF-8. Since this was the test that was going sideways, you can see I've forced a case with "bad" UTF-8 to be certain.

Note that there is no such guard on other tests that are generating random data. It's unlikely they all happen to be hitting the "sweet spot" all the times I've run them, but it's worth keeping in mind.

Similarly, there's a vanishingly small chance that two "users" we be generated with the same key, both here and in other tests. Would the test fail? Would we be able to diagnose what happened? I didn't think this was such a big deal that it was worth addressing in this PR.

poc/src/directory_host.rs
```diff
@@ -73,8 +73,8 @@ where
                         let toc = Instant::now() - tic;
                         let msg = format!(
                             "PUBLISHED '{}' = '{}' in {} s (epoch: {}, root hash: {})",
-                            a,
-                            b,
+                            String::from_utf8_lossy(&a),
+                            String::from_utf8_lossy(&b),
```
davidjattfb, line 77: As we drop into the POC, one can see I begin to use the "lossy" conversion from bytes. Since we're typing on the command line, using strings is more convenient for readability. However, if one does manage to fetch out an invalid string, this will do the thing where "bad" codepoints become those boxy question marks.

See also: https://doc.rust-lang.org/stable/std/string/struct.String.html#method.from_utf8_lossy

davidjattfb:
Hmm, clippy is running here with different params than I've used. It looks like it might be complaining about code outside my changes though?

slawlor:
> Hmm, clippy is running here with different params than I've used. It looks like it might be complaining about code outside my changes though?

Ah yeah you might need to update your version of Rust. The one in the CI is always the latest version. You can update your local rust install with ```rustup``` (it's happened to me as well)

akd/src/directory.rs
```diff
@@ -510,7 +510,7 @@ mod tests {
         let mut akd = Directory::<_>::new::<Blake3>(&db).await?;
 
         akd.publish::<Blake3>(
-            vec![(AkdLabel("hello".to_string()), AkdValue("world".to_string()))],
+            vec![(AkdLabel(Vec::from("hello")), AkdValue(Vec::from("world")))],
```
slawlor, line 513: It's tests so I don't really care, but generally will this give us utf8? We're not specifying any kind of encoding, so it'll be whatever this impl has chosen.

akd/src/storage/types.rs
slawlor: Will an empty slice still be valid? If not we should add a guard to protect against this (there must be SOMEthing for a user label)

poc/src/commands.rs
```diff
@@ -137,7 +137,7 @@ impl Command {
             return None;
         }
         let a = parts[1];
-        let cmd = DirectoryCommand::KeyHistory(String::from(a));
+        let cmd = DirectoryCommand::KeyHistory(Vec::from(a));
```
slawlor, line 140: Same question as before, are these all going to assume utf8? *Vocally thinking* is that was we want for user input? I _think_ so

slawlor: Actually on second thought, maybe it makes the most sense to keep the ```DirectoryCommand``` enum as strings, since that's all the user input, and this means the processing command needs to know encoding, etc. We should probably only do the conversion to bytes within the ```DirectoryProcessor```. But that's just a nit.

slawlor (approved):
Just some minor nits and comments

akd/src/directory.rs
davidjattfb: Yes: https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-From%3C%26%27_%20str%3E

Internally, the `From` implementation is calling `as_bytes` on the `&str` and then loading it like any other array.

akd/src/storage/types.rs
davidjattfb: Oh, huh. I lost the forest for the trees at some point, there's literally already a guard above on (new) line 80. 🥸

poc/src/commands.rs
davidjattfb: The latest commit steers `DirectoryCommand` back into the world of `String` objects, with the translation all happening at lower levels of the POC.

PR #131, opened 2022-01-25
Refactor storage interface and mysql implementation

afterdusk:
This PR makes two main changes:
- Move the struct builders (e.g. `build_azks`) from the `Storage` trait to `DbRecord`. This is so that `DbRecord` can build structs without knowledge of the specific `Storage` implementor.
- Move the `MySqlStorable` trait and impl into its own module. The idea is to separate the logic for the storage API and database handling (e.g. connection setup) from the struct handling logic (e.g. row deserialization). The separation is still not fully complete in this PR and some struct-specific logic is still left in the implementor of `Storage`, but this is a start.
 
Note: Since this PR changes logic and moves code across files, it'll be easier to review commit-by-commit rather than as a combined hunk.


slawlor:
Hey @afterdusk this looks great! Thanks :)

One thing, can you bump the version numbers of the akd and akd_mysql crates? I'll trigger a publish after this merges in. (Just saves a separate PR) 

Thanks!

PR #132, opened 2022-01-25
Adding test for empty tree hash

kevinlewi:
This simply adds a small test for checking that the root hash of an *empty* tree is equal to the same constant value (in this case, the hash of the NodeLabel (0,0) )


slawlor (approved):
LGTM, but seems clippy caught something!

PR #134, opened 2022-01-28
Adding full serde support for all proof structs

kevinlewi:
We want to support serialization for the proof structs, in particular: AppendOnlyProof, HistoryProof, and LookupProof. This enables the support (basically just had to add the `#[serde(bound = "")]` attribute to each struct), and also adds tests to ensure that they serialize/deserialize properly.

### DIFF
--- a/akd/src/node_state.rs
+++ b/akd/src/node_state.rs
@@ -20,7 +20,7 @@ use std::{
 use winter_crypto::Hasher;
 
 /// Represents a node's label & associated hash
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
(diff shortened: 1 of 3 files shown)


akd/src/node_state.rs
```diff
@@ -20,7 +20,7 @@ use std::{
 use winter_crypto::Hasher;
 
 /// Represents a node's label & associated hash
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
```
slawlor, line 23: <3

It's funny, I need this for a down-stream change anyways lol.

slawlor (approved):
L *great* TM! Thanks :)

PR #135, opened 2022-01-28
Future epochs provided don't throw errors

slawlor:
At the moment, giving a future epoch just silently truncates and returns the "latest" root hash, when it should return an error stating that the epoch is invalid. This PR adds an error to protect against this so it throws a ```HistroryTreeNode::NonexistantAtEpoch(root_node_label, latest_epoch)``` error.

### DIFF
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -439,6 +439,13 @@ impl Azks {
         storage: &S,
         epoch: u64,
     ) -> Result<H::Digest, HistoryTreeNodeError> {
+        if self.latest_epoch < epoch {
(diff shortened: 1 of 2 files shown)


afterdusk (approved):
Looks good, thanks for fixing this 🙂 There are other functions in this file that accept epoch as a parameter, should we add the guard for those too?

slawlor:
> Looks good, thanks for fixing this 🙂 There are other functions in this file that accept epoch as a parameter, should we add the guard for those too?

That's probably a good idea, unless they're always calling this common fn under the hood. I can check and apply the guards in a follow-up pr.

PR #137, opened 2022-01-31
Manual implementation of Clone

asonnino:
Manual implementation of Clone for all Proof structures to circumvent a known rust bug. See: https://github.com/rust-lang/rust/issues/41481

### DIFF
--- a/akd/src/proof_structs.rs
+++ b/akd/src/proof_structs.rs
@@ -41,7 +41,7 @@ impl<H: Hasher> Clone for MembershipProof<H> {
             hash_val: self.hash_val.clone(),
             parent_labels: self.parent_labels.clone(),
             siblings: self.siblings.clone(),
-            dirs: self.dirs.clone()
+            dirs: self.dirs.clone(),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


kevinlewi (commented):
LGTM, but will leave it to @slawlor for the final approval in case this conflicts with anything

akd/src/proof_structs.rs
```diff
@@ -33,9 +33,22 @@ pub struct MembershipProof<H: Hasher> {
     pub dirs: Vec<Direction>,
 }
 
+// Manual implementation of Clone, see: https://github.com/rust-lang/rust/issues/41481
+impl<H: Hasher> Clone for MembershipProof<H> {
+    fn clone(&self) -> Self {
+        Self {
+            label: self.label,
+            hash_val: self.hash_val,
```
slawlor, line 41: ```Hasher``` implements copy natively?

slawlor (approved):
Seems fine to me, just want to be sure about the types of ```Hasher```

akd/src/proof_structs.rs
asonnino: Not `Hasher`, but `Hasher::Digest` does implement copy

PR #138, opened 2022-02-02
AKD lean client, which is only dependent on the required hash functions

### LINKED ISSUE #114: Remove cargo bench from CI
Cargo bench was added in 51b756ca, however it is no longer in use. We should remove it from CI workflows.

### LINKED ISSUE #136: AKD client verification crate with limited dependencies
Due to the high likelihood that client's will verify the proofs with limited resources, it would be helpful to have a separate crate with as few dependencies as possible for verification of the proofs provided from the AKD. This would then simply be a dependency of the main AKD crate or could be maintained independently.

slawlor:
The AKD client currently living within the AKD crate is bloated with a lot of dependencies. That crate is useful for publishing and managing the directory structure, however if you're use-case is to verify the proof structures within a client-framework then this may cause too many useless dependencies to be imported presently, bloating binary size.

This PR adds a "lean" client which only depends on the core hash operation from the base crate, and implements the validation calls simply within itself. Additionally it supports static compilation targets via feature definitions within the crate (blake3 vs sha2-256 vs sha3-512 vs etc...).

Lastly it also adds support for a few WASM exported calls for verifying a lookup proof. The struct simply needs to be re-created via an incoming request operation.

Resolves #136 
Resolves #114


slawlor:
NOTE: Added the layer_proof notation to the MembershipProof to combine 3 dependent vecs into a vec of single structs which makes logical serialization/deserialization much easier down-stream.

akd_client/src/tests.rs
```diff
@@ -0,0 +1,173 @@
...
+                AkdValue("world2".to_string()),
+            ),
+        ],
+        false,
+    )
+    .await?;
+
+    // retrieve the lookup proof
+    let lookup_proof = akd.lookup(AkdLabel("hello".to_string())).await?;
+    // retireve the root hash
```
afterdusk, line 154: retrieve

akd_client/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+//!
+//! ## Present proof validation
+//!
+//! At the time of this documentation authoring, we presently support LookupProof verification
+//! without depending on the full AKD library.
+//!
+//! ## Planned future support
+//!
+//! Going forward this crate will re-implement the client verifications of the base crate, but with
+//! this "lean" mentality in mind. Should you not be running in a constrained then feel free to simply
```
afterdusk, line 23: constrained environment

akd_client/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+//! 1. **default** blake3: Blake3 256-bit hashing
+//! 2. sha256: SHA2 256-bit hashing
+//! 3. sha512: SHA3 512-bit hashing
+//! 4. sha3_256: SHA3 256-bit hashing
+//! 5. sha3_512: SHA3 512-bit hashing
+//!
+//! which dictate which hashing function is used by the verification components. Blake3 256-bit hashing is the default
+//! implementation and utilizes the [`blake3`] crate. Features sha256 and sha512 both utilize SHA2 cryptographic functions
+//! from the [`sha2`] crate. Lastly sha3_256 and sha3_512 features utilize the [`sha3`] crate for their hashing implementations.
+//! To utilizing a hash implementation other than blake3, you should compile with
```
afterdusk, line 39: to utilize

afterdusk (approved):
This is great! I'm not that familiar with the finer details around Rust builds and their optimization, but it looks good to me in all other aspects.

PR #140, opened 2022-02-03
Removing dangerous unwrap() calls within the library in non-test envs

### LINKED ISSUE #139: Remove calls to `unwrap()`
We need to strip the repo of any panicking calls, such as `unwrap`. This may also include changing return types for various messages.

slawlor:
In the library, there was a lot of use of ```unwrap()``` in dangerous places. This PR cleans this up

Resolves #139

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Installation
 Add the following line to the dependencies of your `Cargo.toml`:
 
 ```
-akd = "0.3"
+akd = "0.4"
 ```
 
 ### Minimum Supported Rust Version
--- a/akd_mysql/src/mysql.rs
+++ b/akd_mysql/src/mysql.rs
@@ -416,14 +416,18 @@ impl<'a> AsyncMySqlDatabase {
         let tic = Instant::now();
 
         let statement_text = record.set_statement();
+        let params = record
+            .set_params()
+            .ok_or_else(|| Error::Other("Failed to construct MySQL parameters block".into()))?;
+
         let out = match trans {
-            Some(mut tx) => match tx.exec_drop(statement_text, record.set_params()).await {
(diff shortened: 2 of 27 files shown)


README.md
```diff
@@ -23,7 +23,7 @@ Installation
 Add the following line to the dependencies of your `Cargo.toml`:
 
 ```
-akd = "0.3"
+akd = "0.4"
```
eozturk1, line 26: Have we bumped the version to 0.4?

akd_mysql/src/mysql.rs
```diff
@@ -952,82 +960,92 @@ impl Storage for AsyncMySqlDatabase {
...
-                                    DbRecord::get_multi_row_specific_params::<St>(batch).unwrap(),
-                                );
-                            }
-                        }
-
-                        // insert the batches of size = MYSQL_EXTENDED_INSERT_DEPTH
-                        if !params.is_empty() {
-                            let fill_statement = DbRecord::get_batch_fill_temp_table::<St>(Some(
-                                self.tunable_insert_depth,
+                let results = if let Some(create_table_cmd) =
```
eozturk1, line 963: Is it possible to rewrite this with an early else for better readability? i.e., 

```
if else_condition {
  return None;
}
// Proceed with valid results value
```

eozturk1 (approved):
LGTM. Added a comment about the `akd` version bump in the README and a readability question.

README.md
slawlor: Yes due to some unsightly contract changes between lib's, it was necessary to get the CI to not complain.

akd_mysql/src/mysql.rs
slawlor: Not really a great way, we'd lose timing information with an early exit (debug! prompts).

PR #141, opened 2022-02-04
NoSTD for lean client

slawlor:
Allowing no_std for the lean akd client validator. This allows us to not depend on the std library further minimizing dependencies of the application for embedded use. We do need the alloc usage, but that's the minimal amount needed (string formatting, vectors, etc)


afterdusk (approved):
Wow, now *that's* lean :)

PR #146, opened 2022-02-10
Adding AKD audit quorum general structure

slawlor:
This PR introduces an initial skeleton structure for what a AKD Quorum node can look like.

# Overview

Here we introduce the boiler-plate traits and structure for auditor quorum trait. This cluster of nodes independently verify the append-only nature of the AKD and jointly submit a signature & commitment that the AKD is valid.

The traits added here are
1. Unencrypted, basic storage layer: ```storage.rs```
2. Communication layer between nodes in the quorum & externally with the quorum: ```comms/mod.rs```
3. Inter-node message replay protection via nonce's and nonce management.: ```comms/nonces.rs```
4. Cryptographic operations which the membership need to perform: ```crypto/mod.rs```

Additionally since we'll be using protobuf for the messages sent between nodes, I've also included a ```build.rs``` which will automate code generation for protobuf specs as needed.

Addresses: #125 

NOTE: This PR is based off of #126 which was a work-in-progress PR. Subsequent PRs will follow with the inner node logic and message structures

### DIFF
--- a/akd_audit_quorum/src/comms/nonces.rs
+++ b/akd_audit_quorum/src/comms/nonces.rs
@@ -24,6 +24,12 @@ pub struct NonceManager {
     incoming: Arc<RwLock<HashMap<NodeId, Nonce>>>,
 }
 
+impl Default for NonceManager {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl NonceManager {
     /// Create a new nonce manager instance
     pub fn new() -> Self {
--- a/akd_audit_quorum/src/comms/nonces.rs
+++ b/akd_audit_quorum/src/comms/nonces.rs
@@ -74,6 +74,13 @@ impl NonceManager {
                     from, nonce, expected
                 ),
             );
+
+            // reject this message but additionaly reset the nonce to what was received + 1. This handles when
+            // nodes restart or get corrupted and nonces may legitimately drift apart. In the event
+            // that a reply attack occurs, we're rejecting the message anyways if the nonce doesn't match
+            // what's expected so it's still protected.
+            self.incoming.write().await.insert(from, nonce + 1);
+
             Err(err)
         }
     }
--- a/akd_audit_quorum/src/comms/nonces.rs
+++ b/akd_audit_quorum/src/comms/nonces.rs
@@ -10,9 +10,15 @@
 //! The purpose of managing nonces is to protect against replay attacks in inter-node
 //! messages. The [`NonceManager`] manages both _expected_ incoming nonces from nodes
 //! and the outgoing nonces for messages being sent to nodes.
+//!
+//! ## Relevant information
+//!
+//! The nonce reset problem
+//! https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.1061.3009&rep=rep1&type=pdf
 
 use super::{NodeId, Nonce};
 
+use rand::Rng;
 use std::collections::HashMap;
 use std::sync::Arc;
 use tokio::sync::RwLock;
@@ -39,13 +45,43 @@ impl NonceManager {
         }
     }
 
-    /// Retrieve the "next" nonce for message send, will auto-increment the nonce
-    /// mapping
-    pub async fn get_next_outgoing_nonce(&self, to: NodeId) -> Nonce {
+    /// Reset the nonce for a given peer node to a random value
+    /// and return the "next" expected nonce for that peer which will
+    /// be communicated to that peer for setting their nonce options
+    pub async fn reset(&self, from: NodeId) -> Nonce {
+        let mut guard = self.incoming.write().await;
+
+        let mut rng = rand::thread_rng();
+        let new_nonce: Nonce = rng.gen();
+
+        // we'll be expecting new_nonce + 1, but we transmit new_nonce as
+        // the peer's logic will auto-increment it when generating an
+        // outgoing nonce
+        guard.insert(from, new_nonce + 1);
+
+        new_nonce
+    }
+
+    /// Set the outgoing nonce for a given peer, this is in reply to a remote nonce
+    /// reset request
+    pub async fn set_outgoing_nonce_for_peer(&self, to: NodeId, starting_nonce: Nonce) {
         let mut guard = self.outgoing.write().await;
-        let next = guard.get(&to).map_or(0, |a| *a + 1);
-        // safe the updated nonce
-        guard.insert(to, next);
+        guard.insert(to, starting_nonce);
+    }
+
+    /// Retrieve the "next" nonce for an outgoing message. Will auto-increment the nonce
+    /// mapping
+    ///
+    /// Returns: Some(nonce) when a valid next nonce was able to be retrieved, None if
+    /// no starting/previous nonce was registered for this peer
+    pub async fn get_next_outgoing_nonce(&self, to: NodeId) -> Option<Nonce> {
+        let mut guard = self.outgoing.write().await;
+        let next = guard.get(&to).map(|a| *a + 1);
+
+        if let Some(v) = &next {
+            // save the next expected nonce
+            guard.insert(to, *v);
+        }
         next
     }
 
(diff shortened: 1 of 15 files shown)


akd_audit_quorum/src/comms/nonces.rs
```diff
@@ -0,0 +1,102 @@
...
+                    "Nonce mismatch in raft inter-messages: Node {}, Nonce: {}, Expected Nonce: {}",
+                    from, nonce, expected
+                ),
+            );
+
+            // reject this message but additionaly reset the nonce to what was received + 1. This handles when
+            // nodes restart or get corrupted and nonces may legitimately drift apart. In the event
+            // that a reply attack occurs, we're rejecting the message anyways if the nonce doesn't match
+            // what's expected so it's still protected.
+            self.incoming.write().await.insert(from, nonce + 1);
```
alexme22, line 82: Wouldn't that allow replay attacks repeating any two consecutive requests from any node? Ex. of requests to replay: A {from: node, nonce = X}, B {from: node, nonce = X + 1} for arbitrary X. A will be rejected, and B accepted.

Thinking on the solutions, this node1 may store max_nonce of all seen nonces from node2, and allow requests with incoming_nonce > max_nonce updating max_nonce after that (it also ignores packet loses or delays from node2).

The open question here is what to do on a node restart - nonces should be synced. I'd guess some kind of authentication happens when new connection between nodes is established, and nonces could be reset if they linked to current session. That will eliminate replies from the same node with higher nonces but previous session.

slawlor: Hmm yes you're right. We can definitely guarantee they're increasing only, but we do need a reset mechanism somehow. Looking into what's commonly done, it seems generating a random nonce is often the solution [link](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.1061.3009&rep=rep1&type=pdf) but i'm not sure how we'd sync it between 2 nodes.

slawlor: If we're only adding support for replay attacks here, and we're assuming the channel isn't "comprisable" what if we add a "ResetNonce" message which would reset the max nonce to 0. A node comes online, and broadcasts to all peers that it wants to reset it nonces with them. (Assuming some layer of node inter-authentication). In the ResetNonce, the peer would generate a random starting point, which would protect against replaying messages from (0).

alexme22: ResetNonce could be replayed too and may disrupt all communication in the quorum. My suggestion is to take into account environment/type of connections where quorum will be run and safety guarantees it provides and decide if this kind of mechanism is required on application level.

akd_audit_quorum/src/crypto/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+
+    /// Get the number of shards required to reconstruct the quorum key
+    fn shards_required(n: u8) -> u8 {
+        let f = if (n - 1) % 3 == 0 {
+            (n - 1) / 3
+        } else {
+            // there's remainders
+            1 + (n - 1) / 3
+        };
+        2 * f + 1
```
alexme22, line 240: Could you please explain/add a comment to this formula?

akd_audit_quorum/src/crypto/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+// =====================================================
+
+/// The multiplicitave factor of DATA_SIZE which denotes the size of the
+/// quorum key. Probably should be a factor of 2
+pub(crate) const QUORUM_KEY_NUM_PARTS: usize = 8;
+
+/// The size of the quorum key private key in bytes.
+/// NOTE: SSS's DATA_SIZE = 64 bytes, which the quorum key private key
+/// need to be a multiple of
+pub const QUORUM_KEY_SIZE: usize = QUORUM_KEY_NUM_PARTS * DATA_SIZE;
```
alexme22, line 50: I'd add dimension: QUORUM_KEY_SIZE_BYTES

akd_audit_quorum/src/crypto/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+        commitment: QuorumCommitment<H>,
+    ) -> Result<bool, QuorumOperationError>;
+
+    // ==================================================================
+    // Common trait logic
+    // ==================================================================
+
+    /// Get the number of shards required to reconstruct the quorum key
+    fn shards_required(n: u8) -> u8 {
+        let f = if (n - 1) % 3 == 0 {
```
alexme22, line 234: let's add some input validation, like n >= 3 ?

akd_audit_quorum/src/crypto/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+        for i in 0..QUORUM_KEY_NUM_PARTS {
+            let part: [u8; DATA_SIZE] = quorum_key[i * DATA_SIZE..(i + 1) * DATA_SIZE]
+                .try_into()
+                .map_err(|_| {
+                QuorumOperationError::Sharding(format!(
+                    "Unable to convert quorum key slice into SSS shardable component of len {}",
+                    DATA_SIZE
+                ))
+            })?;
+            let results = create_shares(&part, num_shards, num_approvals)?;
```
alexme22, line 263: How about to add immediate verification of shares?

let restored = combine_shares(&results).unwrap();
assert_eq!(restored, Some(part));

akd_audit_quorum/src/crypto/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    /// Generate num_shards shards of the quorum key, and return the shard pieces.
+    /// We take ownership of the quorum key here to help prevent leakage of the key.
+    /// By taking ownership, someone needs to explicitely clone it to use it elsewhere
+    fn generate_shards(
+        quorum_key: [u8; QUORUM_KEY_SIZE],
+        num_shards: u8,
+    ) -> Result<Vec<QuorumKeyShard>, QuorumOperationError> {
+        let num_approvals = Self::shards_required(num_shards);
```
alexme22, line 250: Please add some num_shards validation like > 3 or something?

akd_audit_quorum/src/crypto/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+
+        let mut parts = vec![vec![]; num_shards.into()];
```
alexme22, line 252: Is `shards` a better name instead of `parts`? A part below is split into num_shard shards and this array stores all i-th shards in its [i] element.

akd_audit_quorum/src/crypto/mod.rs
slawlor: Maybe something along those lines, but we generally stay away from unwrap and asserts outside of test env's as they can cause a ```panic!``` in prod code, which will be unhandled.

alexme22: I'd better validate key could be restored and panic! if not rather than distribute corrupted parts.

akd_audit_quorum/src/crypto/mod.rs
alexme22: It currently returns 0 for n = 1 - is this expected? Do we even need SSS for that case?
I'd expect SSS to be applicable for 2 machines at min.

akd_audit_quorum/src/crypto/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+}
+
+/// Represents an encrypted quorum key shard which is encrypted with
+/// the public key of the reliant party. These shards cannot be decoded in
+/// user-space, and must be handled WITHIN the crypto layer such that they
+/// public key cannot be reconstructed by "non secure" memory. The private key
+/// which can decrpyt the shards is the node's "transient" key used for communication
+/// channels, and never is exposed outside of the cryptographer service
+#[derive(Clone)]
+pub struct EncryptedQuorumKeyShard {
```
kevinlewi, line 148: An alternative to this sharding approach would be to use the frost-dalek library: https://docs.rs/frost-dalek/0.2.3/frost_dalek/

The advantage being that no single quorum participant needs to be responsible for fully constructing the signing key. Instead, each quorum participant uses their key shard to produce a "signature shard", and then anyone can aggregate the signature shards together to compose the actual signature.

Thoughts on integrating frost-dalek into here?

slawlor: We're in the process of resurrecting this avenue. We'll definitely look into this usage! We just need to assess growth & shrinkage of the quorum over time, but if there's a solution there it's an attractive option! Thanks for the pointer :)

kevinlewi:
We can reopen this if we revisit it in the future

Issue #149, opened 2022-02-19
Option to initialize AKD without write ability


afterdusk:
Currently, when AKD is initialized, it will fetch the `Azks` record in storage. If it is unable to find the record, it will attempt to write the root node (and its epoch 0 state) along with the `Azks` record. 

In a distributed deployment with dedicated readers and writers, we do not want the readers to perform this write on startup. Instead, they should fail on being unable to retrieve the `Azks` record.

Issue #148, opened 2022-02-11
Invalidate cache on new Azks epoch


afterdusk:
In a distributed deployment, the cache may contain stale records. We want the cache to be invalidated when a new `Azks` record with a higher epoch number is detected, so that stale `HistoryTreeNodes` are not used to generate proofs. This involves remembering the epoch of the `Azks` record in the cache, and at TTL expiration checking it against the newly fetched `Azks` record to see if the epoch number changed. If it has, the entire cache should be dumped.

PR #150, opened 2022-02-22
Add the ability to periodically poll for changes to the AZKS record and invalidate cache

slawlor:
This change adds support to poll for changes direct from the storage layer of the AZKS record, and when a change (epoch) is detected, flushes the object cache so all records get retrieved from storage.

Additionally add support for a read-only directory which cannot perform "write" operations to the directory

Resolves: #149, #148

### DIFF
--- a/akd/src/directory.rs
+++ b/akd/src/directory.rs
@@ -169,7 +169,7 @@ impl<S: Storage + Sync + Send, V: AkdVRF> Directory<S, V> {
                 }
             }
         }
-        let insertion_set: Vec<Node<H>> = update_set.iter().copied().collect();
+        let insertion_set: Vec<Node<H>> = update_set.to_vec();
 
         if use_transaction {
             if let false = self.storage.begin_transaction().await {
--- a/akd/src/directory.rs
+++ b/akd/src/directory.rs
@@ -10,10 +10,12 @@
 use crate::append_only_zks::Azks;
 
 use crate::node_state::{Node, NodeLabel};
+use crate::primitives::akd_vrf::AkdVRF;
 use crate::proof_structs::*;
 
 use crate::errors::{AkdError, DirectoryError, HistoryTreeNodeError, StorageError};
 
+use crate::serialization::from_digest;
 use crate::storage::types::{AkdLabel, AkdValue, DbRecord, ValueState, ValueStateRetrievalFlag};
 use crate::storage::Storage;
 
@@ -23,7 +25,6 @@ use rand::{CryptoRng, RngCore};
 use std::collections::HashMap;
 use std::marker::{Send, Sync};
 use std::sync::Arc;
-use winter_crypto::Digest;
 use winter_crypto::Hasher;
 
 /// Root hash of the tree and its associated epoch
@@ -45,34 +46,26 @@ impl AkdLabel {
 }
 
 /// The representation of a auditable key directory
-pub struct Directory<S> {
+#[derive(Clone)]
+pub struct Directory<S, V> {
     storage: S,
+    vrf: V,
     read_only: bool,
     /// The cache lock guarantees that the cache is not
     /// flushed mid-proof generation. We allow multiple proof generations
     /// to occur (RwLock.read() operations can have multiple) but we want
     /// to make sure no generations are underway when a cache flush occurs
     /// (in this case we do utilize the write() lock which can only occur 1
-    /// at a time).
+    /// at a time and gates further read() locks being acquired during write()).
     cache_lock: Arc<tokio::sync::RwLock<()>>,
 }
 
-impl<S: Storage + Sync + Send> Clone for Directory<S> {
-    fn clone(&self) -> Self {
-        Self {
-            storage: self.storage.clone(),
-            read_only: self.read_only,
-            cache_lock: self.cache_lock.clone(),
-        }
-    }
-}
-
-impl<S: Storage + Sync + Send> Directory<S> {
+impl<S: Storage + Sync + Send, V: AkdVRF> Directory<S, V> {
     /// Creates a new (stateless) instance of a auditable key directory.
     /// Takes as input a pointer to the storage being used for this instance.
     /// The state is stored in the storage.
-    pub async fn new<H: Hasher>(storage: &S, read_only: bool) -> Result<Self, AkdError> {
-        let azks = Directory::get_azks_from_storage(storage, false).await;
+    pub async fn new<H: Hasher>(storage: &S, vrf: &V, read_only: bool) -> Result<Self, AkdError> {
+        let azks = Directory::<S, V>::get_azks_from_storage(storage, false).await;
 
         if read_only && azks.is_err() {
             return Err(AkdError::Directory(DirectoryError::Storage(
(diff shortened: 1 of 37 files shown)


afterdusk (commented):
Hmm, I've some concerns about the poller being managed on a separate thread and manually initialized by the library client. A reader could see a new `Azks` epoch from storage, but the poller may not have had a chance to flush the cache. The reader could end up constructing broken proofs if it mixes data from the old epoch in the cache with records from the new epoch.

Furthermore, it might also be bad for perf if the poller flushes the cache while the user is in the middle of constructing a proof, especially if a new publish recently happened. AKD might see newer `HistoryTreeNodes`, which it can self-correct for but will require another round trip to storage.

We might need to find a way to synchronize cache flushes to be between Directory operations and prevent a new `Azks` from being seen without an immediate cache flush. I'm still not clear on how that can be achieved...

akd_mysql/src/mysql.rs
```diff
@@ -870,6 +870,13 @@ impl Storage for AsyncMySqlDatabase {
         }
 
         // cache miss, log a real sql read op
+        self.get_direct::<St>(id).await
```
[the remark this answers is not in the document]
slawlor: Ah that's my slip-up. ```get_direct``` should not alter the cache at all, simply returning the result, and then if ```get``` wants, it can cache the object. Callers of ```get_direct``` should not play with the cache in any way.

slawlor:
> Furthermore, it might also be bad for perf if the poller flushes the cache while the user is in the middle of constructing a proof, especially if a new publish recently happened. AKD might see newer `HistoryTreeNodes`, which it can self-correct for but will require another round trip to storage.

@afterdusk So I believe this is now resolved with my last commit(s). We now are utilizing a tokio `RwLock` to prevent flushing the cache while there are presently proof generations underway.

It's a little abnormal of a use-case, but tokio's `RwLock` already supports "multiple readers, single writer". This uniquely fits our use-case here, since we want the ability to generate proofs to be easily parallelizable, sharing a cache, and to only flush the cache when no proof generations are underway. The `RwLock` will wait for the reads underway to complete and then block future reads until it's write (cache flush) is completed.

Additionally, we now cache the AZKS element forever, which is only necessary in a read-only mode (a publish will update the AZKS element for obvious reasons). However if you launch a read-only node, you can `poll_for_azks_changes` for azks changes on a given frequency, and when a change occurs a notification is passed back (for logging and observability) and the cache is write-lock acquired and flushed. Additionally added a test covering this case.

akd/src/directory.rs
```diff
@@ -276,24 +303,42 @@ impl<S: Storage + Sync + Send> Directory<S> {
...
+                    let _guard = self.cache_lock.write().await;
+                    // flush the cache in its entirety
+                    self.storage.flush_cache().await;
+                    // re-fetch the azks to load it into cache so when we release the cache lock
+                    // others will see the new AZKS loaded up and ready
+                    last = Directory::get_azks_from_storage(&self.storage, false).await?;
+
+                    // notify change occurred
+                    if let Some(channel) = &change_detected {
+                        channel.send(()).await.map_err(|send_err| AkdError::Directory(DirectoryError::Storage(StorageError::Connection(format!("Tokio MPSC sender failed to publish notification with error {:?}", send_err)))))?;
```
afterdusk, line 338: nit: multiple lines for readability

akd/src/directory.rs
```diff
@@ -257,13 +302,60 @@ impl<S: Storage + Sync + Send> Directory<S> {
...
+            if latest.latest_epoch > last.latest_epoch {
+                {
+                    // acquire a singleton lock prior to flushing the cache to assert that no
+                    // cache accesses are underway (i.e. publish/proof generations/etc)
+                    let _guard = self.cache_lock.write().await;
+                    // flush the cache in its entirety
+                    self.storage.flush_cache().await;
+                    // re-fetch the azks to load it into cache so when we release the cache lock
+                    // others will see the new AZKS loaded up and ready
+                    last = Directory::get_azks_from_storage(&self.storage, false).await?;
```
afterdusk, line 334: Looks like we have to rely on the storage layer to properly implement cache behavior (in this case remembering to insert into the cache after fetching). I think a rewrite of the storage interface might be necessary in the longer term.

afterdusk (approved):
LGTM, the test is great! @kevinlewi  you might want to review this as well, since it deals with the external interface. Thanks for resolving this @slawlor !

akd/src/directory.rs
slawlor: Unfortunately this is rustfmt. I could extract to a function or wrap in curly braces, but Rust chose the linewrapping here

akd/src/directory.rs
slawlor: So yes and maybe no lol. The storage layer does indeed need to cache, if they want to utilize a cache, however it could also have no cache at all and proxy everything to the storage layer. We've implemented a [`TimedCache`] struct but one could either use their own thing or optionally no cache at all. We could make the storage trait have the default caching logic embedded and just like "native" access calls be the ones to implement by the trait directly...

That could be a future option.

akd/src/directory.rs
```diff
@@ -50,17 +50,30 @@ impl AkdLabel {
...
+    pub async fn new<H: Hasher>(storage: &S, vrf: &V, read_only: bool) -> Result<Self, AkdError> {
+        let azks = Directory::<S, V>::get_azks_from_storage(storage, false).await;
+
+        if read_only && azks.is_err() {
+            return Err(AkdError::Directory(DirectoryError::Storage(
+                StorageError::GetData(
+                    "AZKS record not found and Directory constructed in read-only mode".to_string(),
+                ),
+            )));
+        } else if azks.is_err() {
```
alexme22, line 76: What do you think on expecting not_found error here instead of any error? I wonder could azks exist but get_azks_from_storage() having transient error?

akd/src/directory.rs
```diff
@@ -275,13 +307,60 @@ impl<S: Storage + Sync + Send, V: AkdVRF> Directory<S, V> {
...
+    ) -> Result<(), AkdError> {
+        // Retrieve the same AZKS that all the other calls see (i.e. the version that could be cached
+        // at this point). We'll compare this via an uncached call when a change is notified
+        let mut last = Directory::<S, V>::get_azks_from_storage(&self.storage, false).await?;
+
+        loop {
+            // loop forever polling for changes
+            tokio::time::sleep(period).await;
+
+            let latest = Directory::<S, V>::get_azks_from_storage(&self.storage, true).await?;
```
alexme22, line 329: Maybe to ignore/log errors, or retry couple of times here and below instead of crashing (if transient error)?

slawlor: Bubbling up the error just means the caller of this function needs to handle it. We could handle the error handling internally, but there might be non-transient errors which we want passed up. It just makes the inner handling logic more complicated.

If it were to fail, the caller can catch, log/print/etc whatever they want, and re-call the function to continue polling.

akd/src/directory.rs
slawlor: The caller should handle all transient error logic if they want to

PR #151, opened 2022-02-24
Return `Result<bool, JsValue>` for WASM lookup proof

slawlor:
This PR supports adding a `Result<bool, JsValue>` to capture error messages & codes when a verification fails (rather than just panic-ing).

Additionally exposes some different types which are cleaner than all `JsValue`s


afterdusk (approved):
LGTM!

PR #152, opened 2022-02-25
Removing examples/measurements.rs

kevinlewi:
As title. We used to use measurements.rs to do some simple benchmarking, but this file is now pretty out of date, and besides, not compiling with the current main branch. It's also not CI-tested.


slawlor:
There's a clippy warning which is not being caught locally (Are we running unstable in the CI?). Anyways if you rebase on /main I fixed it in another PR that just merged, so it should go away. LGTM!

slawlor (approved):

kevinlewi:
Thanks! Ah sorry, I realized that my clippy was out of date, but after running `rustup update`, this clippy error is indeed showing for me locally.

PR #153, opened 2022-02-25
Replace [`vrf`] crate with pure Rust impl for use in WASM assembly

slawlor:
This PR replaces the [`vrf`] crate which depends on OpenSSL with a more pure-Rust implementation. This is necessary to 

1. Lean out the code in it's dependency tree
2. OpenSSL doesn't compile to WASM (WebAssembly)

Due to an [error with the MAC version of clippy crashing](https://github.com/rust-lang/rust-clippy/issues/8470), I am unable to verify clippy warnings so they may be flagged in this PR.

Additionally there are changes related to the way dependencies are handled in the core AKD crate. We're moving more towards conditional compilation with `Cargo.toml` enabled features rather than having all dependencies required all the time.

For example, from the updated docs

The akd crate supports multiple compilation features

[serde](https://docs.rs/serde/latest/serde/): Will enable serde serialization support on all public structs used in storage & transmission operations. This is helpful in the event you wish to directly serialize the structures to transmit between library <-> storage layer or library <-> clients. If you’re also utilizing VRFs (see (2.) below) it will additionally enable the serde feature in the ed25519-dalek crate.

vrf (on by-default): Will enable support of verifiable random function (VRF) usage within the library. See [ecvrf.rs] for documentation about the VRF functionality being utilized within AKD. This functionality is added protection so auditors don’t see user identifiers directly and applies a level of user-randomness (think hashing) in the node labels such that clients cannot trivially generate node labels themselves for given identifiers, however they can verify that a label is valid for a given identitifier. Transitively will add dependencies on crates [curve25519-dalek] and [ed25519-dalek]. You can disable the VRF functionality by adding the no-default-features flags to your cargo dependencies.

public-tests: Will expose some internal sanity testing functionality, which is often helpful so you don’t have to write all your own unit test cases when implementing a storage layer yourself. This helps guarantee the sanity of a given storage implementation. Should be used only in unit testing scenarios by altering your Cargo.toml as such


kevinlewi:
Nice!!

kevinlewi (approved):

slawlor:
Blocked on #155 fixing the error in the MySQL tests infra

akd_client/src/ecvrf.rs
```diff
@@ -165,8 +164,9 @@ impl VRFPublicKey {
         self.verify(&proof, &message)?;
 
         let output: Output = (&proof).into();
-        let mut expected_truncated_hash: [u8; 32] = [0u8; 32];
-        expected_truncated_hash.copy_from_slice(&output.to_bytes()[..32]);
+        // truncate the 64-byte hash to first 32 bytes
```
eozturk1: I think this was discussed in the chat but let's add a comment here why this is okay.

akd/src/ecvrf/mod.rs
```diff
@@ -5,16 +5,35 @@
 // License, Version 2.0 found in the LICENSE-APACHE file in the root directory
 // of this source tree.
 
+//! This module contains implementations of a
+//! [verifiable random function](https://en.wikipedia.org/wiki/Verifiable_random_function)
+//! (currently only ECVRF). VRFs can be used in the consensus protocol for leader election
```
eozturk1, line 10: Do we have plans to use VRFs for leader election? Is there a doc?

akd/src/ecvrf/no_vrf.rs
```diff
@@ -0,0 +1,123 @@
...
+
+impl Proof {
+    /// Converts a Proof into bytes
+    pub fn to_bytes(&self) -> [u8; 80] {
+        [0u8; 80]
+    }
+}
+
+/// Represents a secure storage of the VRF private key. Since the VRF private key
+/// should change never (if it does, the entire tree is no longer a consistent mapping
```
eozturk1, line 50: At some point we said we could embed this private key in the code. From this sentence, it sounds like that's something we shouldn't do. Maybe we should consider alternatives.

akd/src/ecvrf/no_vrf.rs
```diff
@@ -0,0 +1,123 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+//
+// This source code is licensed under both the MIT license found in the
+// LICENSE-MIT file in the root directory of this source tree and the Apache
+// License, Version 2.0 found in the LICENSE-APACHE file in the root directory
+// of this source tree.
+
+//! This module contains mock usage of the VRF functionality should someone choose they
```
eozturk1, line 8: If we are not using it, let's consider adding it in another PR. It also has incomplete implementations.

akd/src/lib.rs
```diff
@@ -277,6 +277,31 @@
 //! };
 //! ```
 //!
+//! # Compilation Features
```
eozturk1, line 280: Great summary, thank you!

eozturk1 (commented):
Great work bringing in the EC VRF implementation @slawlor! :)

akd/src/ecvrf/mod.rs
```diff
@@ -0,0 +1,363 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
```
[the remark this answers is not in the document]
slawlor: So that whole folder has been deleted, so I'm guessing that they aren't using VRFs anywhere anymore. I think we're on our own sync-wise, but this does follow the spec pretty clearly so unless the RFC changes we should be good.

akd/src/ecvrf/mod.rs
slawlor: We don't, that's the same comment from the Diem repo. I'll clean it up since we're not using it for that, it's more informational anyways.

akd/src/ecvrf/no_vrf.rs
slawlor: We aren't in the end. I spoke to @kevinlewi about it async, and it's secret enough it should go into a secure storage medium. The "HardCodedVRFKey" is only used in testing, and someone should properly implement the trait to a secure storage layer for the private key.

akd/src/ecvrf/no_vrf.rs
slawlor: It's a crate-level feature, if someone does NOT want VRF functionality as it's a computation hit. I'll enable the test suite without VRFs in the CI pipeline, but the implementation _is_ complete. It's supposed to return an empty private key struct, essentially saying "there is no VRF key, and everything computed will be without VRFs"

akd/src/ecvrf/mod.rs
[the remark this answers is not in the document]
eozturk1: Apparently they were removed due to lack of use. Great find!

PR #154, opened 2022-02-25
Fix silently failing mysql tests

eozturk1:
MySQL tests were silently failing due to output parser for docker container list expecting incorrect number of lines.

This PR fixes the issue. MySQL tests should run OK now.


slawlor (approved):
Tests will fail but it's unrelated to this. Our trunk is broken at this point

Issue #144, opened 2022-02-09
Verification failure with small trees


afterdusk:
Lookup and key history verifications encounter failures with small trees. This manifests with the `Blake3` Hasher as well as with `Sha3_256`.

Repros:
- [SHA3_256, Lookup](https://github.com/novifinancial/akd/blob/997826961eb7c036ed2c3e489be78dccbad063d9/akd_client/src/tests.rs#L183-L224)
- [Blake3, KeyHistory](https://github.com/afterdusk/akd/commit/20801823496a4c816470f5237b70dd0f8e5a28d8)

slawlor:
Resolved with #155

slawlor:
We're still seeing this issue with small trees (e.g. 2 nodes). The following is a minimal repro test case (commented out in directory.rs)

```rust
// Test coverage on issue #144, verification failures with small trees (<4 nodes)
    #[tokio::test]
    async fn test_simple_lookup_for_small_tree() -> Result<(), AkdError> {
        let db = AsyncInMemoryDatabase::new();
        let vrf = HardCodedAkdVRF {};
        // epoch 0
        let akd = Directory::<_, _>::new::<Blake3>(&db, &vrf, false).await?;

        let mut updates = vec![];
        for i in 0..1 {
            updates.push((
                AkdLabel(format!("hello{}", i).as_bytes().to_vec()),
                AkdValue(format!("hello{}", i).as_bytes().to_vec()),
            ));
        }

        akd.publish::<Blake3>(updates).await?;

        let target_label = AkdLabel(format!("hello{}", 0).as_bytes().to_vec());

        // retrieve the lookup proof
        let lookup_proof = akd.lookup(target_label.clone()).await?;
        // retrieve the root hash
        let current_azks = akd.retrieve_current_azks().await?;
        let root_hash = akd.get_root_hash::<Blake3>(&current_azks).await?;

        let vrf_pk = vrf.get_vrf_public_key().await?;

        // perform the "traditional" AKD verification
        let akd_result = crate::client::lookup_verify::<Blake3>(
            &vrf_pk,
            root_hash,
            target_label.clone(),
            lookup_proof,
        );

        // check the two results to make sure they both verify
        assert!(matches!(akd_result, Ok(())));

        Ok(())
    }
 ```

eozturk1:
I have added this fix to the [TODO](https://github.com/novifinancial/akd/pull/189#issuecomment-1135376243) list for the optimized AZKS. cc @Jasleen1

[...]

PR #155, opened 2022-02-27
Fixed small tree verification issue

Jasleen1:
Adresses #144

### DIFF
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -86,6 +86,7 @@ impl Azks {
         Ok(azks)
     }
 
+
     /// Inserts a single leaf and is only used for testing, since batching is more efficient.
     /// We just want to make sure batch insertions work correctly and this function is useful for that.
     pub async fn insert_leaf<S: Storage + Sync + Send, H: Hasher>(
@@ -315,14 +316,14 @@ impl Azks {
         let longest_prefix = lcp_node.label;
         // load with placeholder nodes, to be replaced in the loop below
         let mut longest_prefix_children = [Node {
-            label: NodeLabel::root(),
-            hash: crate::utils::empty_node_hash::<H>(),
+            label: EMPTY_LABEL,
+            hash: crate::utils::empty_node_hash_no_label::<H>(),
         }; ARITY];
         let state = lcp_node.get_state_at_epoch(storage, epoch).await?;
-
         for (i, child) in state.child_states.iter().enumerate() {
             match child {
                 None => {
+                    println!("i = {}, empty", i);
                     continue;
                 }
                 Some(child) => {
@@ -332,6 +333,7 @@ impl Azks {
                         self.get_latest_epoch(),
                     )
                     .await?;
+                    println!("Label of child {} is {:?}", i, unwrapped_child.label);
                     longest_prefix_children[i] = Node {
                         label: unwrapped_child.label,
                         hash: unwrapped_child
@@ -341,6 +343,7 @@ impl Azks {
                 }
             }
         }
+        println!("Lcp label = {:?}", longest_prefix);
         Ok(NonMembershipProof {
             label,
             longest_prefix,
@@ -521,8 +524,8 @@ impl Azks {
             prev_node = curr_node.label;
             let curr_state = curr_node.get_state_at_epoch(storage, epoch).await?;
             let mut nodes = [Node::<H> {
-                label: NodeLabel::root(),
-                hash: H::hash(&[0u8]),
+                label: EMPTY_LABEL,
+                hash: H::hash(&EMPTY_VALUE),
             }; ARITY - 1];
             let mut count = 0;
             let direction = dir.ok_or({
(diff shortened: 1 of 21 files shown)


akd/src/append_only_zks.rs
```diff
@@ -315,14 +315,14 @@ impl Azks {
...
-            hash: crate::utils::empty_node_hash::<H>(),
+            label: EMPTY_LABEL,
+            hash: crate::utils::empty_node_hash_no_label::<H>(),
         }; ARITY];
         let state = lcp_node.get_state_at_epoch(storage, epoch).await?;
-
         for (i, child) in state.child_states.iter().enumerate() {
             match child {
                 None => {
+                    println!("i = {}, empty", i);
```
slawlor, line 325: Please stick to `debug!` or `warn!` etc traces. They do proper formatting, and can redirect to other outputs (e.g. file writing) if needed. If this was just for debugging it should get removed before being landed. Same goes for the other println! statements below

akd/src/append_only_zks.rs
```diff
@@ -332,6 +332,7 @@ impl Azks {
                         self.get_latest_epoch(),
                     )
                     .await?;
+                    println!("Label of child {} is {:?}", i, unwrapped_child.label);
```
slawlor, line 335: here

akd/src/append_only_zks.rs
```diff
@@ -341,6 +342,7 @@ impl Azks {
                 }
             }
         }
+        println!("Lcp label = {:?}", longest_prefix);
```
slawlor, line 345: here

akd/src/client.rs
```diff
@@ -61,15 +61,28 @@ pub fn verify_nonmembership<H: Hasher>(
...
+    println!("Lcp real = {:?}", lcp_real);
     for i in 0..ARITY {
         let child_hash = H::merge(&[
             proof.longest_prefix_children[i].hash,
             hash_label::<H>(proof.longest_prefix_children[i].label),
         ]);
         lcp_hash = H::merge(&[lcp_hash, child_hash]);
-        lcp_real = lcp_real.get_longest_common_prefix(proof.longest_prefix_children[i].label);
+        let curr_label = proof.longest_prefix_children[i].label;
+        // if curr_label != EMPTY_LABEL {
```
slawlor, line 77: Is this check necessary or was just a debugging statement?

akd/src/directory.rs
```diff
@@ -666,7 +666,7 @@ mod tests {
         auditor::audit_verify,
         // BUG: Key history verification is failing on small-sized trees. It's being addressed in Issue #144
```
slawlor, line 667: If we're really solving this issue, we should remove notes to Issue #144 in the code comments

akd/src/directory.rs
```diff
@@ -681,10 +681,12 @@ mod tests {
     async fn test_empty_tree_root_hash() -> Result<(), AkdError> {
         let db = AsyncInMemoryDatabase::new();
         let vrf = HardCodedAkdVRF {};
-        let akd = Directory::<_, _>::new::<Blake3>(&db, &vrf, false).await?;
+        let akd = Directory::<_, _>::new::<Blake3_256<BaseElement>>(&db, &vrf, false).await?;
```
slawlor, line 684: nit: Why? `Blake3` is much more concise

akd/src/directory.rs
```diff
@@ -681,10 +681,12 @@ mod tests {
...
 
         let current_azks = akd.retrieve_current_azks().await?;
-        let hash = akd.get_root_hash::<Blake3>(&current_azks).await?;
+        let hash = akd
+            .get_root_hash::<Blake3_256<BaseElement>>(&current_azks)
```
slawlor, line 688: nit: Here too, `Blake3` is just cleaner as a type alias

akd/src/directory.rs
```diff
@@ -740,136 +741,137 @@ mod tests {
     }
 
     // BUG: Key history verification is failing on small-sized trees. It's being addressed in Issue #144
```
slawlor, line 743: If solved, we should remove this

akd/src/node_state.rs
```diff
@@ -747,11 +747,11 @@ mod tests {
 
     #[test]
     pub fn serialize_deserialize() {
-        use winter_crypto::hashers::Blake3_256;
+        use winter_crypto::hashers::Sha3_256;
```
slawlor, line 750: Why are we changing the hashing scheme here?

akd/src/node_state.rs
```diff
@@ -747,11 +747,11 @@ mod tests {
...
         use winter_crypto::Hasher;
         use winter_math::fields::f128::BaseElement;
 
-        type Blake3 = Blake3_256<BaseElement>;
+        type Blake3 = Sha3_256<BaseElement>;
```
slawlor, line 754: This isn't correct, if we are going to use SHA3 256, we should make the alias something else or revert back to blake3

slawlor (changes requested):
In `directory.rs` lines 1104-1111 should also be uncommented and tested as they were failing due to this issue

Additionally a MySQL test is failing, so we'll need to see what is going on. It's another LCP hash <> hash error, so it seems related

akd/src/client.rs
```diff
@@ -61,15 +61,28 @@ pub fn verify_nonmembership<H: Hasher>(
     proof: &NonMembershipProof<H>,
 ) -> Result<bool, AkdError> {
     let mut verified = true;
-    let mut lcp_hash = H::hash(&[]);
+    let mut lcp_hash = H::hash(&EMPTY_VALUE);
     let mut lcp_real = proof.longest_prefix_children[0].label;
+    // if lcp_real == EMPTY_LABEL {
```
slawlor, line 66: Also are the necessary or just debugging statements no longer needed?

akd/src/directory.rs
Jasleen1: This particular test has a hard-coded hash to check against. I had in mind that a future PR could replace the concise `Blake3` with `TestHash` (for e.g.), which could depend on a feature or be a simple change of variable somewhere to try different hash functions. This test will fail, if that is not taken care of by actually fixing the hash fn to `Blake3_256<BaseElement>`.

akd/src/directory.rs
Jasleen1: Same response as above for this test.

Issue #156, opened 2022-02-28
MySQL error `HistoryTreeNode`s `batch_get`


eozturk1:
For big-sized trees, `mysql_tests` are failing in `batch_get` due to errors similar to

`ERROR HY000 (1366): Incorrect integer value: '??%????\u{3}?\\??N??\u{4}a??H??~~S??!DU9D' for column 'label_val' at row 1`

slawlor:
The `get_batch_create_temp_table` function has some mismatching type information from the rest of the MySQL implementation.

It should read something like
```rust
fn get_batch_create_temp_table<St: Storable>() -> Option<String> {
        match St::data_type() {
            StorageType::Azks => None,
            StorageType::HistoryNodeState => {
                Some(
                    format!(
                        "CREATE TEMPORARY TABLE `{}`(`label_len` INT UNSIGNED NOT NULL, `label_val` VARBINARY(32) NOT NULL, `epoch` BIGINT UNSIGNED NOT NULL, PRIMARY KEY(`label_len`, `label_val`, `epoch`))",
                        TEMP_IDS_TABLE
                    )
                )
            },
            StorageType::HistoryTreeNode => {
                Some(
                    format!(
                        "CREATE TEMPORARY TABLE `{}`(`label_len` INT UNSIGNED NOT NULL, `label_val` VARBINARY(32) NOT NULL, PRIMARY KEY(`label_len`, `label_val`))",
                        TEMP_IDS_TABLE
                    )
                )
            },
            StorageType::ValueState => {
                Some(
                    format!(
                        "CREATE TEMPORARY TABLE `{}`(`username` VARCHAR(256) NOT NULL, `epoch` BIGINT UNSIGNED NOT NULL, PRIMARY KEY(`username`, `epoch`))",
                        TEMP_IDS_TABLE
                    )
                )
            },
        }
    }
```

slawlor:
I think we're resolved on this no?

PR #158, opened 2022-03-01
Fix MySQL tests

eozturk1:
MySQL tests were failing initially due to type mismatch in database tables. This PR includes the fix @slawlor suggested in Issue #156. 

After that fix, I started seeing hash lookup proof verification failures and tracked it down to [this](https://github.com/novifinancial/akd/commit/0604caae0adcd561ebe39406b5779f71b081fe83) commit and eventually to incorrect user state retrieval on the MySQL end when `LeqEpoch` is used. The issue is fixed by ordering the user state descending since the max value is needed.


afterdusk (approved):
Looks good, thanks for digging into this @eozturk1 !

slawlor (approved):
Love love love, thank you! Awesome work

Issue #103, opened 2021-12-08
Bulk lookup proofs


slawlor:
In production, under high-load we might need the ability to perform "bulk" lookup proofs. This means that we should do the same breadth-first-search for a collection of users as is done under the ```Directory::publish``` operation (as is done in ```append_only_zks.rs:191```).

This means, when doing a lookup we should calculate all the prefixes and pre-load (via BFS walk of batch-retrievals) all the nodes along the path into the cache. Then we can generate the proofs without doing many individual retrievals.

eozturk1:
Looking at the `preload_nodes_for_insertion` function, it retrieves the nodes from storage but I am not sure how it achieves the caching we would like to have. <I might be missing something here but> the function does not seem to keep the nodes, e.g., in memory. How do we want to cache the nodes for bulk lookup proofs?

slawlor:
Hey @eozturk1, so in the MySQL storage layer we already have a caching scheme implemented. By using the ```preload_nodes_for_insertion``` you're effectively using ```batch get``` operations rather than individual ```get``` ops. Anytime a record is retrieved, we're appending it to the cache so you don't need to worry about manually inserting them into any cache, it's auto-magic ;)

The bulk-get is MUCH fewer db ops which is why it's helpful to retrieve them all before, then after that it's just memory operations which are super cheap. Even though we're double-getting data, it's still way more performant and makes the code transparent to the AKD dev.

eozturk1:
Ah, that's perfect! Just saw the `TimedCache`.:) I'll use the same batch ops for bulk proofs.

eozturk1:
I'll close this one since the bulk of the task has been completed in #163. The remaining optimizations are tracked in #166, #167 and #169.

PR #163, opened 2022-03-07
Merge bulk lookup proofs

eozturk1:
This PR adds support for bulk lookup proofs (See Issue #103). 

First, for each lookup proof requested, necessary labels are calculated. Then, the nodes corresponding to their self- and sibling-prefixes are preloaded before the lookup operations (See the comments above the function `build_lookup_prefixes_set` for my understanding why this is the case). 

To test it, I added a new feature to the MySQL implementation where each storage call is tracked with its self-(`read`/`write`) and data-type (e.g., `HistoryTreeNode`). Here is its output with the following parameters:

```diff
num_users: 50
num_epochs: 5
num_lookups: 100

running 1 test
Metrics after publish(es).
MySQL writes: 3279, MySQL reads: 6, Time read: 0.071272952 s, Time write: 1.13520928 s
        Tree size: 899
        Node state count: 1386
        Value state count: 250
Read call stats: [("get_direct:~Azks", 1), ("get_user_state_versions~", 5)]
Write call stats: [("internal_batch_set~", 20), ("internal_set~", 3)]

- Individual 100 lookups took 3103ms.
Metrics after individual lookups:
MySQL writes: 0, MySQL reads: 969, Time read: 2.041950956 s, Time write: 0 s
        Tree size: 899
        Node state count: 1386
        Value state count: 250
- Read call stats: [("get_direct:~Azks", 1), ("get_direct:~HistoryNodeState", 459), ("get_direct:~HistoryTreeNode", 459), ("get_user_state~", 50)]
Write call stats: []

+ Bulk 100 lookups took 1732ms.
Metrics after lookup proofs: 
MySQL writes: 0, MySQL reads: 75, Time read: 0.59272097 s, Time write: 0 s
        Tree size: 899
        Node state count: 1386
        Value state count: 250
+ Read call stats: [("batch_get~HistoryNodeState", 12), ("batch_get~HistoryTreeNode", 12), ("get_direct:~Azks", 1), ("get_user_state~", 50)]
Write call stats: []
```

Note the reduction in completion time (about 40% but may vary depending on the parameters) and `get_direct` calls replaced with `batch_get`s in bulk lookups.

This code was a bit tricky to test since MySQL metrics are reset when printed and nodes in cache may change the resulting numbers. In addition, the users looked up might result in different time measurements. I tried and accounted for all that but if you notice any issues, please let me know!

### DIFF
--- a/akd/src/directory.rs
+++ b/akd/src/directory.rs
@@ -230,8 +230,6 @@ impl<S: Storage + Sync + Send, V: VRFKeyStorage> Directory<S, V> {
             .get_root_hash_at_epoch::<_, H>(&self.storage, next_epoch)
             .await?;
 
-        // self.storage.log_metrics(log::Level::Info).await;
-
         Ok(EpochHash(current_epoch, root_hash))
         // At the moment the tree root is not being written anywhere. Eventually we
         // want to change this to call a write operation to post to a blockchain or some such thing
@@ -299,8 +297,6 @@ impl<S: Storage + Sync + Send, V: VRFKeyStorage> Directory<S, V> {
                 .await?,
         };
 
-        // self.storage.log_metrics(log::Level::Info).await;
-
         Ok(lookup_proof)
     }
 
@@ -352,8 +348,6 @@ impl<S: Storage + Sync + Send, V: VRFKeyStorage> Directory<S, V> {
                 .await?,
             );
         }
-
-        // self.storage.log_metrics(log::Level::Info).await;
 
         Ok(lookup_proofs)
     }
--- a/akd/src/directory.rs
+++ b/akd/src/directory.rs
@@ -22,6 +22,7 @@ use log::{debug, error, info};
 #[cfg(feature = "rand")]
 use rand::{CryptoRng, RngCore};
 
+use crate::node_state::NodeLabel;
 use std::collections::HashMap;
 use std::marker::{Send, Sync};
 use std::sync::Arc;
@@ -30,6 +31,16 @@ use winter_crypto::Hasher;
 /// Root hash of the tree and its associated epoch
 #[derive(Debug, Clone, Hash, Eq, PartialEq)]
 pub struct EpochHash<H: Hasher>(pub u64, pub H::Digest);
+
+#[derive(Clone)]
+/// Info needed for a lookup of a user for an epoch
+pub struct LookupInfo {
+    value_state: ValueState,
+    marker_version: u64,
+    existent_label: NodeLabel,
+    marker_label: NodeLabel,
+    non_existent_label: NodeLabel,
+}
 
 #[cfg(feature = "rand")]
 impl AkdValue {
@@ -219,7 +230,7 @@ impl<S: Storage + Sync + Send, V: VRFKeyStorage> Directory<S, V> {
             .get_root_hash_at_epoch::<_, H>(&self.storage, next_epoch)
             .await?;
 
-        self.storage.log_metrics(log::Level::Info).await;
+        // self.storage.log_metrics(log::Level::Info).await;
 
         Ok(EpochHash(current_epoch, root_hash))
         // At the moment the tree root is not being written anywhere. Eventually we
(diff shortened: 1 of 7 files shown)


akd/src/directory.rs
```diff
@@ -219,7 +230,7 @@ impl<S: Storage + Sync + Send, V: VRFKeyStorage> Directory<S, V> {
             .get_root_hash_at_epoch::<_, H>(&self.storage, next_epoch)
             .await?;
 
-        self.storage.log_metrics(log::Level::Info).await;
+        // self.storage.log_metrics(log::Level::Info).await;
```
slawlor, line 233: Does it make sense to just remove this commented statement completely?

akd/src/directory.rs
```diff
@@ -233,68 +244,155 @@ impl<S: Storage + Sync + Send, V: VRFKeyStorage> Directory<S, V> {
...
+                )
+                .await?,
+        };
+
+        // self.storage.log_metrics(log::Level::Info).await;
+
+        Ok(lookup_proof)
+    }
+
+    // TODO(eoz): Call proof generations async
```
slawlor, line 307: Can we file an issue to follow and discuss this? What needs to be called async?

akd/src/directory.rs
```diff
@@ -233,68 +244,155 @@ impl<S: Storage + Sync + Send, V: VRFKeyStorage> Directory<S, V> {
...
+    ) -> Result<Vec<LookupProof<H>>, AkdError> {
+        let current_azks = self.retrieve_current_azks().await?;
+        let current_epoch = current_azks.get_latest_epoch();
+
+        // Take a union of the labels we will need proofs of for each lookup.
+        let mut lookup_labels = Vec::new();
+        let mut lookup_infos = Vec::new();
+        for uname in unames {
+            // Save lookup info for later use.
+            let lookup_info = self
```
slawlor, line 321: OK so we're preloading nodes, but we're still iteratively loading this. Can we batch this as well? We should know all the necessary information to retrieve the batch of user value state nodes.

Maybe a follow-up issue perhaps for further optimizations?

akd/src/directory.rs
```diff
@@ -233,68 +244,155 @@ impl<S: Storage + Sync + Send, V: VRFKeyStorage> Directory<S, V> {
...
+            .bfs_preload_nodes::<_, H>(&self.storage, lookup_prefixes_set)
+            .await?;
+
+        // Ensure we have got all lookup infos needed.
+        assert_eq!(unames.len(), lookup_infos.len());
+
+        let mut lookup_proofs = Vec::new();
+        for i in 0..unames.len() {
+            lookup_proofs.push(
+                self.lookup_with_info::<H>(
```
slawlor, line 346: Why do we need the info? I think if we've done the proper BFS loading, we should only need to batch-get the value states in order to load them into the cache, then we can do the regular proof generation operations since we'll only be accessing from local memory

akd/src/node_state.rs
```diff
@@ -148,6 +148,31 @@ impl NodeLabel {
         Self::new(out_val, len)
     }
 
+    pub(crate) fn get_sibling_prefix(&self, mut len: u32) -> Self {
```
slawlor, line 151: nit: Comments explaining what is going on here would be probably good for future support.

akd_mysql/src/mysql.rs
```diff
@@ -169,7 +173,9 @@ impl<'a> AsyncMySqlDatabase {
             trans: LocalTransaction::new(),
 
             num_reads: Arc::new(tokio::sync::RwLock::new(0)),
+            read_call_stats: Arc::new(tokio::sync::RwLock::new(HashMap::new())),
```
slawlor, line 176: Can we make these optional? Or enabled with a compilation feature flag? This is unbounded memory should anyone not flush the call stats. Ints and durations I'm not so worried about, but this is a list that will grow indefinitely. While useful for debugging, it's a mem leak in prod

slawlor (changes requested):
It's close but we need to fix the mysql metrics gathering and I think there's more optimizations that are possible for batch lookups. But the second part could be pushed to a future PR.

akd/src/directory.rs
eozturk1: I think it's a great debugging tool but it requires care to use it correctly. Better to manually add/remove in tests.

akd/src/directory.rs
eozturk1: #166. So I think we can await on the completion of all lookups rather than awaiting for individual lookups and proceeding in the loop. Assuming my understanding of Rust's async is correct.

akd/src/directory.rs
eozturk1: Nice catch, we can do that! Tracked in #167.

akd/src/directory.rs
eozturk1: This is to eliminate duplicate `get_user_state` queries. One in identifying which labels are needed and one in (previously) lookup for the same purpose. This way pre-loaded info is passed to the lookup function, so no re-loads.

Looking at the code for MySQL's `get_user_state`, are we utilizing the cache?

akd_mysql/src/mysql.rs
eozturk1: As is, this will not use unbounded memory. Max number of elements is ` < f * t` where `f` is the number of functions we have in the database and `t` is the type of data loaded through these functions.

Below `get_direct` loads three types of data: `Azks`, `HistoryNodeState` and `HistoryTreeNode`. `get_user_state` doesn't have an associated data type, so it can only have one element at max.

`[("get_direct:~Azks", 1), ("get_direct:~HistoryNodeState", 459), ("get_direct:~HistoryTreeNode", 459), ("get_user_state~", 50)]`

Also, if a function is not called, it is not added to the stats.

I think additional burden of this computation and memory-wise is minimal but let me know, I can look into feature flags.

akd/src/directory.rs
slawlor: Ah so the only thing that's "async" in the underlying logic here is the access to data-storage. Which assuming you've pre-cached all of the nodes will simply be accesses to memory storage but it can be read in parallel. I'm not sure how much we could gain by doing them all in parallel, but it should be trivial to do. You simply don't `.await` the futures and collect them in a `Vec`. Then you can await the batch of them with the `futures` crate

i.e. something like

```rust
use futures::future;

// generate your vector of individual tasks by spawning them on green threads
// but you need to clone "self" in order to apply on different threads, which will
// shard the data-access pointer and cache pointers, etc to yield individual Directory<_>
// instances on each thread
let tasks = proofs_to_generate.into_iter().map(|proof| {
  let self_clone = self.clone();
  tokio::spawn(async move { self_clone.lookup(proof).await })
}

let outputs = future::try_join_all(tasks).await?;
```

akd/src/directory.rs
slawlor: Ah I see, yes we've excluded `get_user_state` from the caching logic because of the filter parameters. How would you know a cached value matches "<= epoch" or some weird filter without going to the backing store? It sounds like we might need a new query at the storage layer to retrieve the max versions for a collection of users which is <= a given epoch. (If I'm remembering right). We could move forward with this however and simply open an issue for it, but we'll want to do it properly for batch proof generation. Here's we will hit order N queries just to gather the user states, while not awful, still not ideal when we can do a better query.

akd_mysql/src/mysql.rs
slawlor: OK yeah I see now, you're keying on the string so there is a finite upper limit. OK I'm fine with that then, I was more worried on dynamic key generation for example, with no pruning. But in this case we're safe.

akd/src/directory.rs
eozturk1: Got it! I can try and add it in the next PR along with #167.

akd/src/directory.rs
eozturk1: Ah I see! Tracked in #169. I can take a stab at #166, #167 and #169 in a new PR. Does that sound good?

eozturk1:
I'll work on #166, #167 and #169 in a different PR. Is there anything I should update in this PR to proceed with merging?

akd/src/directory.rs
slawlor: Sounds great, let's merge this in!

slawlor (approved):

Issue #157, opened 2022-02-28
Improve debugging by adding error contexts


eozturk1:
We have errors that **only** include the error string in the error. When these errors surface up, they don't have any context on where and why they occurred (e.g., a MySQL error). We should improve these by adding

- More context to the errors, e.g., `GetData` error in `batch_get` should mention that the operation was batch get.
- Stack trace info.  --> Since the stack trace leading to the error is cleared up when the error is caught, we lose this info.
- Line numbers and file names when the error occurred.

slawlor:
A good point to check into on this might be the `log` crate, there is already a capturing which occurs with each macro call (e.g. `debug!(...)`) that is capturing a lot of this information. You can check our own log implementations, but we're capturing exactly this kind of info (file, line, thread, etc).

eozturk1:
The logs by the `log` crate are not displayed (except `println!`). We need a logger implementation other than the `log` crate -- maybe [env_logger](https://crates.io/crates/env_logger). And [initialize](https://stackoverflow.com/questions/30177845/how-to-initialize-the-logger-for-integration-tests) it. Then we can get those logs.

slawlor:
Ah I meant looking at how they capture the file, line number, method, etc. Not actually relying on it, sorry!

slawlor:
So I think logging is going to be the way to go here. I was looking into how this would be implemented, but we would need to redefine the error structures we pass around to also collect a "context" object and when an error is to be returned, append the context about module,line,file,etc. This means modifying every location where an error response is generated, and kind of goes against the "spirit" of Rust's `Result<_,_>` type (from what I'm reading).

We can do it, but I think utilizing the log crate more heavily is going to make more sense here. I did do a large cleanup of the error types and how they're propagated around to make more sense of them, but no context as of yet.

eozturk1:
Yeah, +1 on using the log crate. With proper context and unique error messages, we don't need the file/line to be included. I guess the initial issue we should focus on is actually getting the logs using env_logger. Otherwise, if they aren't `println!` I can't see any logs.

slawlor:
> Yeah, +1 on using the log crate. With proper context and unique error messages, we don't need the file/line to be included. I guess the initial issue we should focus on is actually getting the logs using env_logger. Otherwise, if they aren't `println!` I can't see any logs.

[...]

PR #168, opened 2022-03-09
Error handling cleanup

slawlor:
A decent-sized rewrite of the error constructs to clarify a lot of the error patterns (i.e. storage throws clearer errors rather than just get/set). Additionally

1. Small utility functions which cannot throw errors should not return `Result<_, _>` and have been rewritten to not be error potential.
2. Digest serialization changed to `[u8; 32]` from `Vec<u8>` as we know the size and it makes it clearer to other developers downstream

### DIFF
--- a/akd/Cargo.toml
+++ b/akd/Cargo.toml
@@ -45,6 +45,7 @@ proptest = "1.0"
 proptest-derive = "0.3"
 colored = { version = "2" }
 once_cell = { version = "1" }
+ctor = "0.1"
 
 akd = { path =".", features = ["vrf", "public-tests"] }
 
--- a/akd/Cargo.toml
+++ b/akd/Cargo.toml
@@ -11,7 +11,9 @@ readme = "../README.md"
 
 [features]
 bench = ["public-tests"]
-public-tests = ["rand", "bincode", "serde"]
+public-tests = ["rand", "bincode", "serde", "colored", "once_cell", "ed25519-dalek/serde"]
+# In the event that VRF's are enabled, AND builder has requested serde support
+# Add the serde flag to the dalek crate with --features "ed25519-dalek/serde"
 vrf = ["curve25519-dalek", "ed25519-dalek"]
 default = ["vrf"]
 
@@ -32,18 +34,19 @@ bincode = { version = "1", optional = true }
 serde = { version = "1", features = ["derive"], optional = true }
 rand = { version = "0.7", optional = true }
 curve25519-dalek = { version = "3.2", optional = true }
-ed25519-dalek = { version = "1", optional = true}
-
-[target.'cfg(all(features = "serde", features = "vrf"))'.dependencies]
-# In the event that VRF's are enabled, AND builder has requested serde support
-# Add the serde flag to the dalek crate
-ed25519-dalek = { version = "1", features = ["serde"]}
+ed25519-dalek = { version = "1", optional = true }
+colored = { version = "2", optional = true }
+once_cell = { version = "1", optional = true }
 
 [dev-dependencies]
 criterion = "0.3"
 serial_test = "0.5"
 proptest = "1.0"
 proptest-derive = "0.3"
+colored = { version = "2" }
+once_cell = { version = "1" }
+
+akd = { path =".", features = ["vrf", "public-tests"] }
 
 [[bench]]
 name = "azks"
--- a/akd/src/directory.rs
+++ b/akd/src/directory.rs
@@ -736,8 +736,6 @@ mod tests {
 
     #[tokio::test]
     async fn test_empty_tree_root_hash() -> Result<(), AkdError> {
-        crate::test_utils::init_logger(log::Level::Info);
-
         let db = AsyncInMemoryDatabase::new();
         let vrf = HardCodedAkdVRF {};
         let akd = Directory::<_, _>::new::<Blake3_256<BaseElement>>(&db, &vrf, false).await?;
@@ -757,8 +755,6 @@ mod tests {
 
     #[tokio::test]
     async fn test_simple_publish() -> Result<(), AkdError> {
-        crate::test_utils::init_logger(log::Level::Info);
-
         let db = AsyncInMemoryDatabase::new();
         let vrf = HardCodedAkdVRF {};
         let akd = Directory::<_, _>::new::<Blake3>(&db, &vrf, false).await?;
@@ -773,8 +769,6 @@ mod tests {
 
     #[tokio::test]
     async fn test_simple_lookup() -> Result<(), AkdError> {
-        crate::test_utils::init_logger(log::Level::Info);
-
         let db = AsyncInMemoryDatabase::new();
         let vrf = HardCodedAkdVRF {};
         let akd = Directory::<_, _>::new::<Blake3>(&db, &vrf, false).await?;
@@ -806,8 +800,6 @@ mod tests {
 
     #[tokio::test]
     async fn test_simple_key_history() -> Result<(), AkdError> {
-        crate::test_utils::init_logger(log::Level::Info);
-
         let db = AsyncInMemoryDatabase::new();
         let vrf = HardCodedAkdVRF {};
         let akd = Directory::<_, _>::new::<Blake3>(&db, &vrf, false).await?;
@@ -941,8 +933,6 @@ mod tests {
     #[allow(unused)]
     #[tokio::test]
     async fn test_simple_audit() -> Result<(), AkdError> {
-        crate::test_utils::init_logger(log::Level::Info);
-
         let db = AsyncInMemoryDatabase::new();
         let vrf = HardCodedAkdVRF {};
         let mut akd = Directory::<_, _>::new::<Blake3>(&db, &vrf, false).await?;
@@ -1101,8 +1091,6 @@ mod tests {
 
     #[tokio::test]
     async fn test_read_during_publish() -> Result<(), AkdError> {
-        crate::test_utils::init_logger(log::Level::Info);
-
         let db = AsyncInMemoryDatabase::new();
         let vrf = HardCodedAkdVRF {};
         let akd = Directory::<_, _>::new::<Blake3>(&db, &vrf, false).await?;
@@ -1210,8 +1198,6 @@ mod tests {
 
     #[tokio::test]
     async fn test_directory_read_only_mode() -> Result<(), AkdError> {
-        crate::test_utils::init_logger(log::Level::Info);
-
         let db = AsyncInMemoryDatabase::new();
         let vrf = HardCodedAkdVRF {};
         // There is no AZKS object in the storage layer, directory construction should fail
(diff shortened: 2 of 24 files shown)


akd/Cargo.toml
```diff
@@ -32,18 +34,19 @@ bincode = { version = "1", optional = true }
 serde = { version = "1", features = ["derive"], optional = true }
 rand = { version = "0.7", optional = true }
 curve25519-dalek = { version = "3.2", optional = true }
-ed25519-dalek = { version = "1", optional = true}
-
-[target.'cfg(all(features = "serde", features = "vrf"))'.dependencies]
```
slawlor, line 37: This is apparently not supported (and had typos anyways). Need to have builder utilize features on the crate to support it.

akd/Cargo.toml
```diff
@@ -32,18 +34,19 @@ bincode = { version = "1", optional = true }
...
 
 [dev-dependencies]
 criterion = "0.3"
 serial_test = "0.5"
 proptest = "1.0"
 proptest-derive = "0.3"
+colored = { version = "2" }
+once_cell = { version = "1" }
+
+akd = { path =".", features = ["vrf", "public-tests"] }
```
slawlor, line 49: We add ourself back in in the `test` profile in order to adjust the features on the create (i.e. enable the test-logic needed)

akd/src/errors.rs
```diff
@@ -158,57 +169,47 @@ impl fmt::Display for AzksError {
...
     VerifyLookupProof(String),
     /// Key-History proof did not verify
     VerifyKeyHistoryProof(String),
     /// Tried to audit an invalid epoch range
     InvalidEpoch(String),
-    /// Error propagation
-    Storage(StorageError),
-    /// Error propagation for errors from VRF storage
-    VRFStorageErr(VRFStorageError),
+    /// AZKS not found in read-only directory mode
```
afterdusk, line 188: Technically this error is thrown when we try to perform a write operation in read-only mode, right? This includes trying to publish. I think the error message can be clearer, especially in terms of what the bool stands for.

akd/src/errors.rs
```diff
@@ -217,36 +218,61 @@ impl fmt::Display for DirectoryError {
...
+                write!(f, "Other storage error: {}", inner)
+            }
+        }
+    }
 }
 
 /// Represents a VRF-storage-layer error
 #[derive(PartialEq, Debug)]
-pub enum VRFStorageError {
+pub enum VrfError {
```
afterdusk, line 254: I think the comments for the variants should be updated too!

akd/src/errors.rs
```diff
@@ -217,36 +218,61 @@ impl fmt::Display for DirectoryError {
...
             }
-            Self::GetPK(error_string) => {
-                write!(f, "{}", error_string)
+            Self::PublicKey(error_string) => {
+                write!(f, "VRF public key: {}", error_string)
             }
-            Self::VRFErr(error_string) => {
-                write!(f, "{}", error_string)
+            Self::Verification(error_string) => {
+                write!(f, "VRF prooving or verifying: {}", error_string)
```
afterdusk, line 275: nit: sp!

akd/src/directory.rs
```diff
@@ -218,9 +212,7 @@ impl<S: Storage + Sync + Send, V: VRFKeyStorage> Directory<S, V> {
             if let Err(err) = self.storage.commit_transaction().await {
                 // ignore any rollback error(s)
                 let _ = self.storage.rollback_transaction().await;
-                return Err(AkdError::HistoryTreeNode(HistoryTreeNodeError::Storage(
-                    err,
-                )));
+                return Err(AkdError::Storage(err));
```
eozturk1, line 215: nit: Maybe add the origin to the message? "Couldn't commit transaction in publish"

akd/src/directory.rs
```diff
@@ -419,10 +408,10 @@ impl<S: Storage + Sync + Send, V: VRFKeyStorage> Directory<S, V> {
             }
             Ok(HistoryProof { proofs })
         } else {
-            Err(AkdError::Directory(DirectoryError::NonExistentUser(
-                username,
-                current_epoch,
-            )))
+            Err(AkdError::Storage(StorageError::NotFound(format!(
+                "User {} at epoch {}",
```
eozturk1, line 412: Nice!

akd/src/errors.rs
```diff
@@ -11,22 +11,22 @@ use core::fmt;
 use crate::node_state::NodeLabel;
 
 /// Symbolizes a AkdError, thrown by the akd.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
```
eozturk1, line 14: Why do we use `PartialEq` here?

akd_client/src/tests.rs
```diff
@@ -255,7 +255,7 @@ async fn test_simple_lookup() -> Result<(), AkdError> {
         target_label_bytes,
         internal_lookup_proof,
     )
-    .map_err(|i_err| AkdError::AzksNotFound(format!("Internal: {:?}", i_err)));
+    .map_err(|i_err| AkdError::Storage(StorageError::Other(format!("Internal: {:?}", i_err))));
```
afterdusk, line 258: What's the rationale for the Storage variant here?

akd/src/errors.rs
```diff
@@ -45,9 +45,9 @@ impl From<DirectoryError> for AkdError {
     }
 }
 
-impl From<VRFStorageError> for AkdError {
-    fn from(error: VRFStorageError) -> Self {
-        Self::VRFStorageErr(error)
+impl From<VrfError> for AkdError {
```
eozturk1, line 48: Do we lose any information on the error here? i.e., previously it was `VRF"Storage"Error` now it is only `VrfError`

akd/src/errors.rs
```diff
@@ -57,15 +57,25 @@ impl From<AzksError> for AkdError {
...
+            AkdError::Directory(err) => {
+                writeln!(f, "AKD Directory Error: {}", err)
+            }
+            AkdError::AzksErr(err) => {
+                writeln!(f, "AKD AZKS Error: {}", err)
+            }
+            AkdError::Vrf(err) => {
+                writeln!(f, "AKD VRF Error: {}", err)
+            }
+            AkdError::Storage(err) => {
```
eozturk1, line 75: Love this!

akd_mysql/src/mysql.rs
```diff
@@ -794,7 +795,10 @@ impl Storage for AsyncMySqlDatabase {
 
         match self.internal_set(record, None).await {
             Ok(_) => Ok(()),
-            Err(error) => Err(StorageError::SetData(error.to_string())),
+            Err(error) => {
+                error!("MySQL error {}", error);
+                Err(StorageError::Other(format!("MySQL Error {}", error)))
```
afterdusk, line 800: nit: Maybe a `Database` or `Operation` variant? `Other` is fine too though!

akd/src/storage/memory.rs
```diff
@@ -155,15 +155,19 @@ impl Storage for AsyncInMemoryDatabase {
                         return Ok(DbRecord::ValueState(item.clone()));
                     }
                 }
-                return Err(StorageError::GetData("Not found".to_string()));
+                return Err(StorageError::NotFound(format!("ValueState {:?}", id)));
```
eozturk1, line 158: nit: Maybe add "while getting `ValueState` directly from storage"?

akd/src/storage/memory.rs
```diff
@@ -155,15 +155,19 @@ impl Storage for AsyncInMemoryDatabase {
...
             }
         }
         // fallback to regular get/set db
         let guard = self.db.read().await;
         if let Some(result) = (*guard).get(&bin_id).cloned() {
             Ok(result)
         } else {
-            Err(StorageError::GetData("Not found".to_string()))
+            Err(StorageError::NotFound(format!(
```
eozturk1, line 166: The origin of this error would be really helpful here when debugging. Can we clarify this is erroring out in `get_direct`?

akd/src/storage/memory.rs
```diff
@@ -197,7 +201,7 @@ impl Storage for AsyncInMemoryDatabase {
 
             Ok(KeyData { states: results })
         } else {
-            Err(StorageError::GetData("Not found".to_string()))
+            Err(StorageError::NotFound(format!("ValueState {:?}", username)))
```
eozturk1, line 204: Same here. We should add the origins of the calls, i.e., getting user data.

akd/src/test_utils.rs
```diff
@@ -0,0 +1,112 @@
...
+/// This is safe to call multiple times, but it will only initialize the logger
+/// to the log-level _first_ set. If you want a specific log-level (e.g. Debug)
+/// for a specific test, make sure to only run that single test after editing that
+/// test's log-level.
+///
+/// The default level applied everywhere is Info
+pub fn init_logger(level: Level) {
+    EPOCH.get_or_init(Instant::now);
+
+    INIT_ONCE.call_once(|| {
```
eozturk1, line 94: Beautiful!

afterdusk (approved):
Great stuff as usual, thanks @slawlor ! Just some non-blocking nits and questions 🙂

akd/src/tests.rs
```diff
@@ -713,3 +695,14 @@ async fn test_insert_single_leaf_full_tree() -> Result<(), HistoryTreeNodeError>
     assert!(root_val == expected, "Root hash not equal to expected");
     Ok(())
 }
+
+#[test]
+// Comment the following line to see log messages being written in failed testing scenarios
+#[should_panic(expected = "boom.")]
```
eozturk1, line 701: Haha!

akd_mysql/src/mysql.rs
```diff
@@ -794,7 +795,10 @@ impl Storage for AsyncMySqlDatabase {
...
+            }
```
eozturk1, line 801: Mention internal set?

eozturk1 (approved):
It looks great @slawlor! For some errors though I think we need more context (added some comments but there may be more). If you prefer to defer adding these contexts, I'd suggest keeping #157 open. I can work on it.

akd/src/storage/memory.rs
slawlor: We could add log statements, but I specifically excluded "not found" results from log statements as it's a pretty common, expected flow. All of the storage errors which have real failures (i.e not "not found") include a log statement for the error.

akd/src/storage/memory.rs
slawlor: See further down

akd_mysql/src/mysql.rs
slawlor: Yeah `Other` is generally going to be some operation we could retry. If we needed (eventually) an error code which we _should not_ retry, then I'd say at that point to add something. But I think this is OK for now

PR #176, opened 2022-03-16
Making the dalek dependencies a less specific version

slawlor:
Rolling the `curve25519-dalek` crate to only be v 3 instead of v 3.2 which allows for more downstream flexibility in version upgrades.

Also added a new meta-feature for proper serde handling. Will release v 0.5.3 after merging


eozturk1 (approved):
LGTM!

PR #177, opened 2022-03-16
Missed a dalek version

slawlor:
In the `akd` crate, missed downgrading dalek crate from 3.2 -> 3 (bad merge)

### DIFF
--- a/akd/Cargo.toml
+++ b/akd/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "akd"
-version = "0.5.3"
+version = "0.5.4"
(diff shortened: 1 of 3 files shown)


eozturk1 (approved):
I might've said this before but I think we should have a central cargo file where we load these shared repos from.

This is actually being worked on at the moment: https://github.com/rust-lang/cargo/issues/8415.

In the meantime, is it worth investing in writing a script for making these changes automatically? As in "update curve25519-dalek to version 3" and the script does all necessary changes for all crates.

slawlor:
> In the meantime, is it worth investing in writing a script for making these changes automatically? As in "update curve25519-dalek to version 3" and the script does all necessary changes for all crates.

We could but that seems like a significant amount of work honestly. We might look into `cargo tree` to make sure common versions are utilized in our repos, but I wouldn't go much beyond that atm.

Thanks for the link though! Good to hear that they're working on something sensible for this.

PR #179, opened 2022-03-16
Moving around location of tests

kevinlewi:
This moves the existing tests in `tests.rs` over into `history_tree_node.rs`, and then moves the tests in `directory.rs` over into `tests.rs`. This is because the main tests for the high-level API were in directory.rs (which also tested the client and auditor functions), and I thought it would be cleaner to have them in the tests.rs file.


slawlor (approved):
LGTM

PR #180, opened 2022-03-17
Using EMPTY_VALUE in place of &[0u8]

kevinlewi:
Found a couple of places where we should be referring to the constant instead of &[0u8]


slawlor (approved):
LGTM! We're experiencing problems with github PRs/pushes/pulls/etc... pretty much everything at the moment. You might want to wait merging until the statuses are green here:
https://www.githubstatus.com/

PR #181, opened 2022-03-17
Simplify node label hasher

kevinlewi:
Reducing this function from performing 3 hash operations to just 1. We can concatenate the two inputs together and just do one outer hash, rather than first hashing the two items together and then doing the outer hash.

(Based on top of https://github.com/novifinancial/akd/pull/180)

### DIFF
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -748,7 +748,7 @@ mod tests {
         let mut proof = azks
             .get_membership_proof(&db, insertion_set[0].label, 1)
             .await?;
-        let hash_val = Blake3::hash(&[0u8]);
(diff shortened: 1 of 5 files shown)


slawlor (approved):
Github (and generally PRs) don't support stacked PRs well (not really at all). Once you merge #180, you'll want to merge this again on main, then we can land this (presently this PR looks like both since it's just this <-> main instead of this <-> #180 <-> main).

Also I noted on #180, but we're experiencing a lot of issues with github right now. You might have fun problems (i.e. no CI actions or pulls/pushes failing with internal server error). I'd wait until https://www.githubstatus.com/ says all is well before trying much!

LGTM!

Issue #184, opened 2022-03-17
Derive all server secrets from a common string instead of the VRF private key


kevinlewi:
In #123, we need to introduce a "commitment key", which is similar to the VRF private key in that the server needs to store it somewhere and keep it secure.

However, instead of introducing another kind of storage for this, I think it would be better to slightly generalize how we are doing the VRF private key storage at the moment. Right now, we store the VRF private key somewhere, but ideally we would actually store a "ServerSecret" 32-byte string, from which the VRF private key could be derived (say, by hashing this server secret string along with a nonce), and from which a "commitment key" that I want to introduce will also be derived from that same server secret string. Any future server-only secrets could also be derived from this string, without having to change the storage layer.

At the moment, the commitment key is derived directly from the VRF private key for now -- this should be fixed.

kevinlewi:
Honestly, hashing the VRF private key to get the commitment key is OK, and in some ways simpler.

PR #185, opened 2022-03-18
Using proper commitments for the AkdValue leaf node representation

### LINKED ISSUE #123: Creating commits for values
At the moment the values being committed in the tree for prototyping are all dummies. We would like to replace the `value_to_bytes` function in `directory.rs` to return actual commitments instead. This involves 3 main steps:

- [ ] Changing the function `value_to_bytes` to return a byte array derived from a digest instead.
- [ ] To make the commitment hiding, generating and storing a nonce for each committed value.
- [ ] Including the nonce in the proofs sent to the client, and having the client verify these nonces.

kevinlewi:
Introducing a commitment to replace the `value_to_bytes` function used to convert AkdValues to bytestrings that get hashed. We will use a keyed hash (modeled as a random oracle) as a commitment here.

This means we need the server to hold a commitment_key server. The server needs to generate a proof for each label+value, and send this proof as a component of the lookup and update proofs. The client can then verify the proofs by supplying the raw value and checking that the commitment matches.

Construction:

proof = H(commitment_key, label, version, value)
commmitment = H(value, proof)

Resolves #123.

Note that we do a kindof hacky generation of the commitment key for now. Opened #184 to track a fix for this.

### DIFF
--- a/akd/src/utils.rs
+++ b/akd/src/utils.rs
@@ -11,10 +11,11 @@
 
 use crate::{
     node_state::{hash_label, NodeLabel},
+    storage::types::{AkdLabel, AkdValue},
     EMPTY_LABEL, EMPTY_VALUE,
 };
 use std::collections::HashSet;
-use winter_crypto::Hasher;
+use winter_crypto::{Digest, Hasher};
 
 // Builds a set of all prefixes of the input labels
 pub(crate) fn build_prefixes_set(labels: &[NodeLabel]) -> HashSet<NodeLabel> {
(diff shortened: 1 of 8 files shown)


akd/src/utils.rs
```diff
@@ -53,8 +54,50 @@ pub(crate) fn empty_node_hash_no_label<H: Hasher>() -> H::Digest {
...
+            &version.to_be_bytes(),
+            &i2osp_array(value),
+        ]
+        .concat(),
+    )
+}
+
+// Used by the server to produce a commitment for an AkdLabel, version, and AkdValue
+//
+// proof = H(commitment_key, label, version, value)
```
eozturk1, line 87: May I suggest adding here that `proof` is a nonce used to create a hiding and binding commitment using a cryptographic hash function?

eozturk1: For uniqueness, `AkdLabel` and `version` should be enough. Could this save us a few cycles?

kevinlewi: It could. My thinking was that by including the AkdValue, we will for sure have uniqueness, whereas perhaps without AkdValue there might be some edge case that allows for an attack. So this was "just to be safe". But if it is really not necessary, then I suppose we can omit it.

eozturk1: I can't think of one atm but adding `AkdValue` would certainly create a stronger bond between the nonce and the data it is created for. Let's keep it!

kevinlewi: In the latest commit, adjusted the comment here to include the detail about the proof acting as a nonce.

slawlor:
@kevinlewi I think we're generally in agreement on this PR that it's necessary and a good thing to do. I've looked at it and it seems fine to me, do you want to just move forward with stamping it or do you want an explicit review from @Jasleen1 ?

kevinlewi:
I think this has already been reviewed by @Jasleen1, so feel free to stamp it and move forward. Thanks for checking!

slawlor (approved):
LGTM!

PR #187, opened 2022-03-25
Protect storage key deserializations

slawlor:
This change adds a protection on storage-type key deserialization to assert that the first byte does indeed denote the same type of key that is trying to be deserialzed.

### DIFF
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -57,7 +57,7 @@ impl Storable for Azks {
     }
 
     fn key_from_full_binary(bin: &[u8]) -> Result<u8, String> {
-        if bin[0] != StorageType::Azks as u8 {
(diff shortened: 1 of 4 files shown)


akd/src/append_only_zks.rs
```diff
@@ -56,7 +56,10 @@ impl Storable for Azks {
         vec![StorageType::Azks as u8, *key]
     }
 
-    fn key_from_full_binary(_bin: &[u8]) -> Result<u8, String> {
+    fn key_from_full_binary(bin: &[u8]) -> Result<u8, String> {
+        if bin[0] != StorageType::Azks as u8 {
```
alexme22, line 60: nit: check for bin.len() > 0?

alexme22 (approved):

PR #188, opened 2022-03-29
Make VRF(Public|Private) key Cloneable

davidjattfb:
This implements the std::clone::Clone trait on the VRFPublicKey and VRFPrivateKey structs. In the case of the PublicKey, the ed25519_dalek::PublicKey already supports clone so the trait can be #derived. The ed25519_dalek::SecretKey does not, so a manual implementation is required. It simply creates a new SecretKey out of the bytes of the original.

### DIFF
--- a/akd/src/ecvrf/ecvrf_impl.rs
+++ b/akd/src/ecvrf/ecvrf_impl.rs
@@ -54,8 +54,17 @@ impl core::ops::Deref for VRFPrivateKey {
         &self.0
     }
 }
+
+impl std::clone::Clone for VRFPrivateKey {
+    fn clone(&self) -> Self {
+        // In theory, creating a key from bytes could be a DecodingError, except
+        // we just copied these bytes out of the source key, so ...
+        Self(ed25519_PrivateKey::from_bytes(self.0.as_bytes()).unwrap())
+    }
+}
+
 /// An ECVRF public key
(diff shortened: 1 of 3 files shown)


akd/src/ecvrf/ecvrf_impl.rs
```diff
@@ -54,8 +54,17 @@ impl core::ops::Deref for VRFPrivateKey {
         &self.0
     }
 }
+
+impl std::clone::Clone for VRFPrivateKey {
+    fn clone(&self) -> Self {
+        // In theory, creating a key from bytes could be a DecodingError, except
+        // we just copied these bytes out of the source key, so ...
+        Self(ed25519_PrivateKey::from_bytes(self.0.as_bytes()).unwrap())
```
davidjattfb, line 62: This feels a little icky, using `unwrap()` here. New to Rust, I couldn't find good guidance about how to handle an error occurring in `clone()`. It seems that getting an error here borders on impossible, seeing as it creates a key from the bytes of another key, so is the panic the right approach here?

slawlor: Yep exactly, here we would probably want a panic. Since it should theoretically be impossible. Unfortunately we can't really write "can't panic" :/

slawlor (approved):
LGTM!

Issue #186, opened 2022-03-22
Optimized azks implementation


Jasleen1:
The current implementation is based on a minimally optimized version of the construction of a History Tree data structure from the SEEMless paper (see figure 2 in this https://eprint.iacr.org/2018/607.pdf). This data structure stores every state ever of all tree nodes, leading to significant storage costs. We can optimize this away by making the following change: if the value stored at a given leaf in the tree was previously `val`, we now replace it with `hash(val, t)` where `t` is epoch when that leaf was first inserted. Now, in order to verify that a leaf was inserted at the correct time, a client only needs to check that the correct timestamp was hashed in the leaf. The auditors must still audit all consecutive epochs for the append only property and ensure that the epoch hashed in any new leaf is correct, in order to maintain the same security guarantees as SEEMless and our current implementation.

Fortunately, most of the APIs remain the same with only minor changes. At a high level this requires the following code changes:

1. `HistoryTreeNode` no longer needs to exist, we can just have a `TreeNode` with main APIs 
    (a) `insert_single_leaf(&mut self, leaf: Self)`: Inserts a leaf without a hash using a helper function.
    (b) `insert_single_leaf_with_hash(&mut self, leaf: Self)`: Inserts a leaf and updates the hash up the tree.
    (c) `insert_single_leaf_helper(&mut self, leaf: Self, hashing_flag: bool)`: Inserts a single leaf with or without hashing as needed.
    (d) `update_hash(&mut self)`: Updates the hash of this node and sets it at the parent, if relevant.
2. `Azks`: This should remain almost the same, except, the `get_append_only_proof` will probably need to be run as part of a call to `batch_insert_leaves`, which will need to keep track of the unchanged nodes. This is the main algorithmic complexity in this entire change. Additionally the proofs themselves will contain the epoch when some node was added, so the value of the leaf nodes should somehow also store the epoch when the leaf was created.
3. `Directory`: The only change here will be in the `audit` which will now only allow auditing the latest update. One design decision here is whether we would like to store the audit proofs for a few of the most recent epochs. 
4. The  data structure `LookupProof` should now include the epoch when a leaf was inserted whenever a membership proof is sent. The corresponding changes also need to be made to `AuditProof` and `KeyHistoryProof`.
5. All the corresponding client and auditor verification functions need to be modified.

Note that a similar data structure has been implemented and open sourced by Microsoft Research, which might be helpful. See: https://github.com/Microsoft/oZKS.

Jasleen1:
[...]

PR #189, opened 2022-04-14
Optimized azks

eozturk1:
See Issue #186.

With the removal of previous states of `HistoryTreeNode`, we don't have a need for keeping states for multiple epochs. In result, `HistoryTreeNode`, `HistoryNodeState` and `HistoryChildState` can simply be replaced with `TreeNode`.

This diff starts this conversion by adding the labels of its (1) left and right children, and (2) its hash to the `HistoryTreeNode` (not yet re-named). Actual hash updates will be added in the next diffs. Rather than keeping the node hashes at the parent, I will keep them at the node itself. 

NOTE: The PR builds with `cargo build` but not ready for testing. The purpose of the PR is to "get it out there" to get feedback rather than merging it.


slawlor (commented):
Some nits to consider. @eozturk1 are we waiting on @Jasleen1 or @kevinlewi to give it a process pass to see if you're on the right track? There's obviously some clippy and `rustfmt` issues, but functionally it makes sense to me I think.

eozturk1:
### Summary

This PR so far has removed the need for storing states-per-epoch and revised the `HistoryTreeNode` (referred to as node below) to contain the only (current) state (aka `hash`), significantly reducing the storage requirements.

### Important Changes
- Each node keeps a label of their left and right children, and its hash.
- For leaf nodes the hash is over the initial digest and the label.
- For non-leaf nodes, the hash is over the left and right child, and the label.
- For the root, we use the empty node hash: over the empty value and empty label.
- The `last_epoch` has been kept to be used in append-only proofs whereas `birth_epoch` is not needed anymore.
- For each call, `set_child` and `update_node_hash` writes the updated nodes to the storage for consistency. See below for optimizations over this.

### Next Steps:
- [ ] **Clean up:** We should remove the dead-code, debug output, unused `HistoryChildState` and `HistoryNodeState`, along with the storage layer code for these nodes.
- [ ] **Updating proofs:**: Since some hash calculations have been updated, the proofs need to be updated accordingly.
- [ ] **Clippy:** Make sure Clippy passes.
- [ ] **Merge with main (?):** At this point, we should be able to merge this in (possibly to a separate branch).
- [ ] **Addition of epoch to the leaf:** Leaves should include the latest epoch information in their initial `hash`. This will allow us to track when a leaf was inserted into the tree.
- [ ] **Enabling concurrent read/write operations:** During a sequencing operation a node might get updated. This means if a client requests a proof during sequencing, we might return an inconsistent proof (a mix of states from the current and next epochs). To circumvent this, we should keep two hashes in the node and update the second one with the new state while moving the current second state to the first (i.e., shift the state to the left).
- [ ] **Remove birth epoch:** Not needed anymore
- [ ] **Renaming `HistoryTreeNode` as `TreeNode`**: No history, only current.

### Improvements
- **Tests:** We should add more tests for small trees and nodes with non-uniform distribution to increase test coverage for tree creation.
- **Minimizing storage writes:** If a node is not written to the storage as soon as it is updated, a dirty-read may occur for consecutive operations. We might be able to minimize these writes by allowing short-term inconsistencies by restructuring the code. Another question to look into is how expensive are these instant-writes? (@afterdusk mentioned we use transactions to minimize the impact)
- **Tree Visualization:** Noticed it is hard to track the current state of the tree through logs. Maybe a visual representation could assist us in this?

kevinlewi:
Discussed in-person with @eozturk1 and @Jasleen1 , let's disable the currently-failing tests and fix clippy, so that we can land and @Jasleen1 can work on top of the changes here (filling in the rest of the checkboxes).

eozturk1:
I have disabled the failing tests and updated the next steps to list those tests.

kevinlewi:
Not sure what the integration test issue is exactly, but perhaps we can merge into a separate branch so that @Jasleen1 can start working on top of it. cc: @slawlor

PR #192, opened 2022-04-25
Adding verification of sequential nature of history proof

### LINKED ISSUE #191: AKD Clients should prove the sequential nature of a history proof
To identify "missing" history, client verifiers of a key history proof should confirm that the proof updates are sequential (i.e. the version numbers increase by 1 starting from 0 (or 1, need to confirm initial version number))

This is a small verification change, and the information is available in the clients already, but presently would be a second step and should simply be included inside the `key_history_verify` call.

slawlor:
We calculate the expected version number given a maximum version, and work backwards assuring that the versions are sequentially decreasing until the minimum version. This assures there are no gaps in the history 

Resolves #191

### DIFF
--- a/akd/src/client.rs
+++ b/akd/src/client.rs
@@ -167,9 +167,22 @@ pub fn key_history_verify<H: Hasher>(
     allow_tombstones: bool,
 ) -> Result<Vec<bool>, AkdError> {
     let mut tombstones = vec![];
+    let some_max_version = proof.proofs.get(0).map(|item| item.version);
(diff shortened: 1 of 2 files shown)


kevinlewi (approved):
LGTM! (modulo clippy errors)

slawlor:
> LGTM! (modulo clippy errors)

Yep just needed new merge from main! (Seems clippy was broken in master, maybe a rustup update *shrug*). Thanks!

PR #194, opened 2022-05-10
Migrate from structopt to clap 3 and fix random string generation

afterdusk:
`structopt` is now in [maintenance mode](https://docs.rs/structopt/latest/structopt/#maintenance) and has been superseded by `clap` v3. This PR migrates us to `clap`.

In addition, the `get_random_str` helper function did not properly generate 32-byte strings. Instead of creating a 32-char string, it created a string of an array of 32 bytes (e.g. `"[244, 198, 197, 130, ...]"`). This PR fixes the function to return a random 32-char string which can later be interpreted as 32 bytes.


slawlor (approved):
LGTM! Thanks for bringing us up to the modern age :)

PR #198, opened 2022-05-18
Skip duplicate values for a user if they're equal in a publish

### LINKED ISSUE #197: If the user's value is identical to the most recent value, skip publishing that user's key value.
In some cases, there may be a concurrency consideration gathering keys together to form a batch publish. This is especially true if some kind of queueing mechanism is necessary to gather a large batch of changes and periodically publish an epoch.

In this scenario, publication of an epoch may not be performant on a user-by-user key change basis as the cost of doing a tree-walk in the storage layer is non-trivial. In a batch scenario, we can do a breadth first search to gather all the nodes that we'll touch in an epoch publication, so it's far more performance at scale. However this gathering of batches may have some concurrency considerations where we may reprocess keys (in some odd failure scenarios).

Therefore as a minor protection against this scenario, we should add the ability to determine if the key value is the same as the most recent value, and skip publication in this scenario.

slawlor:
When a user requests publishing a new value to the directory, check that it's not ALREADY the latest value. This helps solve some corner cases around potential race conditions batching changes from upstream queue-like services. See Issue #197 for context.

This pull request adds the described functionality as well as a new test around that functionality. It also cleans up existing tests where we apparently were double-publishing the same value for users sequentially.

As a note! Publishing value "a", then "b", then "a" again **is** valid! This is a set of changes. However publishing "a" then "a" again is a no-op in this case, as nothing has changed.

Resolves #197

### DIFF
--- a/akd/src/directory.rs
+++ b/akd/src/directory.rs
@@ -208,9 +208,9 @@ impl<S: Storage + Sync + Send, V: VRFKeyStorage> Directory<S, V> {
             info!("After filtering for duplicated user information, there is no publish which is necessary (0 updates)");
             // The AZKS has not been updated/mutated at this point, so we can just return the root hash from before
             let root_hash = current_azks
-                .get_root_hash_at_epoch::<_, H>(&self.storage, current_epoch)
+                .get_root_hash::<_, H>(&self.storage)
                 .await?;
-            return Ok(EpochHash(current_azks.get_latest_epoch(), root_hash));
+            return Ok(EpochHash(current_epoch, root_hash));
         }
 
         if let false = self.storage.begin_transaction().await {
--- a/akd/src/directory.rs
+++ b/akd/src/directory.rs
@@ -208,7 +208,7 @@ impl<S: Storage + Sync + Send, V: VRFKeyStorage> Directory<S, V> {
             info!("After filtering for duplicated user information, there is no publish which is necessary (0 updates)");
             // The AZKS has not been updated/mutated at this point, so we can just return the root hash from before
             let root_hash = current_azks
-                .get_root_hash_at_epoch::<_, H>(&self.storage, next_epoch)
+                .get_root_hash_at_epoch::<_, H>(&self.storage, current_epoch)
                 .await?;
             return Ok(EpochHash(current_azks.get_latest_epoch(), root_hash));
         }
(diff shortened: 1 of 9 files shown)


slawlor:
Additionally as a small note: We have to bump the minor version since this "breaks" the storage trait contract. So I've migrated published versions from 0.5.5 -> 0.6.0 for when we do the next crate publish

akd/src/directory.rs
```diff
@@ -233,7 +246,7 @@ impl<S: Storage + Sync + Send, V: VRFKeyStorage> Directory<S, V> {
             .get_root_hash_at_epoch::<_, H>(&self.storage, next_epoch)
             .await?;
 
-        Ok(EpochHash(current_epoch, root_hash))
+        Ok(EpochHash(next_epoch, root_hash))
```
afterdusk, line 249: Oops - thanks for catching this!

akd/src/directory.rs
```diff
@@ -200,6 +204,15 @@ impl<S: Storage + Sync + Send, V: VRFKeyStorage> Directory<S, V> {
         }
         let insertion_set: Vec<Node<H>> = update_set.to_vec();
 
+        if insertion_set.is_empty() {
+            info!("After filtering for duplicated user information, there is no publish which is necessary (0 updates)");
+            // The AZKS has not been updated/mutated at this point, so we can just return the root hash from before
+            let root_hash = current_azks
+                .get_root_hash_at_epoch::<_, H>(&self.storage, current_epoch)
+                .await?;
+            return Ok(EpochHash(current_azks.get_latest_epoch(), root_hash));
```
afterdusk, line 213: Any reason we are requesting the latest epoch instead of using `current_epoch`?

akd/src/tests.rs
```diff
@@ -778,6 +778,52 @@ async fn test_tombstoned_key_history() -> Result<(), AkdError> {
...
+
+    // still epoch 1 because the value is the same
+    let epoch1_hash_again = akd
+        .publish::<Blake3>(vec![(
+            AkdLabel::from_utf8_str("user"),
+            AkdValue::from_utf8_str("value"),
+        )])
+        .await?;
+
+    assert_eq!(1u64, epoch1_hash_again.0);
```
afterdusk, line 806: Can we also assert that the hash is the same as epoch 1's?

akd/src/storage/mod.rs
```diff
@@ -151,7 +151,7 @@ pub trait Storage: Clone {
         &self,
         keys: &[types::AkdLabel],
```
afterdusk, line 152: Not related to your change, but could we rename `keys` to `usernames` to be consistent with `get_user_data` and `get_user_state`?

afterdusk (approved):
LGTM, thanks for adding this!

PR #202, opened 2022-05-31
Moving parsers out of the integration tests to independent crate

slawlor:
Creating an independent fixture generation (and parser) which is separate from the integration tests

This allows us to import test fixture generators and parsers more easily for downstream dependencies


afterdusk (approved):
Thanks for the refactor!

PR #204, opened 2022-06-02
Bump version number to v0.6.2

afterdusk:
This version bump is to create a release that has a commit hash at which the `akd_test_tools` crate exists. This will resolve some downstream import issues.


kevinlewi (approved):
LGTM

PR #211, opened 2022-06-21
Eozturk1/optimized azks

Jasleen1:
Merged ozks update with main and updated the corresponding failing tests in `akd_test_tools`.


akd/src/auditor.rs
```diff
@@ -13,36 +13,58 @@ use winter_crypto::Hasher;
 
 use crate::{
     errors::{AkdError, AzksError},
-    proof_structs::AppendOnlyProof,
+    proof_structs::{AppendOnlyProof, SingleAppendOnlyProof},
     storage::memory::AsyncInMemoryDatabase,
     Azks,
 };
 
+// FIXME: Need to add error handling
```
slawlor, line 21: nit: Real fixme?

akd/src/auditor.rs
```diff
@@ -13,36 +13,58 @@ use winter_crypto::Hasher;
...
-    proof: AppendOnlyProof<H>,
+pub async fn verify_consecutive_append_only<H: Hasher + Send + Sync>(
+    proof: &SingleAppendOnlyProof<H>,
     start_hash: H::Digest,
     end_hash: H::Digest,
+    epoch: u64,
 ) -> Result<(), AkdError> {
-    let unchanged_nodes = proof.unchanged_nodes;
-    let inserted = proof.inserted;
+    // FIXME: Need to get rid of the clone here.
```
slawlor, line 48: nit: why? What's wrong with the clone?

akd/src/client.rs
```diff
@@ -211,12 +299,11 @@ fn verify_single_update_proof<H: Hasher>(
     let existence_vrf_proof = proof.existence_vrf_proof;
     let existence_at_ep_ref = &proof.existence_at_ep;
     let existence_at_ep = existence_at_ep_ref;
+    // FIXME: Why does this need a reference??
```
slawlor, line 302: nit: ?

akd/src/client.rs
```diff
@@ -350,25 +390,33 @@ fn build_and_hash_layer<H: Hasher>(
...
+    let new_hash = H::merge(&[hashes[0], hashes[1]]);
+    H::merge(&[new_hash, hash_label::<H>(parent_label)])
 }
 
-fn hash_plaintext_value<H: Hasher>(value: &crate::AkdValue, proof: &[u8]) -> H::Digest {
+fn hash_leaf_with_value<H: Hasher>(
+    value: &crate::AkdValue,
+    epoch: u64,
+    proof: &[u8],
+    // FIXME get rid of this argument
```
slawlor, line 412: nit: Can we address all the FIXME's?

akd/src/proof_structs.rs
```diff
@@ -119,14 +120,33 @@ impl<H: Hasher> Clone for NonMembershipProof<H> {
 )]
 #[cfg_attr(feature = "serde_serialization", serde(bound = ""))]
 pub struct AppendOnlyProof<H: Hasher> {
+    /// Proof for a single epoch being append-only
+    pub proofs: Vec<SingleAppendOnlyProof<H>>,
+    /// Epochs over which this audit is being performed
+    pub epochs: Vec<u64>,
```
slawlor, line 126: nit: These need to be a contiguous range right? Maybe it's better to have start + end epochs, since we don't need all the numbers in the middle

akd/src/tests.rs
```diff
@@ -439,17 +480,11 @@ async fn test_read_during_publish() -> Result<(), AkdError> {
         AkdLabel::from_utf8_str("hello"),
         lookup_proof,
     )?;
-
+    // FIXME: uncomment
```
slawlor, line 483: nit: remove?

akd/src/tests.rs
```diff
@@ -354,7 +385,8 @@ async fn test_simple_audit() -> Result<(), AkdError> {
     Ok(())
 }
 
-#[tokio::test]
+// #[tokio::test]
+#[allow(dead_code)]
```
slawlor, line 389: nit: can we re-enable all `dead_code` tests?

akd/src/utils.rs
```diff
@@ -67,19 +68,19 @@ pub(crate) fn i2osp_array(input: &[u8]) -> Vec<u8> {
...
-    H::hash(
-        &[
-            commitment_key,
-            &i2osp_array(label),
-            &version.to_be_bytes(),
-            &i2osp_array(value),
-        ]
-        .concat(),
-    )
+    // H::hash(
```
slawlor, line 74: nit: defunct?

akd_client/src/tests.rs
```diff
@@ -67,6 +70,8 @@ where
     digest_vec
 }
 
+// FIXME: may want to get rid of this altogether
+#[allow(unused)]
```
slawlor, line 74: nit: if it's unused, can we?

akd_client/src/verify.rs
```diff
@@ -100,11 +114,24 @@ fn verify_nonmembership(
     Ok(verified)
 }
 
+// FIXME might want to remove.
+#[allow(unused)]
```
slawlor, line 118: nit: Can we remove this if it's actually unused?

integration_tests/src/test_util.rs
```diff
@@ -240,6 +371,7 @@ pub(crate) async fn test_lookups<S: akd::storage::Storage + Sync + Send, V: VRFK
 // Reset MySQL database by logging metrics which resets the metrics, and flushing cache.
 // These allow us to accurately assess the additional efficiency of
 // bulk lookup proofs.
+#[allow(unused)]
```
slawlor, line 374: nit: not used?

akd/src/auditor.rs
Jasleen1: This is left for my cleanup PR but I suspect there may be a way to do this using references that doesn't cause the borrow-checker to complain. I at least want to _try_ to reduce clones since that operation is less efficient.

akd/src/tests.rs
Jasleen1: This is the only test that is disabled. We discussed offline that this should be addressed in another PR, likely, since this failure is due to not storing hashes from the current completed epoch while updating. See https://github.com/novifinancial/akd/issues/203

slawlor (approved):

PR #214, opened 2022-06-29
Fix example test fixture

afterdusk:
Noticed that the example test fixture file `test.yaml` was not in a good state post-Ozks merge - this PR fixes it.

This involves serializing `TreeNode` hashes to hex with `serde` (so that they will be formatted as hex strings in the fixture file) and re-generating the fixture file with the following command in the `akd_test_tools` directory:

```
akd_test_tools % cargo run -- \
  --epochs 10 \
  --max_updates 5 \
  --capture_states 9 10 \
  --capture_deltas 10 \
  --out src/fixture_generator/examples/test.yaml
```


slawlor (approved):
LGTM!

Issue #213, opened 2022-06-28
Enable concurrent read/write operations in oZKS


eozturk1:
During a sequencing operation a node might get updated. This means if a client requests a proof during sequencing, we might return an inconsistent proof (a mix of states from the current and next epochs). A solution to this is to keep duplicate information about a node but this can be improved further.

eozturk1:
I'll work on this @Jasleen1 after the micro benchmark tests.

PR #216, opened 2022-07-19
Addressing the concurrency access of the storage layer mid-update

slawlor:
Added TreeNodeWithPreviousValue to handle keeping the "previous" state of a tree node. This allows us to access the previous value when a node might be in a future state, since the entire tree hasn't completed update yet.

Related: #213

### DIFF
--- a/akd/src/storage/types.rs
+++ b/akd/src/storage/types.rs
@@ -9,7 +9,7 @@
 #[cfg(feature = "serde_serialization")]
 use crate::serialization::{bytes_deserialize_hex, bytes_serialize_hex};
 use crate::storage::Storable;
-use crate::tree_node::{NodeType, TreeNode};
+use crate::tree_node::{NodeType, TreeNode, TreeNodeWithPreviousValue};
 use crate::{Azks, NodeLabel};
 use std::convert::TryInto;
 
@@ -210,7 +210,7 @@ pub enum DbRecord {
     /// An Azks
     Azks(Azks),
     /// A TreeNodeNode
-    TreeNode(TreeNode),
+    TreeNode(TreeNodeWithPreviousValue),
     /// The state of the value for a particular key.
     ValueState(ValueState),
 }
@@ -271,16 +271,44 @@ impl DbRecord {
         left_child: Option<NodeLabel>,
         right_child: Option<NodeLabel>,
         hash: [u8; 32],
-    ) -> TreeNode {
-        TreeNode {
-            label: NodeLabel::new(label_val, label_len),
-            last_epoch,
-            least_descendent_ep,
-            parent: NodeLabel::new(parent_label_val, parent_label_len),
-            node_type: NodeType::from_u8(node_type),
-            left_child,
-            right_child,
-            hash,
+        p_last_epoch: Option<u64>,
+        p_least_descendent_ep: Option<u64>,
+        p_parent_label_val: Option<[u8; 32]>,
+        p_parent_label_len: Option<u32>,
+        p_node_type: Option<u8>,
+        p_left_child: Option<NodeLabel>,
+        p_right_child: Option<NodeLabel>,
+        p_hash: Option<[u8; 32]>,
+    ) -> TreeNodeWithPreviousValue {
+        let label = NodeLabel::new(label_val, label_len);
+        let p_node = match (p_last_epoch, p_least_descendent_ep, p_parent_label_val, p_parent_label_len, p_node_type, p_hash) {
+            (Some(a), Some(b), Some(c), Some(d), Some(e), Some(f)) => {
+                Some(TreeNode {
+                    label,
+                    last_epoch: a,
+                    least_descendent_ep: b,
+                    parent: NodeLabel::new(c, d),
+                    node_type: NodeType::from_u8(e),
+                    left_child: p_left_child,
+                    right_child: p_right_child,
+                    hash: f,
+                })
+            },
+            _ => None
+        };
+        TreeNodeWithPreviousValue {
+            label,
+            latest_node: TreeNode {
+                label,
+                last_epoch,
+                least_descendent_ep,
+                parent: NodeLabel::new(parent_label_val, parent_label_len),
+                node_type: NodeType::from_u8(node_type),
+                left_child,
+                right_child,
+                hash,
+            },
+            previous_node: p_node,
         }
     }
 
--- a/akd/src/storage/types.rs
+++ b/akd/src/storage/types.rs
@@ -210,7 +210,7 @@ pub enum ValueStateRetrievalFlag {
 pub enum DbRecord {
     /// An Azks
     Azks(Azks),
-    /// A TreeNodeNode
+    /// A TreeNode
     TreeNode(TreeNodeWithPreviousValue),
     /// The state of the value for a particular key.
     ValueState(ValueState),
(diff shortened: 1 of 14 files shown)


akd/src/tree_node.rs
```diff
@@ -89,6 +224,56 @@ pub struct TreeNode {
...
+        &self,
+        storage: &S,
+    ) -> Result<(), StorageError> {
+        // retrieve the higest node properties, at a previous epoch than this one. If we're modifying "this" epoch, simply take it as no need for a rotation.
+        // When we write the node, with an updated epoch value, we'll rotate the stored value and capture the previous
+        let target_epoch = match self.last_epoch {
+            e if e > 0 => e - 1,
+            other => other,
+        };
+        let previous = match TreeNodeWithPreviousValue::get_from_storage(
```
slawlor, line 239: We want to retrieve the previous ```latest_node``` value, so we want to investigate where epoch - 1. When a request comes in to write the node with a future epoch, epoch - 1 will be the latest node in storage and we'll do a shift-left. 

The get call should ideally utilize a cached value, so this should be safe to call repeatedly.

akd/src/tree_node.rs
```diff
@@ -49,6 +49,141 @@ impl NodeType {
...
+    }
+
+    pub(crate) async fn get_from_storage<S: Storage + Send + Sync>(
+        storage: &S,
+        key: &NodeKey,
+        current_epoch: u64,
+    ) -> Result<TreeNode, StorageError> {
+        match storage.get::<Self>(key).await? {
+            DbRecord::TreeNode(node) => {
+                if node.latest_node.last_epoch > current_epoch {
```
eozturk1, line 137: We take this path if 

- There is a publish in progress + the node with this key has been updated + a proof is being generated.

Else is taken if

- There is a publish in progress  + the publish operation incremented the epoch (but didn't write it to the database) and requested the most up-to-date node state using the incremented epoch, or
- There is no publish in progress

Could we add above cases as comments to the `if/else`?

akd/src/storage/types.rs
```diff
@@ -271,16 +272,49 @@ impl DbRecord {
...
-        TreeNode {
-            label: NodeLabel::new(label_val, label_len),
-            last_epoch,
-            least_descendent_ep,
-            parent: NodeLabel::new(parent_label_val, parent_label_len),
-            node_type: NodeType::from_u8(node_type),
-            left_child,
-            right_child,
-            hash,
+        p_last_epoch: Option<u64>,
```
eozturk1, line 275: We should probably change the name to `build_tree_node_with_previous_value`.

akd_test_tools/src/fixture_generator/examples/test.yaml
```diff
@@ -30,1239 +30,2205 @@ epoch: 9
...
-      last_epoch: 4
-      least_descendent_ep: 4
-      parent:
-        val: F000000000000000000000000000000000000000000000000000000000000000
-        len: 7
-      node_type: Leaf
-      left_child: ~
-      right_child: ~
-      hash: D83CA2E6CA89C340ADE2B533CCAA2738D94AF033C703E14A419EBBDDE4E748EF
+        val: 61B6CF9432424B3139006A631373FA7C62226FEBCD8942868BC2EA1812C3919A
```
eozturk1, line 33: We could forget updating this file. Should we keep it checked in or somehow auto-build?

eozturk1 (approved):

eozturk1 (approved):
Awesome work Sean! Left a few comments, nothing major.

akd_test_tools/src/fixture_generator/examples/test.yaml
afterdusk: Ideally changes to the Storage layer should be accompanied with a minor version bump, which the [`akd_test_tools` crate specifically depends on](https://github.com/novifinancial/akd/blob/7958a46989b5e5d4f633e645a47fff0945350d85/akd_test_tools/Cargo.toml#L24). So long as the person updating `akd_test_tool`'s Cargo file checks for Storage layer changes and re-generates this file accordingly, it should never become "incorrect" for a given akd version.

akd_test_tools/src/fixture_generator/reader/yaml.rs
```diff
@@ -83,7 +84,8 @@ impl YamlFileReader {
                     return Ok(doc);
                 }
                 Some(Ok(line)) => {
-                    doc.push_str(&format!("{}\n", line));
+                    // avoid the extra allocation call with a format!
+                    let _ = writeln!(doc, "{}", line);
```
afterdusk, line 88: 👍

afterdusk (approved):
This looks great! No concerns from me as well.

akd/src/storage/types.rs
```diff
@@ -206,11 +206,12 @@ pub enum ValueStateRetrievalFlag {
     feature = "serde_serialization",
     derive(serde::Deserialize, serde::Serialize)
 )]
+#[allow(clippy::large_enum_variant)]
 pub enum DbRecord {
     /// An Azks
     Azks(Azks),
     /// A TreeNodeNode
-    TreeNode(TreeNode),
+    TreeNode(TreeNodeWithPreviousValue),
```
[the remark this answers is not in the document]
slawlor: I thought about that, but really it's still storing a "tree" node, while there are 2 only 1 is active. It seemed like a large refactoring for little to no benefit beyond having the enum field match the stored type name. If you have strong opinions one way or the other I'm happy to do the renaming though!

akd_test_tools/src/fixture_generator/examples/test.yaml
slawlor: I think I'm going to roll to something other than 0.6.X (i.e. bump to 0.7) since we're doing a pretty big storage contract break here. But yes a version bump was on my radar, thanks for reminding me!

akd/src/tree_node.rs
```diff
@@ -49,6 +49,141 @@ impl NodeType {
...
+        storage: &S,
+    ) -> Result<(), StorageError> {
+        storage.set(DbRecord::TreeNode(self.clone())).await
+    }
+
+    pub(crate) async fn get_from_storage<S: Storage + Send + Sync>(
+        storage: &S,
+        key: &NodeKey,
+        current_epoch: u64,
+    ) -> Result<TreeNode, StorageError> {
```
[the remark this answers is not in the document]
slawlor: Yeah so for the normal get, that does make sense, but for the batch get it'd require a 2nd iteration of the vector since we'd only check if which node to retrieve in the TreeNode impl. I'm going to rename the functions however to be more logical, they were bad copy-pasta names initially as I was trying to work out the best way to do this shift-left logix.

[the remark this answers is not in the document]
slawlor: This is what I'm going to modify it to: 
```rust
pub(crate) async fn get_appropriate_tree_node_from_storage<S: Storage + Send + Sync>(
      storage: &S,
      key: &NodeKey,
      target_epoch: u64,
  ) -> Result<TreeNode, StorageError> {
      ...
  }

  pub(crate) async fn batch_get_appropriate_tree_node_from_storage<S: Storage + Send + Sync>(
      storage: &S,
      keys: &[NodeKey],
      target_epoch: u64,
  ) -> Result<Vec<TreeNode>, StorageError> {
      ...
  }
```

akd_test_tools/src/fixture_generator/examples/test.yaml
eozturk1: It's possible to glue a few Github actions together to re-gen this file if there is a change in the storage layer. Let me know whether folks would be interested in such an automated solution.

akd/src/storage/types.rs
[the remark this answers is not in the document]
eozturk1: I think this is fine as long as we provide enough documentation in the code as to which one to use and when.

PR #218, opened 2022-07-25
Remove unused module node_state.rs

afterdusk:
[`node_state.rs`](https://github.com/novifinancial/akd/blob/97b2e90589083fef4600f732618715fac29a24f2/akd/src/storage/node_state.rs) is an unused, stale copy of [`node_label.rs`](https://github.com/novifinancial/akd/blob/97b2e90589083fef4600f732618715fac29a24f2/akd/src/node_label.rs).

Verified that no other instances of `node_state` found in `*.rs` files:
```
$ grep -l -r --exclude-dir target --include \*.rs node_state .
```


slawlor (approved):
Thanks for the cleanup!

PR #219, opened 2022-07-28
[RFC] Remove epochs vector from HistoryProof struct

afterdusk:
HistoryProof contains a vector of `UpdateProof`s and a vector of epochs:

https://github.com/novifinancial/akd/blob/97b2e90589083fef4600f732618715fac29a24f2/akd/src/proof_structs.rs#L227-L234

An `UpdateProof` itself contains an epoch:

https://github.com/novifinancial/akd/blob/97b2e90589083fef4600f732618715fac29a24f2/akd/src/proof_structs.rs#L274-L276

From what I can tell, these two fields are populated with the same value:

https://github.com/novifinancial/akd/blob/97b2e90589083fef4600f732618715fac29a24f2/akd/src/directory.rs#L429-L436

- An update proof is created (line 429) with `create_single_update_proof()` and appended to the end of `update_proofs` (line 430)
- `user_state.epoch` is appended to the end of `epochs` vector (line 436)

https://github.com/novifinancial/akd/blob/97b2e90589083fef4600f732618715fac29a24f2/akd/src/directory.rs#L716-L721

- In `create_single_update_proof()`, `user_state.epoch` is assigned to the `epoch` attribute (line 721)

As such, I propose removing the `epochs` vector of the `HistoryProof` struct in favor of relying on the `epoch` attribute of the contained `UpdateProof` structs.

### DIFF
--- a/akd/src/client.rs
+++ b/akd/src/client.rs
@@ -203,7 +203,7 @@ pub fn key_history_verify<H: Hasher>(
         }
     }
 
-    // Check that all the individual update proofs check
+    // Verify all individual update proofs
     let mut maybe_previous_update_epoch = None;
     for update_proof in proof.update_proofs.into_iter() {
         last_version = if update_proof.version > last_version {
--- a/akd/src/client.rs
+++ b/akd/src/client.rs
@@ -188,17 +188,6 @@ pub fn key_history_verify<H: Hasher>(
         )));
     }
 
-    // Make sure this proof has the same number of epochs as update proofs.
-    if num_proofs != proof.epochs.len() {
-        return Err(AkdError::Directory(DirectoryError::VerifyKeyHistoryProof(
-            format!(
-                "The number of epochs included in the proofs for user {:?} 
-                did not match the number of update proofs!",
-                akd_key
-            ),
-        )));
-    }
-
     // Check that the sent proofs are for a contiguous sequence of decreasing versions
     for count in 0..num_proofs {
         if count > 0 {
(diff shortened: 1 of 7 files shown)


akd/src/client.rs
```diff
@@ -215,34 +204,35 @@ pub fn key_history_verify<H: Hasher>(
     }
 
     // Check that all the individual update proofs check
```
eozturk1, line 206: Can we add a comment what the actual `check` here is? Or better yet move it to a separate function with an apropos name.

eozturk1: Should we also create a test for it?

eozturk1 (commented):
LGTM, added a possible improvement. Feel free to create a task for creating the new test. @Jasleen1, could you please confirm we could remove the epochs?

Is there a performance concern or something to that end?

akd/src/client.rs
Jasleen1: We can remove these epochs for sure -- just need to access them and make sure we do all the same checks as we are doing with the listed epochs. Thanks so much for noticing this!

eozturk1: Separate function should also help with the visibility of the checks that are being done here.

Jasleen1 (approved):

akd/src/client.rs
afterdusk: There were two checks:
1. Checking that the listed epochs are the same length as the number of update proofs. This check is now obsolete because the epochs are within the update proofs.
2. Listed epochs should be decreasing in value - this check has been ported to act on update proofs.

So we should be good! I've updated this comment to be a bit clearer that this loop verifies single update proofs.

I've also added a simple test asserting failure if the update proofs are in the wrong order (decreasing versions and epochs).

eozturk1: Awesome! Thank you @afterdusk! :) Feel free to merge after CI!

eozturk1 (approved):

PR #223, opened 2022-08-05
Fix significant cache utilization drop for TreeNodes

eozturk1:
Recently we enabled concurrent reads (proof generations) and writes (publish) by keeping also the previous value of a `TreeNode`. During experiments, we noticed that the cache utilization significantly dropped. For comparison, this is pre-:

```
MySQL writes: 887, MySQL reads: 16, Time read: 0.076431639 s, Time write: 0.921964539 s
        Tree size: 499
        Value state count: 150
Read call stats: [("get_direct:~Azks", 1), ("get_user_data~", 2), ("get_user_state_versions~", 3), ("get_user_state~", 10)]
Write call stats: [("internal_batch_set~", 9), ("internal_set~", 2)]
```

Post:

```
MySQL writes: 878, MySQL reads: 515, Time read: 0.992529763 s, Time write: 2.507405077 s
        Tree size: 499
        Value state count: 150
Read call stats: [("get_direct:~Azks", 1), ("get_direct:~TreeNode", 499), ("get_user_data~", 2), ("get_user_state_versions~", 3), ("get_user_state~", 10)]
Write call stats: [("internal_batch_set~", 9), ("internal_set~", 2)]
```
Note the **("get_direct:~TreeNode", 499)** change.

After some digging, I noticed that this is due to for each node we write we look up the previous version of the node. However, this is unnecessary for **new** nodes since their previous value is already known and is `None`. More concerningly, all these calls hit the database -- since this information is not in the cache either.

This change ensures that (1) we don't go to the database for previous values of new nodes (and just return `None` for them), (2) new nodes are always saved so they can be fetched from the cache.

For comparison, new values:

```
MySQL writes: 891, MySQL reads: 16, Time read: 0.072592335 s, Time write: 2.734630483 s
        Tree size: 499
        Value state count: 150
Read call stats: [("get_direct:~Azks", 1), ("get_user_data~", 2), ("get_user_state_versions~", 3), ("get_user_state~", 10)]
Write call stats: [("internal_batch_set~", 9), ("internal_set~", 3)]
```

### DIFF
--- a/akd/src/tree_node.rs
+++ b/akd/src/tree_node.rs
@@ -244,10 +244,12 @@ impl TreeNode {
         &self,
         storage: &S,
     ) -> Result<(), StorageError> {
-        self.save_to_storage(storage, false).await
+        self.write_to_storage_impl(storage, false).await
     }
 
-    async fn save_to_storage<S: Storage + Send + Sync>(
+    /// Internal function to be used for storage operations. If a node is new (i.e., is_new_node=true), the node's previous version
+    /// will be used as None without the cost of finding this information in the cache or worse yet in the database.
+    async fn write_to_storage_impl<S: Storage + Send + Sync>(
         &self,
         storage: &S,
         is_new_node: bool,
@@ -371,7 +373,7 @@ impl TreeNode {
             right_child,
             hash,
         };
-        new_node.save_to_storage(storage, true).await?;
+        new_node.write_to_storage_impl(storage, true).await?;
         Ok(new_node)
     }
 
--- a/akd/src/tree_node.rs
+++ b/akd/src/tree_node.rs
@@ -587,8 +587,7 @@ impl TreeNode {
                 .update_node_hash::<_, H>(storage, epoch, exclude_ep)
                 .await?;
         } else {
-            // If no hashing, we need to manually save the nodes.
-            new_leaf.write_to_storage(storage).await?;
+            // If no hashing, we need to manually save the nodes (new leaf already saved above).
             new_node.write_to_storage(storage).await?;
             parent.write_to_storage(storage).await?;
         }
--- a/akd/src/tree_node.rs
+++ b/akd/src/tree_node.rs
@@ -879,8 +879,8 @@ pub(crate) fn optional_child_state_hash<H: Hasher>(
     }
 }
 
-/// Retrieve an empty root node
-pub async fn get_empty_root<H: Hasher, S: Storage + Sync + Send>(
+/// Create an empty root node.
+pub async fn create_empty_root<H: Hasher, S: Storage + Sync + Send>(
     storage: &S,
     ep: Option<u64>,
     least_descendant_ep: Option<u64>,
@@ -909,8 +909,8 @@ pub async fn get_empty_root<H: Hasher, S: Storage + Sync + Send>(
     Ok(node)
 }
 
-/// Get a specific leaf node
-pub async fn get_leaf_node<H: Hasher, S: Storage + Sync + Send>(
+/// Create a specific leaf node.
+pub async fn create_leaf_node<H: Hasher, S: Storage + Sync + Send>(
     storage: &S,
     label: NodeLabel,
     value: &H::Digest,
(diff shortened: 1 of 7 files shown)


akd/src/tree_node.rs
```diff
@@ -258,16 +266,24 @@ impl TreeNode {
             e if e > 0 => e - 1,
```
Jasleen1, line 266: Why do we use only `e-1` here? Shouldn't this be the previous time this node was modified? A node could remain untouched for a period of time.

Jasleen1 (approved):
See minor comment but the tests pass and otherwise looks great. Thank you!

akd_mysql/src/mysql.rs
```diff
@@ -911,6 +911,10 @@ impl Storage for AsyncMySqlDatabase {
         )
         .await;
 
+        if St::data_type() == StorageType::TreeNode {
+            println!("Get direct tree node.");
+        }
+
```
afterdusk, line 917: Did you intend to remove this?

akd/src/tree_node.rs
```diff
@@ -243,6 +243,14 @@ impl TreeNode {
     pub(crate) async fn write_to_storage<S: Storage + Send + Sync>(
         &self,
         storage: &S,
+    ) -> Result<(), StorageError> {
+        self.save_to_storage(storage, false).await
+    }
+
+    async fn save_to_storage<S: Storage + Send + Sync>(
```
afterdusk, line 250: Suggestion: `write_to_storage_impl` instead of `save_to_storage`? It's not quite clear what the difference is between `write_to_storage` and `save_to_storage`, but with the `impl` suffix it's clear that one is a wrapper of another.

akd/src/tree_node.rs
afterdusk: I was confused by this previously too (maybe `target_epoch` can have a better name), but this value is simply used to compare against the last epoch stored by a node. If `target_epoch` is smaller, the older set of node attributes are used.


https://github.com/novifinancial/akd/blob/97b2e90589083fef4600f732618715fac29a24f2/akd/src/tree_node.rs#L122-L144

afterdusk (changes requested):
Nice work investigating this issue and for the clean fix!

I do have a request - could we keep the `new()` and `get_*()` functions stateless and have the callers take care of writing? I would prefer to keep write operations to functions that explicitly express modification (i.e. `insert_*()` functions)

akd_mysql/src/mysql.rs
eozturk1: Good catch! The byproduct of fixing the cache issue was that this message didn't appear anymore, so was missed. Thanks for pointing out!

akd/src/tree_node.rs
eozturk1: Updated names and added docs for the `impl` version.

eozturk1:
> Nice work investigating this issue and for the clean fix!
> 
> I do have a request - could we keep the `new()` and `get_*()` functions stateless and have the callers take care of writing? I would prefer to keep write operations to functions that explicitly express modification (i.e. `insert_*()` functions)

I've thought about this approach. The issue is that it requires the developer to be vigilant about making sure that they wrote every root/leaf they generate to the storage. This is easy to miss and will incur performance hit without getting noticed. Point in case, the current one went unnoticed for a while and required further tracking it down. Furthermore, the writes should happen with the correct flag (`is_new_node=true` at node creation, `false` later) otherwise we generate the previous state of a node wrong. We could provide a separate API for it but in my opinion, it is better to avoid it. What do you think?

If the concern is keeping the function names adhering to the conventions, could we rename it as `create_new` or similar, and document the behavior?

eozturk1:
After `git pull origin` and fixing the conflicts, the changes in the main repo is showing up as changes in these diffs. Should have rebased maybe?

afterdusk (approved):
@eozturk1 Your line of reasoning makes sense - could we meet in the middle? Rename the following functions and have them write to storage but move the writing out of `TreeNode::new()` :)
- `get_empty_root` -> `create_empty_root_node`
- `get_leaf_node` -> `create_leaf_node`

There might be future scenarios where we want to initialize a `TreeNode` without writing to storage (e.g. in tests), and a `new()` function without side-effects would be nice.

In the above functions, we are explicitly initializing `TreeNode` objects, so it should obvious that `is_new_node` should be false when writing to storage.

If you feel strongly about keeping the writing in `TreeNode::new()`, then let's at least do the rename! LGTM, feel free to merge when ready!

afterdusk:
P.S. I think these writes can be dropped now that creation of root and leaf nodes will commit them to the database.

This one: https://github.com/novifinancial/akd/blob/97b2e90589083fef4600f732618715fac29a24f2/akd/src/tree_node.rs#L569

and this one: https://github.com/novifinancial/akd/blob/97b2e90589083fef4600f732618715fac29a24f2/akd/src/append_only_zks.rs#L90

PR #224, opened 2022-08-08
Updated history checks

Jasleen1:
Minor update to key history verification to ensure previous versions were retired at appropriate times. Addresses https://github.com/novifinancial/akd/issues/222


afterdusk (approved):
Perhaps we can consider adding a test for this path, but otherwise it looks good to me!

slawlor (approved):
> Perhaps we can consider adding a test for this path, but otherwise it looks good to me!

Totally agree with this, adding a small test coverage and looks good-to-go. Thanks!

Jasleen1:
Added a test for the key history check. Note that the method for testing here is using a modified version of the server's `publish` function, along with an `enum` to represent the specific types of corruptions. This method would likely generalize for testing other kinds of server misbehavior as well.

akd/src/directory.rs
```diff
@@ -872,3 +872,196 @@ pub async fn get_directory_root_hash_and_ep<
...
+
+#[cfg(test)]
+#[derive(Debug, Clone)]
+pub(crate) enum PublishCorruption {
+    /// Indicates to the malicious publish function to not mark a stale version
+    UnmarkedStaleVersion(AkdLabel),
+    /// Indicates to the malicious publish to mark a certain version for a username as stale.
+    MarkVersionStale(AkdLabel, u64),
+    /// No corruption
+    NoCorruption,
```
kevinlewi, line 886: @Jasleen1 : Seems like NoCorruption isn't used anywhere. Can we delete this to resolve the warning?

Also, can we move this code into tests.rs?

kevinlewi: Discussed offline with Jasleen, it should be ok to keep in directory.rs. Let's land this!

kevinlewi:
LGTM, thanks!

kevinlewi (approved):

PR #215, opened 2022-07-07
Ozks refactor

Jasleen1:
Addressing the checklists in https://github.com/novifinancial/akd/issues/212

Closes #212

PR #225, opened 2022-08-09
Remove forced logger initialization in test suite

afterdusk:
This PR removes the logger initialization in the `run_test_cases_for_storage_impl` test suite. This is so that clients calling this test suite can initialize their own logger without conflicting with the AKD logger. The AKD logger can still be manually initialized by calling `akd::test_utils::init_logger` in client code.

This PR also renames remaining instances of `previous_val_*` variables to `previous_version_*`. This variable/attribute was renamed in #215 but a few instances were left behind.

### DIFF
--- a/akd/src/storage/tests.rs
+++ b/akd/src/storage/tests.rs
@@ -48,7 +48,6 @@ mod memory_storage_tests {
 /// This is public because it can be used by other implemented storage layers
 /// for consistency checks (e.g. mysql, memcached, etc)
 pub async fn run_test_cases_for_storage_impl<S: Storage + Sync + Send>(db: &S) {
-    crate::test_utils::init_logger(log::Level::Info);
     test_get_and_set_item(db).await;
     test_user_data(db).await;
     test_transactions(db).await;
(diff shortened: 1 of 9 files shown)


akd/src/storage/tests.rs
```diff
@@ -48,7 +48,6 @@ mod memory_storage_tests {
 /// This is public because it can be used by other implemented storage layers
 /// for consistency checks (e.g. mysql, memcached, etc)
 pub async fn run_test_cases_for_storage_impl<S: Storage + Sync + Send>(db: &S) {
-    crate::test_utils::init_logger(log::Level::Info);
```
slawlor, line 51: Do we need to enable the logger in the individual tests inside our storage tests if this isn't here?

slawlor (approved):
LGTM, just 1 small comment

akd/src/storage/tests.rs
afterdusk: Oh good catch! I don't have to enable it for storage tests within the AKD crate - `ctor` does that for us, but I have to enable it for the test in the `akd_mysql` crate.

Issue #227, opened 2022-08-23
Public repository support for Audit Proofs


slawlor:
In order to support both public auditing and a generated signature over root hashes by a hosted auditing pool (#125) we need to be able to host audit proofs in some public repository for consumption by any third party auditor should they see fit.

Requirements
- [x] Store roots for all epochs to allow audits between arbitrary epochs (#209)
- [x] Audit proofs support serialization in a backwards compatible format (e.g. protobuf)
- [x] A new tool needs to be created to retrieve said audit proofs from some public storage, verify them, and emit the previous & new root hashes to verify against local versions
  - [x] We are going to start with S3 as the storage medium, since it's a well known standard blob storage 
  - [x] Will build command-line app to support download and verifying the audit proofs (which will be a new crate having a dependency on the `akd` crate)
  - [x] (future) Eventually support a QR code generation say in a web page after verification which can be verified in the client application(s)

slawlor:
Items checked assuming draft PR passes

PR #228, opened 2022-08-23
Add new "public_auditing" functionality backed by S3 storage

slawlor:
First of multiple changes to support public auditors using open-source tooling to verify audit proofs. Adds rudimentary console client w/QR code generation and the protobuf spec for audit proofs


slawlor:
Related to @209 and #227

eozturk1:
PR is in draft state, will review when ready! Thank you for working on this @slawlor!

akd/src/proto/audit.rs
```diff
@@ -0,0 +1,732 @@
+// This file is generated by rust-protobuf 2.8.1. Do not edit
+// @generated
+
+// https://github.com/Manishearth/rust-clippy/issues/702
```
dillonrg, line 4: This looks like a broken link. Also, any reason we're using this fork instead of [rust-lang/rust-clippy](https://github.com/rust-lang/rust-clippy)?

akd/src/proto/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+
+// ************************ Helper Functions ************************ //
+
+const NAME_SEPARATOR: char = '/';
+
+/// The constructed blobs with naming encoding the
+/// blob name = "EPOCH/PREVIOUS_ROOT_HASH/CURRENT_ROOT_HASH"
+pub struct AuditBlob {
+    /// The name of the blob, which can be decomposed into logical components (phash, chash, epoch)
+    pub name: String,
```
dillonrg, line 162: This feels like a good spot to leverage something like the [new type idiom](https://doc.rust-lang.org/rust-by-example/generics/new_types.html) to denote what exactly composes a blob name instead of simply relying on a comment. We'd be potentially be guarded against funky ordering as well, as the comments already specify a different order for previous hash, current hash, and epoch based on ordering on line 159 vs 161.

akd/src/proto/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    /// Decompose the blob's name into the (previous_hash, current_hash, epoch) tuple
+    pub fn decompose_name(name: &str) -> Result<(Digest, Digest, u64), String> {
+        let parts = name.split(NAME_SEPARATOR).collect::<Vec<_>>();
+        if parts.len() < 3 {
+            return Err(String::from(
+                "Name is malformed, there are not enough components to reconstruct!",
+            ));
+        }
```
dillonrg, line 179: With the new type comment in mind, it seems like we'd really clean this up (or maybe even remove this method entirely) depending on the implementation. Perhaps we can do something like this to be able to get to each of the parts within the type: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5fee50b5a9a6bb71eceb57f4c6f0798a

Of course, we can always just go with a struct here if we'd prefer names as well. Just a thought to clean up the implementation a bit and make this a little less error prone.

akd/src/proto/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+            .map_err(|_| format!("Failed to parse {} into an unsigned integer", parts[0]))?;
+        let previous_epoch = crate::serialization::to_digest::<Hasher>(
+            &hex::decode(parts[1])
+                .map_err(|_| format!("Failed to parse {} as a hex string into bytes", parts[1]))?,
+        )
+        .map_err(Self::convert_akd_err)?;
+        let current_epoch = crate::serialization::to_digest::<Hasher>(
+            &hex::decode(parts[2])
+                .map_err(|_| format!("Failed to parse {} as a hex string into bytes", parts[2]))?,
+        )
```
dillonrg, line 191: Any for `crate::serialization::to_digest::<Hasher>` (see: hash_from_bytes)? Any reason we can't use that here?

akd/src/proto/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+            })?;
+        let (phash, chash, epoch) = Self::decompose_name(&self.name)?;
+        let local_proof: Result<crate::proof_structs::SingleAppendOnlyProof<Hasher>, String> =
+            proof.try_into();
+
+        Ok((phash, chash, epoch, local_proof?))
+    }
+}
+
+/// Convert an append-only proof to "Audit Blobs" which are to be stored in a publically readable storage medium
```
dillonrg, line 249: nitpick: publically -> publicy

akd/src/proto/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+    proof: crate::proof_structs::AppendOnlyProof<Hasher>,
+) -> Result<Vec<AuditBlob>, String> {
+    if proof.epochs.len() + 1 != hashes.len() {
+        return Err(format!(
+            "The proof has a different number of epochs than needed for hashes.
+            The number of hashes you provide should be one more than the number of epochs!
+            Number of epochs = {}, number of hashes = {}",
+            proof.epochs.len(),
+            hashes.len()
+        ));
```
dillonrg, line 262: What do you think about representing errors as types as opposed to raw strings?

akd_local_auditor/src/auditor/mod.rs
```diff
@@ -0,0 +1,215 @@
...
+
+const HISTORY_FILE: &str = ".akd_local_auditor_history";
+
+fn format_qr_record<H>(p_hash: H::Digest, c_hash: H::Digest, epoch: u64) -> Vec<u8>
+where
+    H: winter_crypto::Hasher,
+{
+    let p_bytes = akd::serialization::from_digest::<H>(p_hash);
+    let c_bytes = akd::serialization::from_digest::<H>(c_hash);
+    let epoch_bytes = epoch.to_ne_bytes();
```
dillonrg, line 26: What's the reason for native endianness here?

akd_local_auditor/src/auditor/mod.rs
```diff
@@ -0,0 +1,215 @@
...
+    result
+}
+
+pub async fn audit_epoch<H>(blob: akd::proto::AuditBlob, qr: bool) -> Result<()>
+where
+    H: winter_crypto::Hasher + Sync + Send,
+{
+    let (p_hash, c_hash, epoch, proof) = blob.decode().map_err(|err| anyhow!("{}", err))?;
+
+    // DO THE FREAKING VERIFICATION ALREADY
```
dillonrg, line 43: LOL is this comment still relevant? 😜

akd_local_auditor/src/auditor/mod.rs
```diff
@@ -0,0 +1,215 @@
...
+#[derive(Subcommand, Clone, Debug)]
+pub enum AuditCommand {
+    /// Audit a specific epoch
+    Audit {
+        /// The epoch to audit
+        #[clap(long)]
+        epoch: u64,
+        /// Show a QR code of the audit
+        #[clap(long)]
+        qr: bool,
```
dillonrg, line 81: This is really cool! Only one tiny suggestion: should we make this indicative of the flag that it's a flag, e.g., `disp_qr` or something? Dont' want to really make it verbose, but we can always specify a short-flag for it if that's a concern.

akd_local_auditor/src/auditor/mod.rs
```diff
@@ -0,0 +1,215 @@
...
+    info!(
+        "Audit history is available between epochs ({}) and ({}), inclusively.",
+        min.epoch, max.epoch
+    );
+
+    Ok(())
+}
+
+async fn process_command(
+    storage: &Box<dyn crate::storage::AuditProofStorage>,
```
dillonrg, line 126: As the linter suggests, is the `Box` really necessary here?

akd_local_auditor/src/console_log.rs
```diff
@@ -0,0 +1,93 @@
...
+pub struct ConsoleLogger {
+    pub level: Level,
+}
+
+impl ConsoleLogger {
+    pub fn touch() {
+        EPOCH.get_or_init(Instant::now);
+    }
+
+    pub(crate) fn format_log_record(io: &mut (dyn Write + Send), record: &Record, no_color: bool) {
```
dillonrg, line 28: I have two suggestions with the `no_color` arguments: 

1. Can we utilize a positive naming convention to avoid double negatives to indicate a positive? That is, prefer something like `colored` as opposed to `no_color` to avoid NOT `no_color`?
2. In general, move away from a raw bool here. Generally, I think that [arguments convey meaning through types](https://rust-lang.github.io/api-guidelines/type-safety.html#c-custom-type). As such, we can do something like `struct Colored(bool)` at no cost and better convey the flags intent.

akd_local_auditor/src/console_log.rs
```diff
@@ -0,0 +1,93 @@
...
+            Instant::now() - *epoch
+        } else {
+            Duration::from_millis(0)
+        };
+
+        let seconds = toc.as_secs();
+        let hours = seconds / 3600;
+        let minutes = (seconds / 60) % 60;
+        let seconds = seconds % 60;
+        let miliseconds = toc.subsec_millis();
```
dillonrg, line 51: `let miliseconds` => `let milliseconds`

akd_local_auditor/src/main.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub(crate) fn to_log_level(&self) -> log::Level {
+        match &self {
+            PublicLogLevel::Error => log::Level::Error,
+            PublicLogLevel::Warn => log::Level::Warn,
+            PublicLogLevel::Info => log::Level::Info,
+            PublicLogLevel::Debug => log::Level::Debug,
+            PublicLogLevel::Trace => log::Level::Trace,
+        }
+    }
+}
```
dillonrg, line 45: Suggestion: This is a good fit for the `From` trait.

akd_local_auditor/src/storage/dynamodb.rs
```diff
@@ -0,0 +1,18 @@
...
+//! Audit Proofs
+
+use clap::Args;
+
+#[derive(Args, Debug, Clone)]
+pub struct DynamoDbClapSettings {
+    /// The S3 bucket where the audit proofs are stored
+    #[clap(long)]
+    test: String,
+}
```
dillonrg, line 18: Likely just a side-effect of this still being a WIP, but figured I'd point out that this looks like it has some copy-paste remnants form the world of S3 :)

akd_local_auditor/src/storage/s3.rs
```diff
@@ -0,0 +1,180 @@
...
+use tokio::sync::RwLock;
+
+const MIN_BUCKET_CHARS: usize = 3;
+const MAX_BUCKET_CHARS: usize = 63;
+const ALLOWED_BUCKET_CHARS: [char; 38] = [
+    'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o', 'p', 'q', 'r', 's',
+    't', 'u', 'v', 'w', 'x', 'y', 'z', '.', '-', '1', '2', '3', '4', '5', '6', '7', '8', '9', '0',
+];
+
+fn is_bucket_name_valid(s: &str) -> Result<String, String> {
```
dillonrg, line 28: I assume that we're aiming to implement a portion of the [rules for S3 bucket names](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html) as validation here; is that correct? If so, do you think that's worth trying to keep up? I ask, because a user of the tool should/will eventually see an error from AWS in any case when we attempt to read from the bucket. Additionally, if bucket naming conventions change, then there's potential that the validation here becomes stale. I'm less concerned about the latter, as I don't see bucket names being named oddly often.

akd_local_auditor/src/storage/s3.rs
```diff
@@ -0,0 +1,180 @@
...
+use std::convert::TryInto;
+use std::sync::Arc;
+use tokio::sync::RwLock;
+
+const MIN_BUCKET_CHARS: usize = 3;
+const MAX_BUCKET_CHARS: usize = 63;
+const ALLOWED_BUCKET_CHARS: [char; 38] = [
+    'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o', 'p', 'q', 'r', 's',
+    't', 'u', 'v', 'w', 'x', 'y', 'z', '.', '-', '1', '2', '3', '4', '5', '6', '7', '8', '9', '0',
+];
```
dillonrg, line 26: If you'd like to avoid typing all of this out, you can do something like this:
```rust
    let chars: Vec<char> = ('a'..='z')
        .into_iter()
        .chain(vec!['-', '.'].into_iter())
        .chain(('0'..='9').into_iter())
        .collect();
```

However, I don't believe the strategy above can be const. It does [generate the same set of chars we've manually added here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7835e98b72700257e082760a28ecb50d), so a bit more convenient. Your call!

akd_local_auditor/src/storage/s3.rs
```diff
@@ -0,0 +1,180 @@
...
+        }
+        Ok(str)
+    } else {
+        Err(format!(
+            "Bucket name must be between [{}, {}] characters in length. Gave {}",
+            MIN_BUCKET_CHARS,
+            MAX_BUCKET_CHARS,
+            str.len()
+        ))
+    }
```
dillonrg, line 44: Just a suggestion, but you can invert the if and return the `Err` to avoid a level a nesting with the inner for-loop. Additionally, I think the inner check for validating all characters by doing something like [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=be8299413bfb44c6bb22bcda7ba02f85).

akd_local_auditor/src/storage/s3.rs
```diff
@@ -0,0 +1,180 @@
...
+                                    Err(error) => {
+                                        debug!("Error parsing blob key into DecomposedBlobItem ({}). Skipping.", error);
+                                    }
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        }
```
dillonrg, line 144: Nit: This is a wee-bit intense. Is there a way we can break this up into something a bit more readable?

akd/src/proto/audit.proto
```diff
@@ -0,0 +1,37 @@
...
+// Command to re-generate (from 'proto' directory):
+// protoc --rust_out . audit.proto
+
+syntax = "proto2";
+
+/* NodeLabel represents the label of a history tree node in the AKD tree with a
+ * supplied label-length and label value (location) */
+message NodeLabel {
+    optional bytes label_val = 1;
+    optional uint32 label_len = 2;
```
dillonrg, line 24: So I know that `optional` is the recommended way of adding fields to messages for compatibility reasons, but do we really foresee these ever being removed? That is, is it possible that we can mark these as required? I ask because we can potentially eliminate the macro we wrote for ensuring that fields are present and instead rely on proto to handle that for us.

dillonrg (changes requested):
Thanks for the change, Sean! Overall, the functionality looks like it's there. I've mostly left some comments around suggestions that may leverage types a bit more in order to eliminate some existing code and potentially prevent bugs in the future. Seeing as how this is still a bit of a work in-progress, I am requesting changes. That said, I don't feel that anything I've seen here is a hard-blocker at this point. Considering the size of the PR, I'll do one more read-through with a pair of fresh eyes to ensure I didn't miss something.

akd/src/proto/audit.proto
slawlor: So from the protobuf guide:

> Required Is Forever You should be very careful about marking fields as required. If at some point you wish to stop writing or sending a required field, it will be problematic to change the field to an optional field – old readers will consider messages without this field to be incomplete and may reject or drop them unintentionally. You should consider writing application-specific custom validation routines for your buffers instead.
> 
> A second issue with required fields appears when someone adds a value to an enum. In this case, the unrecognized enum value is treated as if it were missing, which also causes the required value check to fail.

At this point, the macros are there, they work as expected, I think it's better to leave them if we do ever break something in the future. I'd much prefer as *much* backwards compatibility and flexibility as possible

akd_local_auditor/src/auditor/mod.rs
slawlor: ah total typo, that should be le bytes. my bad!

akd_local_auditor/src/auditor/mod.rs
slawlor: Once again the nightly clippy strikes! Local clippy shows nothing :p

akd_local_auditor/src/storage/s3.rs
slawlor: It's more that we won't even try and make a request unless it matches the _basics_ of naming conventions which is 3-63 chars, lower-case a-z, 0-9, and/or ., -

That's all it is and it hasn't changed in a long time. I'm not saying we should go full-spec match across the board, but this is an easy check to see if they mistyped the name of the bucket. They're URL names afterall so there's already a tight limit on what's possible

akd_local_auditor/src/storage/dynamodb.rs
slawlor: Yeah it's more as a placeholder should it come, and showing how you "might" implement other storage traits. I'll note it

akd_local_auditor/src/storage/s3.rs
slawlor: Ah a remnant of my functional programming days. early exit's are "frowned upon" often, as it makes tracing exits a lot more complicated sometimes. But we aren't writing functional code here lol

akd_local_auditor/src/auditor/mod.rs
dillonrg: I'm good with `qr`!

akd_local_auditor/src/storage/s3.rs
dillonrg: Haha no worries! If you prefer this style, that's fine with me. This was mostly a nitpick type comment that I didn't explicitly categorize as such!

akd/src/proto/audit.proto
dillonrg: Works for me! Figured I'd ask to spark the conversation! I think you make good points and am good with leaving as much backwards compatibility/flexibility as possible at this point.

slawlor:
@dillonrg good with you?

.github/workflows/ci.yml
```diff
@@ -34,6 +34,8 @@ jobs:
           echo "MySQL container is up"
       - name: Check container akd-test-db logs
         run: docker logs akd-test-db
+      - name: Setup protoc
```
dillonrg, line 37: Infinitely small nitpick: "Setup" -> "Set up" since we're using this as a verb.

.github/workflows/ci.yml
```diff
@@ -75,7 +82,8 @@ jobs:
           toolchain: stable
           components: clippy
           override: true
-
+      - name: Setup protoc
```
dillonrg, line 85: Same comment here about "set up"

akd_local_auditor/src/auditor/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+                    )
+                }
+            });
+
+    if !is_contiguous {
+        bail!("The audit proofs appear to not be continguous. There's a break in the linear history at epoch {}", maybe_broken_epoch.name.epoch);
+    }
+
+    info!(
+        "Audit history is available between epochs ({}) and ({}), inclusively.",
```
dillonrg, line 123: Thinking out loud: think it makes sense to display this with interval notation?

akd_local_auditor/src/main.rs
```diff
@@ -0,0 +1,99 @@
...
+//     pub(crate) fn to_log_level(&self) -> log::Level {
+//         match &self {
+//             PublicLogLevel::Error => log::Level::Error,
+//             PublicLogLevel::Warn => log::Level::Warn,
+//             PublicLogLevel::Info => log::Level::Info,
+//             PublicLogLevel::Debug => log::Level::Debug,
+//             PublicLogLevel::Trace => log::Level::Trace,
+//         }
+//     }
+// }
```
dillonrg, line 56: Can we delete this?

akd_local_auditor/src/main.rs
```diff
@@ -0,0 +1,99 @@
...
+    console_log::ConsoleLogger::touch();
+
+    let args = Arguments::parse();
+
+    // initialize the logger
+    let log_level: log::Level = (&args.log_level).into();
+
+    log::set_logger(&LOGGER)
+        .map(|()| log::set_max_level(log_level.to_level_filter()))
+        .expect("Failed to setup logging");
```
dillonrg, line 87: I'm the "set up" police 🤣

akd_local_auditor/src/storage/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+    /// Amazon S3 compatible storage
+    S3(s3::S3ClapSettings),
+    // /// DynamoDB
+    // DynamoDb(dynamodb::DynamoDbClapSettings),
+}
+
+// ************************************ Trait and Type Definitions ************************************ //
+
+/// Represents the summary of an epoch, and a unique key referring to the raw object in native storage (if needed)
+#[derive(Clone)]
```
dillonrg, line 33: Might as well [derive other common traits](https://rust-lang.github.io/api-guidelines/interoperability.html#c-common-traits) that might come in handy.

dillonrg (approved):

akd_local_auditor/src/auditor/mod.rs
slawlor: This is just a starting point I think. At some point I'm going to give a menu with proper formatting, but still TODO once the AWS account is up and running lol

PR #231, opened 2022-08-29
Fix the in-memory database implementations to properly handle transactions

slawlor:
The `batch_set` implementations in both memory databases wasn't updating the transaction logs, causing improper construction of the transaction log. This makes both `batch_set` implementations adhere to the spec for transaction handling, as well as adding some test coverage around the ordering of the aZKS commitment to storage (aZKS record must be **last**).

### DIFF
--- a/akd/src/storage/memory.rs
+++ b/akd/src/storage/memory.rs
@@ -87,9 +87,10 @@ impl Storage for AsyncInMemoryDatabase {
     async fn commit_transaction(&self) -> Result<(), StorageError> {
         // this retrieves all the trans operations, and "de-activates" the transaction flag
         let ops = self.trans.commit_transaction().await?;
+
         let _epoch = match ops.last() {
             Some(DbRecord::Azks(azks)) => Ok(azks.latest_epoch),
-            other => Err(StorageError::Other(format!(
+            other => Err(StorageError::Transaction(format!(
                 "The last record in the transaction log is NOT an Azks record {:?}",
                 other
             ))),
@@ -534,7 +535,7 @@ impl Storage for AsyncInMemoryDbWithCache {
         let ops = self.trans.commit_transaction().await?;
         let _epoch = match ops.last() {
             Some(DbRecord::Azks(azks)) => Ok(azks.latest_epoch),
-            other => Err(StorageError::Other(format!(
+            other => Err(StorageError::Transaction(format!(
                 "The last record in the transaction log is NOT an Azks record {:?}",
                 other
             ))),
--- a/akd/src/storage/memory.rs
+++ b/akd/src/storage/memory.rs
@@ -87,6 +87,13 @@ impl Storage for AsyncInMemoryDatabase {
     async fn commit_transaction(&self) -> Result<(), StorageError> {
         // this retrieves all the trans operations, and "de-activates" the transaction flag
         let ops = self.trans.commit_transaction().await?;
+        let _epoch = match ops.last() {
+            Some(DbRecord::Azks(azks)) => Ok(azks.latest_epoch),
+            other => Err(StorageError::Other(format!(
+                "The last record in the transaction log is NOT an Azks record {:?}",
+                other
+            ))),
+        }?;
         self.batch_set(ops).await
     }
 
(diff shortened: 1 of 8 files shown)


akd/src/storage/memory.rs
```diff
@@ -87,6 +87,13 @@ impl Storage for AsyncInMemoryDatabase {
     async fn commit_transaction(&self) -> Result<(), StorageError> {
         // this retrieves all the trans operations, and "de-activates" the transaction flag
         let ops = self.trans.commit_transaction().await?;
+        let _epoch = match ops.last() {
+            Some(DbRecord::Azks(azks)) => Ok(azks.latest_epoch),
+            other => Err(StorageError::Other(format!(
+                "The last record in the transaction log is NOT an Azks record {:?}",
+                other
```
dillonrg, line 94: What do you think about adding a specific error enum to represent what's going on here? Alternatively, should we go with `StorageError::Transaction` here?

akd/src/storage/transaction.rs
```diff
@@ -40,6 +40,12 @@ impl std::fmt::Debug for Transaction {
     }
 }
 
+impl Default for Transaction {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
```
dillonrg, line 48: It looks like we can add `#[derive(Default)]` to `TransactionState` and `Transaction` if we'd like to avoid the the explicit impl here.

akd_mysql/src/mysql.rs
```diff
@@ -740,6 +740,15 @@ impl Storage for AsyncMySqlDatabase {
...
         // this retrieves all the trans operations, and "de-activates" the transaction flag
         let ops = self.trans.commit_transaction().await?;
+
+        let _epoch = match ops.last() {
+            Some(DbRecord::Azks(azks)) => Ok(azks.latest_epoch),
+            other => Err(StorageError::Other(format!(
+                "The last record in the transaction log is NOT an Azks record {:?}",
+                other
+            ))),
+        }?;
```
dillonrg, line 750: If this is solely to handle the error case, I think we can make this a bit more explicit with `if let` and avoiding the whole match + handling of `Some`. I think this same comment applies to any block where we're ignoring the result of the match.

akd/src/storage/memory.rs
```diff
@@ -87,6 +87,13 @@ impl Storage for AsyncInMemoryDatabase {
     async fn commit_transaction(&self) -> Result<(), StorageError> {
         // this retrieves all the trans operations, and "de-activates" the transaction flag
         let ops = self.trans.commit_transaction().await?;
+        let _epoch = match ops.last() {
+            Some(DbRecord::Azks(azks)) => Ok(azks.latest_epoch),
```
dillonrg, line 91: For my own understanding, can you expand on why checking the last item in the `Vec` is appropriate in these checks for determining if the kind of `DbRecord` is `Azks`?

akd/src/storage/memory.rs
```diff
@@ -558,6 +584,18 @@ impl Storage for AsyncInMemoryDbWithCache {
...
+            // nothing to do, save the cycles
+            return Ok(());
+        }
+
+        if self.is_transaction_active().await {
+            for record in records {
+                self.trans.set(&record).await;
+            }
+            return Ok(());
+        }
```
dillonrg, line 597: It looks like there is a decent amount of duplication across the impl blocks for the traits. Maybe we can come back to this eventually with some work focused on clean up. Not a blocker for this diff by any means!

dillonrg (changes requested):
Overall, this looks good! I'm requesting changes solely to get your take on potentially adding a new type of error for what we're using the `StorageError::Other` kind for currently. Other than that, most of my comments are stylistic type comments and non-blocking. If you'd prefer to keep this as is, please let me know and I'll approve.

akd/src/storage/memory.rs
slawlor: We need to make sure that the last thing committed is the Azks record as that is what triggers concurrent reading entities that a new epoch has occurred. The transaction is applying a sorting when it's generating the transaction set, and this just asserts that the sort was correct and the last element is the right one to trigger the notification to everything that there's something new.

akd_mysql/src/mysql.rs
slawlor: If we use an `if let` statement, we can only bind the result 1 time. There's no `if let Some(DbRecord::Azks(azks)) ..` and then still able to handle the result without using a `match` statement unfortunately.

dillonrg: Yes, you're right! I misinterpreted this originally and now better realized the intent here. Thanks for clarifying!

dillonrg (approved):

PR #233, opened 2022-08-30
Migrate to usage of rustyrepl for easy REPL processing + S3 integration tests

slawlor:
`rustyrepl` is a new crate to facilitate a lot of the boiler plate with REPL processing. It provides a cleaner syntax and is easy to manage.

Additionally this PR adds integration tests around the `s3` integration logic utilizing an s3-minic (`minio`) docker container.

### DIFF
--- a/.github/workflows/s3.yml
+++ b/.github/workflows/s3.yml
@@ -1,0 +1,58 @@
+name: CI
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'akd_local_auditor/'
+  pull_request:
+    types: [opened, repoened, synchronize]
+    paths:
+      - 'akd_local_auditor/'
+
+jobs:
+  test-auditor:
+    name: Test Rust ${{matrix.toolchain}} on ${{matrix.os}}
+    runs-on: ${{matrix.os}}-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        toolchain: [stable]
+        os: [ubuntu]
+    steps:
+      - uses: actions/checkout@main
+
+      - name: Install rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{matrix.toolchain}}
+          override: true
+
+      - name: Cargo build
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
+
+      # Borrowed from: https://github.com/engula/engula/pull/124/files
+      - name: Setup minio
+        run: |
+          docker run --detach \
+            -p 9000:9000 \
+            -p 9001:9001 \
+            -v "$(mktemp -d)":/data \
+            -e "MINIO_ROOT_USER=test" \
+            -e "MINIO_ROOT_PASSWORD=someLongAccessKey" \
+            -e "MINIO_REGION_NAME=us-east-2" \
+            quay.io/minio/minio server /data --console-address ":9001"
+      - name: Liveness probe
+        timeout-minutes: 1
+        run: |
+          until curl http://127.0.0.1:9000/minio/health/live; do
+              sleep 1
+          done
+
+      - name: Cargo test
+        uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --manifest-path Cargo.toml -p akd_local_auditor integration_test_ -- --ignored
--- a/.github/workflows/s3.yml
+++ b/.github/workflows/s3.yml
@@ -34,7 +34,7 @@ jobs:
           command: build
 
       # Borrowed from: https://github.com/engula/engula/pull/124/files
-      - name: Setup minio
+      - name: Set up minio
         run: |
           docker run --detach \
             -p 9000:9000 \
--- a/akd_local_auditor/src/auditor/mod.rs
+++ b/akd_local_auditor/src/auditor/mod.rs
@@ -9,13 +9,12 @@
 
 use anyhow::{anyhow, bail, Result};
 use clap::{Parser, Subcommand};
-use log::{debug, error, info, warn};
+use log::{error, info, warn};
 use std::marker::{Send, Sync};
 
-use rustyline::error::ReadlineError;
-use rustyline::Editor;
+use rustyrepl::ReplCommandProcessor;
 
-const HISTORY_FILE: &str = ".akd_local_auditor_history";
+pub(crate) const HISTORY_FILE: &str = ".akd_local_auditor_history";
 
 fn format_qr_record<H>(p_hash: H::Digest, c_hash: H::Digest, epoch: u64) -> Vec<u8>
 where
@@ -93,6 +92,33 @@ pub struct AuditArgs {
     command: AuditCommand,
 }
 
+/// Audit processing crate
+#[derive(Debug)]
+pub struct AuditProcessor {
+    pub(crate) storage: Box<dyn crate::storage::AuditProofStorage>,
+}
+
+#[async_trait::async_trait]
+impl ReplCommandProcessor<AuditArgs> for AuditProcessor {
+    fn is_quit(&self, command: &str) -> bool {
+        matches!(command, "quit" | "exit")
+    }
+
+    async fn process_command(&self, cmd: AuditArgs) -> Result<()> {
+        match &cmd.command {
+            AuditCommand::Audit { epoch, qr } => {
+                let proof = self.storage.get_proof(*epoch).await?;
+                audit_epoch::<crate::Hasher>(proof, *qr).await?;
+            }
+            AuditCommand::ShowEpochs => {
+                let mut proofs = self.storage.list_proofs().await?;
+                display_audit_proofs_info(&mut proofs)?;
+            }
+        }
+        Ok(())
+    }
+}
+
 fn display_audit_proofs_info(info: &mut Vec<crate::storage::EpochSummary>) -> Result<()> {
     info.sort_by(|a, b| a.name.epoch.cmp(&b.name.epoch));
     if info.is_empty() {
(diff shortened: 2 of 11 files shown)


slawlor:
Will only be able to validate the S3 pipeline once landed + merged. However `docker compose` tests worked just fine :)

slawlor:
Related issue: #227

.github/workflows/s3.yml
```diff
@@ -0,0 +1,58 @@
...
+          toolchain: ${{matrix.toolchain}}
+          override: true
+
+      - name: Cargo build
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
+
+      # Borrowed from: https://github.com/engula/engula/pull/124/files
+      - name: Setup minio
```
dillonrg, line 37: Tiny nitpick: Setup => set up since it's being used as a verb.

akd_local_auditor/src/storage/s3/mod.rs
```diff
@@ -45,6 +51,11 @@ fn is_bucket_name_valid(s: &str) -> Result<String, String> {
     Ok(str)
 }
 
+fn is_uri_valid(s: &str) -> Result<String, String> {
+    let _uri: http::Uri = s.parse::<http::Uri>().map_err(|err| err.to_string())?;
+    Ok(s.to_string())
+}
```
dillonrg, line 57: This feels like sort of an odd name given we're not returning a `bool`, but instead validating a URI and returning a Result. Perhaps something more reflective of what's happening here would be `validate_uri` or `parse_uri`.

akd_local_auditor/src/storage/s3/mod.rs
```diff
@@ -53,11 +64,36 @@ pub struct S3ClapSettings {
...
     )]
     bucket: String,
+
+    /// The AWS region to operate in
+    #[clap(long)]
+    region: String,
+
+    /// [OPTIONAL] An custom URI for the AWS endpoint
+    #[clap(long, value_parser = is_uri_valid)]
+    endpoint: Option<String>,
```
dillonrg, line 74: Any reason not to make this an `Option<URI>` to make a slightly stronger type than just `String`?

akd_local_auditor/src/storage/s3/mod.rs
```diff
@@ -69,22 +105,47 @@ impl From<&S3ClapSettings> for S3AuditStorage {
...
+            let mut shared_config_loader = aws_config::from_env().region(region_provider);
+
+            if let (Some(access_key), Some(secret_key)) = (&self.access_key, &self.secret_key) {
+                let credentials = aws_types::Credentials::from_keys(access_key, secret_key, None);
+                shared_config_loader = shared_config_loader.credentials_provider(credentials);
+            }
+
+            if let Some(endpoint) = &self.endpoint {
+                // THE URI IS ALREADY VALIDATED BY CLAP ARGS, hence the expect here
+                let endpoint_resolver = Endpoint::immutable(endpoint.parse().expect("valid URI"));
```
dillonrg, line 136: Likely a good expect. It looks like URI might allow relative URIs, which I don't _think_ you want here based on usage. See: https://docs.rs/http/latest/http/uri/struct.Uri.html#examples-5.

akd_local_auditor/src/storage/s3/test.rs
```diff
@@ -0,0 +1,368 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
```
dillonrg, line 1: Please use updated headers! See: https://github.com/novifinancial/akd/pull/237

akd_local_auditor/src/storage/s3/test.rs
```diff
@@ -0,0 +1,368 @@
...
+struct AuditInformation {
+    proof: akd::proof_structs::AppendOnlyProof<Hasher>,
+    phash: Digest,
+    chash: Digest,
+}
+
+/// Generate N audit proofs from a new tree
+async fn generate_audit_proofs(n: usize) -> Result<Vec<AuditInformation>, akd::errors::AkdError> {
+    let db = AsyncInMemoryDatabase::new();
+    let vrf = HardCodedAkdVRF {};
```
dillonrg, line 68: We should probably add an associated new fn to this type: https://rust-lang.github.io/api-guidelines/predictability.html#c-ctor.

akd_local_auditor/src/storage/s3/test.rs
```diff
@@ -0,0 +1,368 @@
...
+        akd.publish::<Hasher>(vec![(
+            AkdLabel::from_utf8_str("user"),
+            AkdValue::from_utf8_str(&t_value),
+        )])
+        .await?;
+        // generate the append-only proof
+        let proof = akd
+            .audit::<Hasher>((_epoch - 1) as u64, _epoch as u64)
+            .await?;
+        // update the azks + retrieve the "next" hash
```
dillonrg, line 87: Isn't this the "current" hash? 😅

akd_local_auditor/src/storage/s3/test.rs
```diff
@@ -0,0 +1,368 @@
...
+        phash = chash;
+    }
+    Ok(proofs)
+}
+
+/// Flush the storage bucket of all blobs + delete the bucket (if it exists)
+async fn maybe_flush_storage(shared_config: &SdkConfig, bucket_name: &str) -> Result<()> {
+    // Build the S3 config from the shared SdkConfig
+    let config = S3ConfigBuilder::from(shared_config)
+        .retry_config(RetryConfig::disabled())
```
dillonrg, line 109: I assume we're not using a retry since this is a test; is that right?

akd_local_auditor/src/storage/s3/test.rs
```diff
@@ -0,0 +1,368 @@
...
+
+                    // From: https://docs.aws.amazon.com/sdk-for-rust/latest/dg/rust_s3_code_examples.html
+                    let objects = client.list_objects_v2().bucket(bucket_name).send().await?;
+                    let mut delete_objects: Vec<ObjectIdentifier> = vec![];
+                    for obj in objects.contents().unwrap_or_default() {
+                        let obj_id = ObjectIdentifier::builder()
+                            .set_key(Some(obj.key().unwrap().to_string()))
+                            .build();
+                        delete_objects.push(obj_id);
+                    }
```
dillonrg, line 132: Can you do something like below to make this slightly more succinct? 
```rust
let delete_objects = objects.contents().unwrap_or_default().iter()
    .map(...builder_stuff)
    .collect();
```

akd_local_auditor/src/storage/s3/test.rs
```diff
@@ -0,0 +1,368 @@
...
+    let client = Client::from_conf(config);
+    // check if the bucket exists
+    let buckets = client.list_buckets().send().await?;
+
+    if let Some(bucket_names) = buckets.buckets() {
+        log::info!("Found {} buckets in the test storage", bucket_names.len());
+
+        for bucket in bucket_names {
+            if let Some(maybe_name) = bucket.name() {
+                if maybe_name == bucket_name {
```
dillonrg, line 121: Seeing as how we're iterating over buckets to find the first one here, I think you can use [`find`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find) on the iterator to get an option of the first result. From there, you can map it to do what you're doing in this loop. I _think_ that should be slightly cleaner, but ultimately you're call!

akd_local_auditor/src/storage/s3/test.rs
```diff
@@ -0,0 +1,368 @@
...
+async fn populate_test_storage(
+    shared_config: &SdkConfig,
+    bucket: &str,
+    n_blobs: usize,
+) -> Result<()> {
+    maybe_flush_storage(shared_config, bucket).await?;
+
+    // Build the S3 config from the shared SdkConfig
+    let config = S3ConfigBuilder::from(shared_config)
+        .retry_config(RetryConfig::disabled())
```
dillonrg, line 165: Ditto for past comment about no retries.

akd_local_auditor/src/storage/s3/test.rs
```diff
@@ -0,0 +1,368 @@
...
+
+fn get_s3_storage(bucket: &str) -> S3AuditStorage {
+    let clap_args = S3ClapSettings {
+        bucket: bucket.to_string(),
+        region: TEST_REGION.to_string(),
+        endpoint: Some(TEST_ENDPOINT.to_string()),
+        access_key: Some(ACCESS_KEY.to_string()),
+        secret_key: Some(SECRET_KEY.to_string()),
+    };
+    let storage: S3AuditStorage = (&clap_args).into();
```
dillonrg, line 219: Can't you just return `(&clap_args).into()` here? I _think_ Rust should be smart enough to infer the type, but maybe I'm wrong.

akd_local_auditor/src/storage/s3/test.rs
```diff
@@ -0,0 +1,368 @@
...
+    );
+    assert_eq!(10, epoch_summaries.len());
+
+    // check the linear history of the proofs
+    log::info!("Checking linear history of audit proofs");
+    let mut i = 0u64;
+    for summary in epoch_summaries {
+        assert_eq!(i, summary.name.epoch);
+        i += 1;
+    }
```
dillonrg, line 257: I believe you can use [`enumerate`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.enumerate) here to avoid having to maintain your own `i` variable.

akd_local_auditor/src/storage/s3/test.rs
```diff
@@ -0,0 +1,368 @@
...
+    log::info!(
+        "There are {} epochs in the storage layer",
+        epoch_summaries.len()
+    );
+    assert_eq!(3, epoch_summaries.len());
+
+    // verify all fo the audit proofs
+    for epoch in 0..=2 {
+        let proof_blob = storage.get_proof(epoch).await?;
+        log::info!("Verification epoch {} -> {}", epoch, epoch + 1);
```
dillonrg, line 294: Is it misleading that this will print 0 -> 1, 1 -> 2, 2 -> 3 since we technically only have proofs 0,1,2?

akd_local_auditor/src/storage/s3/test.rs
```diff
@@ -0,0 +1,368 @@
...
+        crate::auditor::audit_epoch::<Hasher>(proof_blob.clone(), false).await?;
+        crate::auditor::audit_epoch::<Hasher>(proof_blob, true).await?;
```
dillonrg, line 296: Insanely tiny nitpick: can we assign the bools to variables indicating what they signify or something? It's hard to tell as a reader without seeking out the `audit_epoch` fn separately. Feel free to ignore if you don't find much value in that. Generally, not a major fan of bool arguments (for several reasons) so I may be a bit biased!

akd_local_auditor/src/storage/s3/test.rs
```diff
@@ -0,0 +1,368 @@
...
+#[test]
+fn test_uri_parsing() -> Result<()> {
+    let non_uri = "hasdf!@#";
+    let bad_port = "http://localhost:1234%567891230";
+
+    assert!(matches!(super::is_uri_valid(non_uri), Err(_)));
+    assert!(matches!(super::is_uri_valid(bad_port), Err(_)));
+
+    assert!(matches!(super::is_uri_valid("www.ok.com"), Ok(_)));
+    assert!(matches!(super::is_uri_valid("http://ok.com"), Ok(_)));
```
dillonrg, line 357: Maybe a good spot to verify my question about partial URIs 😬

dillonrg (approved):
Looks good! The minio stuff is really cool! 

I don't believe any of my comments are hard blockers, so please feel free to carry my approval.

akd_local_auditor/src/auditor/mod.rs
```diff
@@ -119,102 +145,10 @@ fn display_audit_proofs_info(info: &mut Vec<crate::storage::EpochSummary>) -> Re
         bail!("The audit proofs appear to not be continguous. There's a break in the linear history at epoch {}", maybe_broken_epoch.name.epoch);
     }
 
-    info!(
+    println!(
```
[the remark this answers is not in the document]
slawlor: Ah logs can be redirected, whereas println is for response in the REPL. If we use log statements, it adds the log-level and all the other metadata to it. That often say could be redirected to a file, whereas response from the REPLs should go out on the command prompt.

akd_local_auditor/src/storage/s3/mod.rs
slawlor: Yes, clap requires value parsers to return `Result<String, String>` not a bool type. But because of that we can definitely use a better name

akd_local_auditor/src/storage/s3/mod.rs
slawlor: Again a Clap thing. Uri isn't implementing Debug, parser, etc which are all necessary traits to parse the clap args properly

akd_local_auditor/src/storage/s3/mod.rs
slawlor: Good point, I'm updating the parser to reflect relative uris not being supported

akd_local_auditor/src/storage/s3/test.rs
slawlor: yes, exactly. And with local network it shouldn't be a problem for testing scenarios

akd_local_auditor/src/storage/mod.rs
```diff
@@ -54,7 +54,7 @@ impl TryFrom<&str> for EpochSummary {
 
 /// Represents a storage of audit proofs and READ ONLY interaction to retrieve the proof objects
 #[async_trait]
-pub trait AuditProofStorage: Sync + Send {
+pub trait AuditProofStorage: Sync + Send + std::fmt::Debug {
```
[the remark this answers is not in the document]
dillonrg: Ah, got it. So you're basically looking to disambiguate between `core::fmt` and `std::fmt`? Thanks!

akd_local_auditor/src/auditor/mod.rs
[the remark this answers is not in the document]
dillonrg: Ah, okay, figured it something to do with that! Thanks for the additional context.

akd_local_auditor/src/storage/s3/mod.rs
dillonrg: Makes sense! Thanks for clarifying :)

akd_local_auditor/src/storage/s3/test.rs
slawlor: that was originally copy-pasta from the AWS docs, but this is cleaner for sure

akd_local_auditor/src/storage/s3/test.rs
slawlor: I don't think so, since an audit proof is the proof of the changes from epoch X -> X + 1. So we're validating three proofs, but cover epoch changes
0 to 1
1 to 2
2 to 3

etc

dillonrg: Makes sense! I think I overthought this last night 😅

akd_local_auditor/src/storage/s3/test.rs
```diff
@@ -116,37 +116,46 @@ async fn maybe_flush_storage(shared_config: &SdkConfig, bucket_name: &str) -> Re
...
-                    }
-                    client.delete_bucket().bucket(bucket_name).send().await?;
-                    break;
+        if bucket_names
+            .iter()
+            .find(|bucket| {
+                if let Some(maybe_bucket) = bucket.name() {
+                    return maybe_bucket == bucket_name;
                 }
+                false
```
dillonrg, line 125: If you want a one-liner with [`map_or_else`](https://doc.rust-lang.org/std/option/enum.Option.html#method.map_or_else): 

```rust
bucket.name().map_or_else(|| false, |n| n == bucket_name)
```

dillonrg (approved):
Looks great!

akd_local_auditor/src/storage/s3/test.rs
slawlor: ah missed this! Oh well :p

dillonrg: Yeah, no biggie! 🤷

PR #234, opened 2022-08-30
Add serialized proof verification to akd_client

eozturk1:
Adding the API and tests for verifying serialized proofs. JSON is used as the medium to serialize with since it offers easier debugging and more interoperability (maybe).

Since an AKD user will need to convert the AKD types to AKD_CLIENT types first before serializing proofs, this PR also adds support for the `converters` feature.

Testing:

`cargo build/test` passes as well as `cargo build/test --features=converters`

### DIFF
--- a/akd_client/Cargo.toml
+++ b/akd_client/Cargo.toml
@@ -21,7 +21,7 @@ winter-utils = { version = "0.2", optional = true }
 sha2 = { version = "0.10.1", optional = true, default-features = false }
 sha3 = { version = "0.10.0", optional = true, default-features = false }
 blake3 = { version = "1.3.1", optional = true, default-features = false }
-serde_json = "1.0"
+serde_json = "1"
 
 wasm-bindgen = { version = "0.2.79", optional = true, features = ["serde-serialize"] }
 serde = { version = "1.0", features = ["derive"]}
--- a/akd_client/Cargo.toml
+++ b/akd_client/Cargo.toml
@@ -16,9 +16,10 @@ crate-type = ["cdylib", "rlib"]
 sha2 = { version = "0.10.1", optional = true, default-features = false }
 sha3 = { version = "0.10.0", optional = true, default-features = false }
 blake3 = { version = "1.3.1", optional = true, default-features = false }
+serde_json = "1.0"
 
 wasm-bindgen = { version = "0.2.79", optional = true, features = ["serde-serialize"] }
-serde = { version = "1.0", optional = true, features = ["derive"]}
+serde = { version = "1.0", features = ["derive"]}
 
 # `wee_alloc` is a tiny allocator for wasm that is only ~1K in code size
 # compared to the default allocator's ~10K. It is slower than the default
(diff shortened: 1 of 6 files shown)


akd_client/src/verify.rs
```diff
@@ -135,6 +136,23 @@ fn verify_vrf(
     vrf_pk.verify_label(uname, stale, version, pi, label)
 }
 
+/// Verifies a serialized JSON lookup proof after deserializing it first.
+pub fn serialized_lookup_verify(
+    _vrf_public_key: &[u8],
+    root_hash: Digest,
+    _akd_key: AkdLabel,
+    serialized_json_proof: &str,
```
slawlor, line 144: So this is going to be a non-backwards compatible JSON string, since we're using `serde` which doesn't guarantee backwards compat right? If we are expecting this data to come over the wire, we should really be using something like protobuf unfortunately... :( Even the same structure, depending on json serializer, can have breakable differences in the serialized payload...

slawlor (changes requested):
If we do this, it means what's sent over-the-wire from the AKD to the client is a serialized JSON blob of the proof material. This requires that the server and client have exactly the same json serialization & deserialization logic with no backwards compatibility breaks.

If we really want to go this route, we'll need to encode protobuf or something similar for the over-the-wire payloads :/

akd_client/src/verify.rs
eozturk1: If the JSON serializer/deserializer are incompatible for the same structure, we would expect a major version difference between them since such a change would be breaking. I don't see this issue being applicable to `serde_json` since it is quite stable at this point and there hasn't been even a minor version bump in the last 5 years.

Overall, as long as we follow the general backward compatibility rules (e.g., never remove a field) I don't see much difference between `serde_json` and `protobuf` from backward compatibility perspective -- except `protobuf` requiring more manual work to write the structs, generate them by hand and re-parsing the primitive parts individually and mapping them to the Rust types (which is what we are trying to avoid). Is it possible to just do `let proof: LookupProof = protobuf:from_str(&serialized_proof)`? 

I also looked into using Thrift but `fbthrift` crate has not been updated in quite some time.

akd_client/Cargo.toml
```diff
@@ -16,9 +16,10 @@ crate-type = ["cdylib", "rlib"]
 sha2 = { version = "0.10.1", optional = true, default-features = false }
 sha3 = { version = "0.10.0", optional = true, default-features = false }
 blake3 = { version = "1.3.1", optional = true, default-features = false }
+serde_json = "1.0"
```
slawlor, line 19: nit: can you loosen this to just ` = "1"` so we get minor upgrade compat (should it occur)?

slawlor (approved):

slawlor:
Looks good!

eozturk1:
Merge after a rebase... not recommended. I had multiple commits that were overwriting each other with conflicts during cherry-picking. Clean up took a bit. :D

PR #239, opened 2022-09-09
Adding dynamo DB storage support for local auditor

slawlor:
This PR adds support for a dynamo DB based index of the epochs which provides faster listing operations than iterating the S3 bucket (in principal).

We can support further optimizations for queries of epochs in Dynamo in the future as well.

It also fixes a header which was incorrect


akd_local_auditor/src/storage/s3/test.rs
```diff
@@ -308,14 +245,69 @@ async fn integration_test_audit_verification() -> Result<()> {
...
 #[ignore]
 async fn populate_test_bucket() -> Result<()> {
     // Populates the test bucket for use with the command-line REPL via the command
     // cargo run -p akd_local_auditor -- s3 --bucket populatetestbucket --region us-east-2 --endpoint http://127.0.0.1:9000 --access-key test --secret-key someLongAccessKey
 
     // make sure we have a valid bucket name, that's "somewhat" unique
-    let bucket = bucket_name!();
+    let bucket = crate::common_test::alphanumeric_function_name!();
     log::debug!("Test bucket name is {}", bucket);
     assert!(matches!(validate_bucket_name(&bucket), Ok(_)));
```
eozturk1, line 309: I am a bit lost on the bucket name logic here... this generation and validation logic is duplicated in many places. I think we should have tests for the generation in a separate function and assume it works correctly when we use it.

Can you provide a bit more context on these bucket names?

akd_local_auditor/src/storage/s3/test.rs
```diff
@@ -308,14 +245,69 @@ async fn integration_test_audit_verification() -> Result<()> {
     Ok(())
 }
 
+// ============================ IMPORTANT ============================ //
+// NEVER run this operation in a CI pipeline, it takes 40+ min to complete
+// on reasonable hardware, and is only used for manual testing & timing information
+#[tokio::test]
+#[ignore]
+async fn expensive_audit_verification() -> Result<()> {
```
eozturk1, line 250: Maybe add a non-expensive audit verification to test this functionality works?

akd_local_auditor/src/storage/dynamodb/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+            let shared_config = shared_config_loader.load().await;
+
+            *lock = Some(shared_config.clone());
+            shared_config
+        }
+    }
+
+    fn parse_summary(row: &HashMap<String, AttributeValue>) -> Result<Option<EpochSummary>> {
+        if let (
+            Some(AttributeValue::N(epoch)),
```
eozturk1, line 181: nit: just a suggestion for readability, maybe rename as `dynamoAttributeValue`?

akd_local_auditor/src/storage/dynamodb/test.rs
```diff
@@ -0,0 +1,346 @@
...
+// This source code is licensed under both the MIT license found in the
+// LICENSE-MIT file in the root directory of this source tree and the Apache
+// License, Version 2.0 found in the LICENSE-APACHE file in the root directory
+// of this source tree.
+
+//! A note to readers:
+//!
+//! We are marking most of the following tests as `#[ignored]` so they don't run automatically
+//! with `cargo test`. These require a docker container running `minio` to be running which is
+//! run in our CI pipeline when changes to this crate are made, but not otherwise since they
```
eozturk1, line 12: With ignore they will be ignored regardless of where they are run, right? So we need to enable them later. Maybe create an issue for it?

eozturk1 (approved):
No blockers, some comments. Awesome work @slawlor! :)

akd_local_auditor/src/common_test.rs
```diff
@@ -0,0 +1,100 @@
...
+                    let value = format!("value{}_{}", _epoch, item);
+                    (
+                        AkdLabel::from_utf8_str(&user),
+                        AkdValue::from_utf8_str(&value),
+                    )
+                })
+                .collect::<Vec<(AkdLabel, AkdValue)>>();
+            akd.publish::<Hasher>(data).await?;
+        } else {
+            let t_value = format!("certificate{}", _epoch);
```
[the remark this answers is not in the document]
slawlor: just some binary key data, the `AkdValue`

akd_local_auditor/src/storage/dynamodb/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+            "Table name must be between [{}, {}] characters in length. Gave {}",
+            MIN_TABLE_CHARS,
+            MAX_TABLE_CHARS,
+            str.len()
+        ));
+    }
+
+    for c in str.chars() {
+        if !ALLOWED_TABLE_CHARS.iter().any(|v| c == *v) {
+            return Err(format!("Character '{}' is not allowed in table name. Table names must contain lower-case letters, numbers, '-', '_', and '.' only.", c));
```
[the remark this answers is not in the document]
slawlor: It avoids the round-trip and pre-validates prior to a dynamo call being made. If we wait for dynamo, the client needs to start the REPL and make a command before they get a table name error.

This fronts it so the repl will crash + report the problem to the user before an op occurs.

akd_local_auditor/src/storage/s3/test.rs
slawlor: So these are S3 storage bucket names. To guarantee parallel tests don't conflict with each other, each test uses a different bucket for its tests which is derived from the function name. However buckets cannot contain the "_" char and must be capped @ 63 chars. It's only for generating independent test cases

akd_local_auditor/src/storage/dynamodb/test.rs
slawlor: The AWS CI pipeline (`aws.yml`) will run them specifically with the --ignored flag in cargo. If you take a look there, you can see it only triggers on file-changes in a specific code path which then runs these more expensive tests. If no one changes this crate, the tests don't run to save cycles.

akd_local_auditor/src/storage/s3/test.rs
slawlor: added

PR #241, opened 2022-09-20
Verify that there is only 1 version of an object in S3 storage

slawlor:
S3's object lock does not prevent multiple versions of an object from being present (with the same key). This means, that naively once could enumerate the proofs and there be an updated proof without noticing, leading to a vulnerability. Since the previous + current hashes are part of the key naming, the entire proof would *technically* need to contain the same content, or there would be 2 separate keys with the same epoch, but this PR adds an explicit check of the object's version information to assert that if there are > 1 version for an object, they're all the same hash (`etag`), and if not error out and not validate the proof.

This should not be necessary, but in the event a publisher publishes the same proof multiple times (due to some technical glitch for example), that should not prohibit proof validation, and therefore we still want to be able to utilize this tool. However if multiple versions of a key are present, then we should return a legitimate error!

### DIFF
--- a/akd_local_auditor/src/storage/s3/mod.rs
+++ b/akd_local_auditor/src/storage/s3/mod.rs
@@ -205,7 +205,7 @@ impl S3AuditStorage {
                                 version
                                     .e_tag()
                                     .map(|tag| tag.to_string())
-                                    .unwrap_or("".to_string()),
+                                    .unwrap_or_else(|| "".to_string()),
                             );
 
                             let this_time: std::time::SystemTime = (*mod_time).try_into()?;
@@ -226,9 +226,9 @@ impl S3AuditStorage {
         } else if version_count > 1 {
             // check all the versions for their associated etags, and make sure they're all the same
             let first = &etags[0];
-            for i in 1..etags.len() {
-                if first.cmp(&etags[i]) != std::cmp::Ordering::Equal {
-                    return Err(anyhow::anyhow!("There were duplicate objects with the same key that have different etags which indicates different values. This epoch cannot be trusted ({} != {})", first, etags[i]));
+            for etag in etags.iter().skip(1) {
+                if first.cmp(etag) != std::cmp::Ordering::Equal {
+                    return Err(anyhow::anyhow!("There were duplicate objects with the same key that have different etags which indicates different values. This epoch cannot be trusted ({} != {})", first, etag));
                 }
             }
         }
--- a/akd_local_auditor/src/storage/s3/mod.rs
+++ b/akd_local_auditor/src/storage/s3/mod.rs
@@ -181,10 +181,11 @@ impl S3AuditStorage {
     async fn get_object(&self, key: &str) -> Result<s3::client::fluent_builders::GetObject> {
         // We need to retrieve the minimum version of an object, to guarantee we see the ORIGINAL
         // status of the blob. Future versions are not supported
+
         let attributes = s3::Client::from_conf(self.get_config().await)
             .list_object_versions()
             .bucket(self.bucket.clone())
-            .key_marker(key.to_string())
+            .prefix(key.to_string())
             .max_keys(1)
             .send()
             .await?;
--- a/akd_local_auditor/src/storage/s3/mod.rs
+++ b/akd_local_auditor/src/storage/s3/mod.rs
@@ -193,7 +193,7 @@ impl S3AuditStorage {
         let mut version_count = 0usize;
         let mut min_time = std::time::SystemTime::now();
         let mut min_version = String::new();
-        let mut etags = vec![];
+        let mut first_etag = None;
         if let Some(versions) = attributes.versions() {
             for version in versions {
                 if let Some(potential_key) = version.key() {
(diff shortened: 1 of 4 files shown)


slawlor:
Related PR: #227

akd_local_auditor/src/storage/s3/test.rs
```diff
@@ -60,14 +60,40 @@ pub async fn maybe_flush_storage(shared_config: &SdkConfig, bucket_name: &str) -
...
+            let object_versions = client
+                .list_object_versions()
+                .bucket(bucket_name)
+                .max_keys(1000)
+                .send()
+                .await?;
+            let map = object_versions
+                .versions()
+                .unwrap()
+                .iter()
```
dillonrg, line 73: I think you can call `into_iter()` here since you're not using object versions again.

akd_local_auditor/src/storage/s3/mod.rs
```diff
@@ -178,14 +178,68 @@ impl S3AuditStorage {
...
+        let attributes = s3::Client::from_conf(self.get_config().await)
+            .list_object_versions()
+            .bucket(self.bucket.clone())
+            .prefix(key.to_string())
+            .max_keys(1)
+            .send()
+            .await?;
+
+        let mut version_count = 0usize;
+        let mut min_time = std::time::SystemTime::now();
```
eozturk1, line 194: We shouldn't assume clock synchrony here between the device this code is running on and the S3 modification times. Why not use only the modification times to calculate the min?

akd_local_auditor/src/storage/s3/mod.rs
```diff
@@ -178,14 +178,68 @@ impl S3AuditStorage {
...
+
+        if version_count == 0 {
+            return Err(anyhow::anyhow!(
+                "Object not found with any version information"
+            ));
+        } else if version_count > 1 {
+            // check all the versions for their associated etags, and make sure they're all the same
+            let first = &etags[0];
+            for etag in etags.iter().skip(1) {
+                if first.cmp(etag) != std::cmp::Ordering::Equal {
```
eozturk1, line 231: Can't we perform this check in the previous loop and remove this?

akd_local_auditor/src/storage/s3/test.rs
```diff
@@ -60,14 +60,40 @@ pub async fn maybe_flush_storage(shared_config: &SdkConfig, bucket_name: &str) -
...
+                .map(|obj_ver| {
+                    (
+                        obj_ver.key().unwrap().to_string(),
+                        obj_ver.version_id().unwrap().to_string(),
+                    )
+                })
+                .collect::<std::collections::HashMap<String, String>>();
+            let get_version = |key: &str| {
+                let result = map.get(key).cloned();
```
dillonrg, line 82: I think you can call `map_or_else` and panic if you hit the else. I think this may be sightly more readable and (potentially) microscopically more performant since you won't always call `clone`.

akd_local_auditor/src/storage/s3/mod.rs
```diff
@@ -181,10 +181,11 @@ impl S3AuditStorage {
     async fn get_object(&self, key: &str) -> Result<s3::client::fluent_builders::GetObject> {
         // We need to retrieve the minimum version of an object, to guarantee we see the ORIGINAL
         // status of the blob. Future versions are not supported
+
         let attributes = s3::Client::from_conf(self.get_config().await)
             .list_object_versions()
             .bucket(self.bucket.clone())
-            .key_marker(key.to_string())
+            .prefix(key.to_string())
```
dillonrg, line 188: Why does this change to prefix?

eozturk1 (commented):
Thank you for working on this! I added couple of comments.  Can we add a test for multiple versions different values case we are adding a prevention for here?

dillonrg (approved):
None of my comments are hard-blockers, but mostly suggestions with a question thrown in for good measure. Please feel free to incorporate or ignore the suggestions that I left!

akd_local_auditor/src/storage/s3/mod.rs
slawlor: So that's exactly what we're using, however (unfortunately) the AWS SDK's implementation of `DateTime` doesn't support comparisons. So we convert it to a local `SystemTime` to do the comparisons. Now I understand, if we were trying to compare against a local system time, we'd have a problem with clock drift, however we're converting ALL of the aws `DateTime`s to `SystemTime` to the relative differences should be the same (i.e. identifying the minimum time shouldn't be impacted).

There's many quirks in the AWS SDK...

akd_local_auditor/src/storage/s3/test.rs
slawlor: `clippy` actually reverted my `into_iter()`. Since we're calling `.to_string()` anyways, it said we can do it with a borrow.

akd_local_auditor/src/storage/s3/test.rs
slawlor: No matter what, here we need to clone, since we want an `Option<String>` not an `Option<&String>` which wouldn't support downstream comparisons on equality I think.

akd_local_auditor/src/storage/s3/test.rs
dillonrg: Ah, interesting! Good to know :)

PR #252, opened 2022-11-06
Disable build.rs in akd/ as well

kevinlewi:
Similar to #251 but also covers the akd/ Cargo.toml in addition to akd_client/.

Right now, running `cargo test` in the root directory produces an error because of this.


slawlor (approved):
Looks good, thanks for doing this!

PR #255, opened 2022-11-07
Retry: Update to most recent VRF draft.

slawlor:
Previously reverted PR for VRF update can now be landed with updated clippy lints applied

* Update to most recent VRF draft.

* Apply same changes to testing ECVRF implementation.


slawlor:
> Sorry for this! What should I have run to prevent it?

I don't think it was you actually! I think it's because Rust released a new version which came with a new Clippy instance. I prematurely removed the VRF update, but we're just checking if we want to merge it now or leave it on a "dev" branch until @kevinlewi is complete with a bunch of other code cleanups too. You're good!

slawlor:
I think this is ready to merge in now, as any Rust update identified problems have been resolved.

kevinlewi (approved):

PR #256, opened 2022-11-08
Upgrade protobuf to version 3.2.0

slawlor:
As titled, we're bringing protobuf up to the latest specification.

This PR 

1. Regenerates the protobuf generated code
2. Updates all the converters to handle the new type formats
3. Re-checks all tests


akd_client/src/proto/mod.rs
```diff
@@ -355,31 +369,28 @@ impl TryFrom<&types::UpdateProof> for crate::UpdateProof {
...
+            .as_ref()
+            .map(|item| item.to_vec());
         let previous_version_stale_at_ep: Option<crate::MembershipProof> =
-            if input.has_previous_version_stale_at_ep() {
-                Some(input.get_previous_version_stale_at_ep().try_into()?)
+            if let Some(item) = input.previous_version_stale_at_ep.as_ref() {
+                Some(item.try_into()?)
             } else {
                 None
             };
```
afterdusk, line 383: You can also do:
```rust
let previous_version_stale_at_ep = input
    .previous_version_stale_at_ep
    .as_ref()
    .map(|item| item.try_into())
    .transpose()?;
```

akd/src/proto/mod.rs
```diff
@@ -60,6 +60,17 @@ macro_rules! require {
...
+    ($obj:ident, $field:ident) => {
+        if !$obj.$field.is_some() {
+            return Err(LocalAuditorError::RequiredFieldMissing(
+                stringify!($obj).to_string(),
+                stringify!($has_field).to_string(),
+            ));
+        }
+    };
+}
+
```
afterdusk, line 73: It's odd that struct fields need to be handled differently from native type fields, but from a quick look at the library docs I couldn't find a unified API, so this is probably the best we can do for now.

afterdusk (approved):
LGTM :)

akd/src/proto/mod.rs
slawlor: yeah I spent a good amount of time on this, and I don't see a workaround unfortunately :(. It'll be fun with the other protobufs we have...

PR #263, opened 2022-11-12
Moving hasher generic from directory functions to its struct

### LINKED ISSUE #178: Push generic for hasher from directory functions to directory struct
Right now, the API of a Directory in directory.rs requires that all of the functions implemented on Directory need to specify the Hasher. Since the hasher should be consistent across all of the functions used for the same directory, it would be nice for a consumer to only have to specify the hasher on directory initialization. Therefore, we could add the Hasher to the Directory instead of to each of its functions.

This means changing the current struct definition:
`pub struct Directory<S, V>`
into
`pub struct Directory<S, V, H>`

and then removing `<H: Hasher>` from each of its functions.

This will affect any downstream code that uses these high-level functions of course, so care should be taken in making sure everything is compatible still.

kevinlewi:
Closes #178.

Should be merged only after merging #262.

This could be merged against the dev branch instead of the main branch if that makes more sense. Since it introduces breaking changes to the high-level API.

### DIFF
--- a/akd/src/directory.rs
+++ b/akd/src/directory.rs
@@ -24,7 +24,7 @@ use log::{debug, error, info};
 use rand::{distributions::Alphanumeric, CryptoRng, Rng};
 
 use std::collections::HashMap;
-use std::marker::{Send, Sync};
+use std::marker::{PhantomData, Send, Sync};
 use std::sync::Arc;
 use winter_crypto::{Digest, Hasher};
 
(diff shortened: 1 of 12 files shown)


akd/src/directory.rs
```diff
@@ -24,7 +24,7 @@ use log::{debug, error, info};
 use rand::{distributions::Alphanumeric, CryptoRng, Rng};
 
 use std::collections::HashMap;
-use std::marker::{Send, Sync};
+use std::marker::{PhantomData, Send, Sync};
```
slawlor, line 27: Question: Is there a reason we're putting phantomdata back? We did a big removal at one point to remove it, but I guess this maybe makes a cleaner API?

kevinlewi: Oh, oops, what was the reason for removing PhantomData? I was indeed thinking that this makes for a cleaner API but there are alternatives if we want to avoid using PhantomData.

slawlor: I think it was to minimize the number of generics on the struct, but it's also isn't a big "technical" change. Probably not specifying the generic on every method call is overall easier I guess.

PR #266, opened 2022-11-18
Changing lookup + history verification APIs to return VerifyResult structs

kevinlewi:
Lookup proofs and history proofs now return a `VerifyResult` struct, defined as follows:
```
pub struct VerifyResult {
    /// The epoch of this record
    pub epoch: u64,
    /// Version at this update
    pub version: u64,
    /// The plaintext value associated with the record
    pub value: AkdValue,
}
```
That way, a consumer of the lookup proof / history proof verification API can potentially use these plaintext values to check consistency with the "expected" values.

Also in this PR, I adjusted the API for history proof generation and verification to use enums for custom parameters:

For history proof generation, in order to support the creating of a limited key history, we now have this enum:
```
pub enum HistoryParams {
    /// Returns a complete history for a label
    Complete,
    /// Returns up to the most recent N updates for a label
    MostRecent(usize),
    /// Returns all updates since a specified epoch (inclusive)
    SinceEpoch(u64),
}
```
So a HistoryParams::MostRecent(3) will return the 3 most recent updates for the history proof.

For history proof verification, we used to have a boolean value denoting whether or not we should allow tombstoned values. This has been replaced with the following enum:
```
pub enum HistoryVerificationParams {
    /// No customization to the verification procedure
    Default,
    /// Allows for the encountering of missing (tombstoned) values
    /// instead of attempting to check if their hash matches the leaf node
    /// hash
    AllowMissingValues,
}
```

### DIFF
--- a/akd/src/directory.rs
+++ b/akd/src/directory.rs
@@ -409,104 +409,10 @@ impl<S: Storage + Sync + Send, V: VRFKeyStorage, H: Hasher> Directory<S, V, H> {
     /// this function returns all the values ever associated with it,
     /// and the epoch at which each value was first committed to the server state.
     /// It also returns the proof of the latest version being served at all times.
-    pub async fn key_history(&self, uname: &AkdLabel) -> Result<HistoryProof<H>, AkdError> {
-        // The guard will be dropped at the end of the proof generation
-        let _guard = self.cache_lock.read().await;
-
-        let username = uname.to_vec();
-        let current_azks = self.retrieve_current_azks().await?;
-        let current_epoch = current_azks.get_latest_epoch();
-
-        if let Ok(this_user_data) = self.storage.get_user_data(uname).await {
-            let mut user_data = this_user_data.states;
-            // reverse sort from highest epoch to lowest
-            user_data.sort_by(|a, b| b.epoch.partial_cmp(&a.epoch).unwrap());
-
-            let mut update_proofs = Vec::<UpdateProof<H>>::new();
-            let mut last_version = 0;
-            for user_state in user_data {
-                // Ignore states in storage that are ahead of current directory epoch
-                if user_state.epoch <= current_epoch {
-                    let proof = self.create_single_update_proof(uname, &user_state).await?;
-                    update_proofs.push(proof);
-                    last_version = if user_state.version > last_version {
-                        user_state.version
-                    } else {
-                        last_version
-                    };
-                }
-            }
-            let next_marker = get_marker_version(last_version) + 1;
-            let final_marker = get_marker_version(current_epoch);
-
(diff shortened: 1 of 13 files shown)


slawlor (approved):
I'm not sure I 100% agree with the removal of the "limited" key history, but perhaps it does make more sense to move to an epoch-limited window. It's not currently utilized here, but in a downstream application I would think the last >= 3 updates do make sense because that would allow a client, who doesn't hold their anything but their current key, to see if their key was replaced. 

I.e. 

1. Key `A` is the client's current key
2. The client performs a limited key history of the last 3 updates
3. This would allow them to identify if an attack from the server (or another party) where the history shows `A`, then a false `A*`, then back to the true `A` occurred.

Does that make sense?

kevinlewi:
I see. Yes, this sounds valuable! I am still hoping for enabling the support of being able to view the entire key history since the last epoch that was checked, though. Maybe one way we can support both would be to add an optional configuration flag to the key history function which allows the API consumer to specify what types of limitations on the key history it would like (last N updates versus epoch-limited window, or a mix of both even). That way, we aren't getting rid of functionality.

Also, I realized I wanted to extend this PR slightly to have the epoch and version returned by these functions, in addition to the value. Let me put this back in draft mode and make some edits.

Thanks for the feedback!

kevinlewi:
@slawlor: Took another stab at things and should be ready for review now. Hopefully this is overall a cleaner API, but open to suggestions and let me know if you disagree!

I also re-instated the limited_key_history functionality and its associated test, by combining it with key_history but with custom parameters specified by an enum.

akd/src/directory.rs
```diff
@@ -409,189 +409,101 @@ impl<S: Storage + Sync + Send, V: VRFKeyStorage, H: Hasher> Directory<S, V, H> {
...
+        // reverse sort from highest epoch to lowest
+        user_data.sort_by(|a, b| b.epoch.cmp(&a.epoch));
+
+        // apply filters specified by HistoryParams struct
+        user_data = match params {
+            HistoryParams::Complete => user_data,
+            HistoryParams::MostRecent(n) => user_data.into_iter().take(n).collect::<Vec<_>>(),
+            HistoryParams::SinceEpoch(epoch) => user_data
+                .into_iter()
+                .filter(|val| val.epoch >= epoch)
```
slawlor, line 433: Do we want to sort again after the filtering? I am assuming ordering will be maintained, but we should at least check if another sort is necessary.

slawlor (approved):
Looks great! Very clean, thanks for this!

akd/src/directory.rs
kevinlewi: Thanks, I went ahead and added another sorting after the filtering, just in case.

PR #267, opened 2022-11-18
Fixing mysql_async dependency issues

kevinlewi:
funty 1.2.0 was yanked, which was a dependency for mysql_async v0.29. Upgrading to mysql_async v0.31 fixes the issue.

### DIFF
--- a/akd_mysql/src/mysql_storables.rs
+++ b/akd_mysql/src/mysql_storables.rs
@@ -620,7 +620,11 @@ impl MySqlStorable for DbRecord {
         Self: std::marker::Sized,
     {
         fn cast_err() -> MySqlError {
-            MySqlError::from("Failed to cast label:val into [u8; 32]".to_string())
+            MySqlError::from(mysql_async::ServerError {
+                state: "".to_string(),
+                code: 0,
+                message: "Failed to cast label:val into [u8; 32]".to_string(),
+            })
         }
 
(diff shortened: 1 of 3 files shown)


akd_mysql/src/mysql_storables.rs
```diff
@@ -641,8 +645,16 @@ impl MySqlStorable for DbRecord {
                             val.try_into().map_err(|_| cast_err())?,
                             len,
                         ))),
-                        (Err(_len_err), _) => Err(Error::from("Error decoding length")),
-                        (_, Err(_val_err)) => Err(Error::from("Error decoding value")),
+                        (Err(_len_err), _) => Err(Error::from(mysql_async::ServerError {
```
slawlor, line 648: Are these conversions necessary? It looks like we're just converting to the base Error type. Or am I misreading it and it's a MySQL error type?

slawlor (approved):
Lgtm! Thanks for keeping up with this!

akd_mysql/src/mysql_storables.rs
kevinlewi: It's a MySQL error type. The problem was that in 0.29, MySQL implemented the From<& static str> trait, but support for that was removed in 0.30 onwards. So now I decided to just the more cumbersome notation here...

PR #268, opened 2022-11-20
Removing the ability to disable VRFs by baking in the vrf feature

### LINKED ISSUE #261: Bake in VRF functionality (eliminating support for disabling VRFs)
The default behavior is to enable VRFs. But this causes funky logic which could ideally be simplified. For instance, [here](https://github.com/novifinancial/akd/blob/832f66ddd1e700728453fb8d42aadddb44269be8/akd_client/src/verify.rs#L159-L164), a client is supposed to pass a VRF public key and AkdLabel if VRF is enabled, but if VRF is disabled, then these two fields can just be empty.

Let's simplify this by not allowing for consumers of the library to disable VRFs (since I think it should be considered a core privacy feature anyway). This would involve getting rid of the VRF feature (so that it can't be disabled).

kevinlewi:
Closes #261.


slawlor (approved):
LGTM! Thanks for making our feature set cleaner and more concise!

PR #269, opened 2022-11-22
Performance enhancements related to caching

### LINKED ISSUE #217: Fix faulty azks polling test
The following test is intended to assert that AZKS polling produces a correct view of the tree for a reader in the presence of a concurrent writer and independent caches.

https://github.com/novifinancial/akd/blob/97b2e90589083fef4600f732618715fac29a24f2/akd/src/tests.rs#L596-L655

However, the in-memory DB used (`AsyncInMemoryDatabase`) does not have a cache. `AsyncInMemoryDbWithCache` should be used instead. That being said, a simple substitution to `AsyncInMemoryDbWithCache` would not work as the cache is managed by the DB. Both the reader and writer would need to share a cache to access the same in-memory DB.

A fix would be to port over this test to the `integration_tests` crate and use a `AsyncMySqlDatabase`. Both the reader and writer can initialize separate DB clients with independent caches connected to the same DB.

slawlor:
This PR adds support for 

1. `dashmap` which is a more performant multi-access cache when there are multiple readers and writers
2. Removing `tokio::sync::RwLock<bool>` and replacing it with an `AtomicBool` which is also more performant under high-parallel load
3. Adding support for memory pressure processing to limit the cache size based on the memory pressure on the system
4. A complete rewrite of the storage layer to a `Database` trait and `StorageManager` struct which manages the specified database. This greatly simplifies what a db needs to implement and abstracts away caching & transaction management to a shared logic space.

Update: closes #217


slawlor:
A note to reviewers, it's probably better to review this commit-by-commit. It'll make more sense and break-down the changes to more manageable amounts

akd/src/append_only_zks.rs
```diff
@@ -44,6 +44,12 @@ pub struct Azks {
     pub num_nodes: u64, // The size of the tree
 }
 
+impl SizeOf for Azks {
+    fn size_of(&self) -> usize {
+        std::mem::size_of::<u64>() * 2
```
eozturk1, line 49: Why not `mem::size_of::<Azks>()` and similarly for other structs?

slawlor: It can't measure dynamic structs which are on the heap, like `Vec<_>`'s. It'll give a constant size for all vectors since the `Vec` struct itself is backed by a `RawVec` which is just a pointer into memory. We need our own size measurement tool to properly multiply by the size of the vector and measure the size of the contained elements. I use the standard one for primitive types, but for safety stay away from it for our structs.

akd/src/storage/cache/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+//! This module handles the caching implementation and testing for a time-based cache
+//! which supports memory pressure shedding
+
+use crate::storage::DbRecord;
+use std::time::Instant;
+
+#[cfg(test)]
+mod tests;
+
+/// item's live for 30s
```
kevinlewi, line 17: nit: remove extra apostrophe

kevinlewi (approved):
Overall looks good to me. The separating of the cache logic from the db is nice to have for sure!

PR #271, opened 2022-11-24
Update bugged get_root_hash_at_epoch function

afterdusk:
The API does not function as intended - instead of getting the root hash at a specific epoch, it gets the root hash at the latest epoch. 

As we no longer store all nodes from past epochs, we are unable to generate root hashes of past epochs with the directory. Therefore I update the function to `get_root_hash_safe`, which instead checks that the input epoch is equivalent to the latest epoch before returning the root hash (of the latest epoch). This gives callers a way to know if the root hash they are trying to retrieve is not from the epoch that they expect.

### DIFF
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -9,7 +9,7 @@
 use crate::serialization::to_digest;
 use crate::storage::manager::StorageManager;
 use crate::{
-    errors::TreeNodeError,
+    errors::{DirectoryError, TreeNodeError},
     proof_structs::{AppendOnlyProof, MembershipProof, NonMembershipProof, SingleAppendOnlyProof},
     storage::{Database, SizeOf, Storable},
     tree_node::*,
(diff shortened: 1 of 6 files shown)


akd/src/append_only_zks.rs
```diff
@@ -523,31 +523,27 @@ impl Azks {
...
             .await
     }
 
-    /// Gets the root hash of the tree at a epoch.
-    /// Since this is accessing the root node and the root node exists at all epochs that
-    /// the azks does, this would never be called at an epoch before the birth of the root node.
-    pub async fn get_root_hash_at_epoch<S: Storage + Sync + Send, H: Hasher>(
+    /// Gets the root hash of the tree at the latest epoch if the passed epoch
+    /// is equal to the latest epoch. Will return an error otherwise.
+    pub async fn get_root_hash_safe<S: Storage + Sync + Send, H: Hasher>(
```
slawlor, line 532: Should we just make this `get_root_hash` and not take an epoch at all? I know it's a bigger refactoring, but if there's only 1 valid value of `epoch` we support, and we can get it from `self.latest_epoch` then perhaps we shouldn't take it as an arg at all...?

afterdusk: It's more for the caller to validate that the root hash being fetched is at the epoch that the caller expects. Say for example the caller is in a state where it thinks the epoch is 5, and wants to get the root hash for epoch 5. However the directory is already at epoch 6.

With the regular `get_root_hash`function, the caller would get the root hash for epoch 6 and treat it as the root hash for epoch 5. With this API the directory can return an Err and trigger any recovery mechanisms for inconsistent state. Does that make sense?

slawlor: Yeah OK that makes sense. Thanks!

slawlor (approved):

slawlor:
Please merge from main before pushing commit.

afterdusk:
Rebased on main

Issue #270, opened 2022-11-23
Compiling to WebAssembly targets


chris-wood:
`akd` seems to depend upon `tokio` for some basic functionality like time and read-write locks, as well as test support. This is all well and fine, but unfortunately `tokio` has a dependency on `mio`, which does not compile for Wasm targets. Is it possible to support compilation to wasm targets? That would allow this code to be run in browser and related runtimes.

kevinlewi:
I think mainly the server-side functionality is what needs the tokio dependency for async functions. However, the client-side functionality (and audit verification functionality) should be able to be run without tokio, and should hopefully be wasm-compatible.

At the moment, it's not organized like this, but thanks for bringing this up! We can certainly use this issue to track progress on wasm support for proof verification operations.

@chris-wood: Is this what you meant by compiling to WebAssembly targets? Or were you also looking for compatibility for server-side functionality?

chris-wood:
Yeah, I'm thinking about the client-side functionality, which includes audit verification. I'm having trouble compiling to wasm. It's very possible I'm holding it wrong. What features would one need to include to implement auditing functionality? At a high level, I would expect that to be `public_auditing` and `serde_serialization`.

kevinlewi:
Another option you can take a look into is using the akd_client repo instead, which does support wasm, but only provides client functionality (not verification of audit proofs). Actually, it's likely that we will keep the main akd repo to have the tokio dependency and not support wasm, whereas the akd_client repo will be where all of the wasm-compatible code lives. Sorry, I think my original answer did not factor in the existing wasm support in akd_client, which might have caused some confusion.

Now, akd_client doesn't support the [audit_verify](https://github.com/novifinancial/akd/blob/main/akd/src/auditor.rs#L22) operation at the moment, but perhaps this could be added in. @slawlor heads up on that

chris-wood:
Open to PRs on the client repo to add audit_verify? =)

kevinlewi:
I think that should be fine if it is gated behind a feature!

chris-wood:
Great! What do you think about moving the audit code inside `akd` into an `akd_audit` module, similar to what's done for `akd_client`, `akd_mysql`, etc?

kevinlewi:
Sounds like a good idea. Thanks for the suggestion!

slawlor:
[...]

PR #277, opened 2022-11-28
Minimize the dependencies on Tokio. 

slawlor:
This change leans out the features in the `tokio` crate to just "sync" and "time" for synchronization and communication management features utilized

in the `akd` crate. In the test configuration, more functionality is enabled with "rt" and "macros" due to the tokio testing frameworks utilized in `akd`. This removes the inherited
dependency on `mio` as well as many other downstream crates which should assist in WASM compliation (see: #270)

### DIFF
--- a/akd/Cargo.toml
+++ b/akd/Cargo.toml
@@ -12,7 +12,7 @@ readme = "../README.md"
 # build = "src/build.rs"
 
 # [build-dependencies]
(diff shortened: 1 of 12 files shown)


kevinlewi (approved):
LGTM!

chris-wood (approved):

PR #278, opened 2022-11-29
Migrate shared logic to `akd_core` crate which contains all the base type implementations

slawlor:
Shared logic relates to 

1. hashing (with this we can remove dependence on `winter_*`)
2. ecvrf (removes code duplication in multiple crates)
3. types specifications and protobuf specifications (more common logic removal)
4. clearer logic separation and test coverage on core logic components

`akd_core` is specifically very flexible (i.e. a lot of features) in order to facilitate multiple ways of utilization of `akd` and `akd_client`. It isn't meant to be utilized directly, and its features should be set and managed by the parent crate(s).

Additional CI coverage has also been added.

NOTE: This also addresses potentially compiling `akd` to WASM targets for auditing (#270)


slawlor:
# NOTE: To all reviewers

This change is in 3 commits, and is likely best reviewed in that order. 

1. is the initial creation of `akd_core` along with most of the necessary feature flags along with the porting of `akd_client` which is mainly deleting code
2. Is the porting of the `akd` crate onto `akd_core` and removal of all `winter_*` references. This is a rather large refactoring and the other crates are broken after this commit
3. Is the final cleanup of all remaining crates for the migration and removal of all instances of `winter_*` globally

kevinlewi:
Amazing

akd/src/errors.rs
```diff
@@ -273,36 +280,36 @@ impl fmt::Display for StorageError {
...
+//             }
+//             Self::PublicKey(error_string) => {
+//                 write!(f, "VRF public key: {}", error_string)
+//             }
+//             Self::Verification(error_string) => {
+//                 write!(f, "VRF proving or verifying: {}", error_string)
+//             }
+//         }
+//     }
+// }
```
afterdusk, line 312: To be removed?

kevinlewi:
One consequence of this PR is that we now use features to select which underlying hash function AKD uses, whereas before we used traits/types.

An alternative design would be to use the `Digest` trait provided by https://crates.io/crates/digest (for which the `sha2`, `sha3` crates implement as well), and retain the ability for consumers of the API to specify the hash function in code instead of using feature flags. This would be similar to the use of the `Hasher` trait that we previously inherited from `winter_crypto`, except that `Digest` is widely supported among hash function implementation in Rust and already a dependency in sha2 and sha3.

I'm ok with shifting over to feature-gating the hash function but just wanted to call out that this isn't strictly necessary.

kevinlewi (approved):
Overall looks great though! Very excited for this refactor.

slawlor:
> An alternative design would be to use the `Digest` trait provided by https://crates.io/crates/digest (for which the `sha2`, `sha3` crates implement as well), and retain the ability for consumers of the API to specify the hash function in code instead of using feature flags. This would be similar to the use of the `Hasher` trait that we previously inherited from `winter_crypto`, except that `Digest` is widely supported among hash function implementation in Rust and already a dependency in sha2 and sha3.

So @kevinlewi, after looking at it a bit, you're right that `sha2` and `sha3` will auto-import this crate and already implement it, however `blake3` doesn't by default and requires enabling a "preview" functionality. Additionally from the `blake3` crate's readme docs

> However, the traits aren’t stable, and they’re expected to change in incompatible ways before that crate reaches 1.0. For that reason, this crate makes no SemVer guarantees for this feature, and callers who use it should expect breaking changes between patch versions.

So I'd be a little wary of depending on the Digest crate just yet I think. Using that trait everywhere would be going back to a similar notation like we had in `winter_*` with `<D: Digest>`'s everywhere to support the generic based on the implementation someone was trying to use. Realistically, you can't run two hash functions in the same tree, so it seems like this would be something someone would choose upon import of the crate, and hard-set for all future. 

Personally, my opinion, is that our usage warrants a compile-time enforcement of the hash utility being utilized. We also get digest sizing at compile time which can give some further guarantees (example: database column sizing, an example is in `akd_mysql` which takes the `akd::DIGEST_BYTES` size for the column size of a hash). `Digest` does have an `output_size()` functionality, but I think this is a weaker guarantee than length asserted byte arrays given the hash implementation.

akd_client/Cargo.toml
```diff
@@ -1,70 +1,50 @@
...
 
 [lib]
 crate-type = ["cdylib", "rlib"]
 
 [dependencies]
-akd = { path = "../akd", version = "^0.7.7", features = ["public-tests"], optional = true }
-blake3 = { version = "1.3.1", optional = true, default-features = false }
-curve25519-dalek = "3"
-ed25519-dalek = { version = "1", default-features = false, features = ["u64_backend", "rand"] }
+# ======== Required Dependencies ======== #
```
afterdusk, line 16: Nit: Let's keep the comment headers consistent between `Cargo.toml` files. `akd/Cargo.toml` uses:
```
## Required dependencies ##

## Optional Dependencies ##
```

akd_core/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+//! which features are utilized, you can pass --no-default-features on the command line
+//! or `default-features = false` in your Cargo.toml import to disable all the default features
+//! which you can then enable one-by-one as you wish.
+
+#![warn(missing_docs)]
+#![cfg_attr(docsrs, feature(doc_cfg))]
+#![cfg_attr(feature = "nostd", no_std)]
+extern crate alloc;
+
+#[cfg(all(feature = "protobuf", not(feature = "nostd")))]
```
afterdusk, line 23:
```suggestion
#[cfg(all(feature = "protobuf_serialization", not(feature = "nostd")))]
```

akd_client/src/lib.rs
```diff
@@ -44,184 +34,20 @@
...
-// Types are re-exported at the root level for visbility
-mod types;
-pub use types::*;
-// verify types are not re-exported, to not clutter the root path
-pub mod verify;
+// Re-expose the core functionality from akd_core (verifications, types, etc)
+pub use akd_core::verify;
+pub use akd_core::*;
 
 #[cfg(feature = "protobuf")]
```
afterdusk, line 47:
```suggestion
#[cfg(feature = "protobuf_serialization")]
```

afterdusk: Hmm actually `protobuf` should work as well, there's an implicit feature defined by the `protobuf` dependency. Feel free to ignore this comment

akd/src/storage/types.rs
```diff
@@ -303,15 +227,15 @@ impl DbRecord {
...
+        hash: [u8; crate::DIGEST_BYTES],
         p_last_epoch: Option<u64>,
         p_least_descendant_ep: Option<u64>,
         p_parent_label_val: Option<[u8; 32]>,
         p_parent_label_len: Option<u32>,
         p_node_type: Option<u8>,
         p_left_child: Option<NodeLabel>,
         p_right_child: Option<NodeLabel>,
-        p_hash: Option<[u8; 32]>,
+        p_hash: Option<[u8; crate::DIGEST_BYTES]>,
```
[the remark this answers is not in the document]
slawlor: Great suggestion! Cleaning up :)

akd_client/src/lib.rs
slawlor: <3

akd_core/Cargo.toml
```diff
@@ -0,0 +1,59 @@
...
+
+[build-dependencies]
+protobuf-codegen = "3.2"
+protobuf-parse = "3.2"
+
+[features]
+# Disable all STD for the crate
+nostd = []
+# Default feature mix (blake3 + vrf verification logic)
+default = ["blake3/std", "vrf"]
```
afterdusk, line 22: Does this mean `std` will always have to be imported when using `blake3`?

.github/workflows/ci.yml
```diff
@@ -25,10 +25,29 @@ jobs:
...
+          command: test
+          args: --package akd_core --features nostd,blake3,vrf --no-default-features
+
+      - name: Test the core crate (akd_core) with protobuf and serde enabled
+        uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --package akd_core --features protobuf,serde_serialization
+
+      - name: Test the core crate (akd_core) with no standard library
```
afterdusk, line 46: Nit: Specify that sha hashing is enabled, as currently the stage name is identical to the one above

afterdusk (approved):
Finally done looking over all files 🙂 Excellent work @slawlor ! Removing Winter and using conditional compilation over generics to specify the hash cleans up the code SO much!

akd_core/Cargo.toml
slawlor: So at it's core import, we do **not** bring in the `std` feature on `blake3`, so if you wanted nostd you would do something like

```toml
akd_client = { version = 0.8, default-features=false, features = ["blake3"] }
```

which would not select the `default` feature and therefore not bring in std through `blake3`. The syntax is a bit weird, I'll admit, and I'm looking at cleaning it up. It seems the more common pattern is to not have a `nostd` feature, but rather an `std` feature so the native functionality is `nostd`. I might (in a later PR) match that common pattern.

PR #281, opened 2022-12-06
Best practices for build.rs code generation

slawlor:
We shouldn't be generating the protobuf sources **directly** in the source tree apparently. This change generates to the targets directory and then uses the standard-practice `include!` macro to include the generated code when needed

https://doc.rust-lang.org/cargo/reference/build-script-examples.html


akd_core/src/build.rs
```diff
@@ -14,23 +14,24 @@
 const PROTOBUF_BASE_DIRECTORY: &str = "src/proto/specs";
 /// The list of protobuf files to generate inside PROBUF_BASE_DIRECTORY
 const PROTOBUF_FILES: [&str; 1] = ["types"];
+/// The output directory in the cargo build folder to emit the generated sources to
+const OUTPUT_RELATIVE_DIR: &str = "protos";
```
dillonrg, line 18: Super tiny nit: This still feels quite specific to protos despite being in a seemingly generic location. I think a better name would be `PROTOS_OUTPUT_DIR` (can include relative if you'd like).

dillonrg (approved):
Looks good to me! Just one tiny suggestion that isn't blocking.

PR #283, opened 2022-12-07
Adding specific targets to CI for testing wasm and no_std

kevinlewi:
This should ensure that if we make a change that prevents a successful compilation wasm, our CI should catch it. Same with no_std compatibility.


kevinlewi:
No longer needed, closing

PR #286, opened 2022-12-08
Moving the local auditing module and adjusting proof generation return types

### LINKED ISSUE #285: Generated proofs should contain the associated root hash to solve them from the same epoch guaranteed
In the current system, it's 2 calls to generate a root hash and a proof. This means that between proof generation and root-hash generation it's possible the epoch is advanced meaning the root hash no longer applies to the associated proof that was just generated.

A work-around for this is 

1. Get the epoch
2. Generate the proof
3. Get the root hash, using `Directory::get_root_hash_safe` which will err if the epoch changed
4. Try again

But ideally we'd avoid this by having the proof generations just return the `EpochHash` as an additional result.

slawlor:
1. Moving the `akd::proto` module to a new location `akd::local_auditing` so that it doesn't mask `akd_core::proto` which is re-exported by `akd`.
2. Resolve #285 by returning an epoch hash with each proof generation request. This assures that proofs generated are on the same epoch as the resolved hash, minimizing race conditions. See Issue FMI


akd/src/lib.rs
```diff
@@ -158,16 +158,16 @@
 //! # tokio_test::block_on(async {
 //! #     let vrf = HardCodedAkdVRF{};
 //! #     let mut akd = Directory::<_, _>::new(&storage_manager, &vrf, false).await.unwrap();
-//! #     let EpochHash(epoch, root_hash) = akd.publish(entries)
+//! #     let _ = akd.publish(entries)
 //! #         .await.expect("Error with publishing");
-//! #     let lookup_proof = akd.lookup(
+//! #     let (lookup_proof, root_hash) = akd.lookup(
```
afterdusk, line 163: Nit: It might be clearer to call this `epoch_hash`

akd/src/tests.rs
```diff
@@ -108,16 +108,13 @@ async fn test_simple_lookup() -> Result<(), AkdError> {
...
-    let current_azks = akd.retrieve_current_azks().await?;
-    let root_hash = akd.get_root_hash(&current_azks).await?;
+    let (lookup_proof, root_hash) = akd.lookup(AkdLabel::from_utf8_str("hello")).await?;
     // Get the VRF public key
     let vrf_pk = akd.get_public_key().await?;
     // Verify the lookup proof
     lookup_verify(
         vrf_pk.as_bytes(),
-        root_hash,
+        root_hash.1,
```
afterdusk, line 117: Nit: `root_hash.hash()`, as well as maybe naming the variable `epoch_hash`

A few more instances of `root_hash.0` and `root_hash.1` in this file that can also be updated :)

poc/src/directory_host.rs
```diff
@@ -114,32 +114,21 @@ where
             }
             (DirectoryCommand::Lookup(a), Some(response)) => {
                 match directory.lookup(AkdLabel::from_utf8_str(&a)).await {
-                    Ok(proof) => {
+                    Ok((proof, root_hash)) => {
                         let hash = get_root_hash::<_, V>(directory).await;
```
afterdusk, line 118: I think this can be removed

afterdusk (approved):
Thanks for updating this so quickly!

PR #288, opened 2022-12-11
General cleanups and renamings

kevinlewi:
- Changed the name of the field `least_descendant_ep` to `min_descendant_epoch`
- Removed some `debug!(...)` statements
- Eliminated some unnecessary functions that weren't being used
- Adjusted visibility for some functions only used internally

### DIFF
--- a/akd/src/tree_node.rs
+++ b/akd/src/tree_node.rs
@@ -16,14 +16,12 @@ use crate::{node_label::*, Direction, EMPTY_LABEL};
 #[cfg(feature = "serde_serialization")]
 use akd_core::utils::serde_helpers::{bytes_deserialize_hex, bytes_serialize_hex};
 use async_recursion::async_recursion;
-use log::debug;
 use std::cmp::min;
 use std::convert::TryInto;
 use std::marker::{Send, Sync};
 
 /// There are three types of nodes: root, leaf and interior.
-/// This enum is used to mark nodes using the node_type variable
-/// of a TreeNode.
+/// This enum is used to mark the type of a [TreeNode].
 #[derive(Eq, PartialEq, Debug, Copy, Clone, Hash)]
 #[cfg_attr(
     feature = "serde_serialization",
@@ -31,10 +29,10 @@ use std::marker::{Send, Sync};
 )]
 pub enum NodeType {
     /// Nodes with this type only have dummy children. Their value is
-    /// input when they're created and the hash is H(value, creation_epoch)
+    /// an input when they're created and the hash is H(value, creation_epoch)
     Leaf = 1,
     /// Nodes with this type do not have parents and their value,
-    /// like Interior below, is a hash of their children's
+    /// like Interior, is a hash of their children's
     /// hash along with their respective labels.
     Root = 2,
     /// Nodes of this type must have non-dummy children
@@ -59,9 +57,7 @@ impl NodeType {
     }
 }
 
-pub(crate) type InsertionNode<'a> = (Direction, &'a mut TreeNode);
-
-/// Represents a `TreeNode` with its current state and potential future state.
+/// Represents a [TreeNode] with its current state and potential future state.
 /// Depending on the `epoch` which the Directory believes is the "most current"
 /// version, we may need to load a slightly older version of the tree node. This is because
 /// we can't guarantee that a "publish" operation is globally atomic at the storage layer,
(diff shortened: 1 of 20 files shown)


akd/src/tree_node.rs
```diff
@@ -16,25 +16,23 @@ use crate::{node_label::*, Direction, EMPTY_LABEL};
...
 use async_recursion::async_recursion;
-use log::debug;
 use std::cmp::min;
 use std::convert::TryInto;
 use std::marker::{Send, Sync};
 
 /// There are three types of nodes: root, leaf and interior.
-/// This enum is used to mark nodes using the node_type variable
-/// of a TreeNode.
+/// This enum is used to the type of a [TreeNode].
```
slawlor, line 24: nit: typo, I'm guessing

```rust
/// This enum is used to determine the type of a [TreeNode]
```
?

akd/src/tree_node.rs
```diff
@@ -884,13 +854,13 @@ pub(crate) fn optional_child_state_hash(input: &Option<TreeNode>) -> Digest {
...
 /// Create an empty root node.
-pub async fn create_empty_root<S: Database + Sync + Send>(
+pub(crate) async fn create_empty_root<S: Database + Sync + Send>(
     storage: &StorageManager<S>,
     ep: Option<u64>,
-    least_descendant_ep: Option<u64>,
+    smallest_descendant_ep: Option<u64>,
 ) -> Result<TreeNode, StorageError> {
-    // Empty root hash is the same as empty node hash
-    let empty_root_hash = crate::utils::empty_node_hash_no_label();
```
slawlor, line 893: nit: Can we leave this as a ref into this own crate? Example

```rust
crate::hash::hash(&crate::EMPTY_VALUE)
```

That way we don't have explicit calls into `akd_core` which imo is cleaner. If you see any other fully referenced calls to `akd_core` you could definitely relate them to `crate::` prefixes! We should have re-imported the entire core lib. 

The reason: If anyone looks at the `akd` crate source, or say examples, comments, if everything is `crate::` referenced they know they don't need to explicitly pull sub-dependencies (like `akd_core`) which helps mitigate versioning compat problems downstream.

akd/src/utils.rs
```diff
@@ -45,10 +45,6 @@ pub(crate) fn empty_node_hash() -> crate::Digest {
     akd_core::hash::merge(&[akd_core::hash::hash(&EMPTY_VALUE), EMPTY_LABEL.hash()])
```
slawlor, line 45: nit: Like above, it looks like we could do the same here with `crate::hash...` instead of `akd_core::hash...`

akd_mysql/src/mysql_storables.rs
```diff
@@ -138,7 +138,7 @@ impl MySqlStorable for DbRecord {
                 "hash" => node.latest_node.hash,
                 // "Previous" node values
                 "p_last_epoch" => node.previous_node.clone().map(|a| a.last_epoch),
-                "p_least_descendant_ep" => node.previous_node.clone().map(|a| a.least_descendant_ep),
+                "p_least_descendant_ep" => node.previous_node.clone().map(|a| a.smallest_descendant_ep),
```
slawlor, line 141: Ah OK so this is "technically" a contract break, since it's something MySQL sees. Therefore can you bump the minimum version of `akd` that `akd_mysql` depends on? Additionally we should version-bump all crates with this, since it's not backwards compatible.

slawlor (changes requested):
One important change, but otherwise it looks good!

akd/src/tree_node.rs
kevinlewi: Whoops, thanks!

akd/src/tree_node.rs
kevinlewi: Sounds good. Replaced a bunch of instances in `akd` from `akd_core::hash` to `crate::hash`

akd_mysql/src/mysql_storables.rs
kevinlewi: Bumped everything to 0.8.1

kevinlewi:
Did a CTRL-F in `akd/` for for `akd_core::hash` and replaced with `crate::hash`. Also renamed the variable to `min_descendant_epoch`

slawlor (approved):
Looks good, just need a rebase on `main` I think

kevinlewi:
Thanks, will merge!

PR #289, opened 2022-12-11
Updating VRF label derivation, commitment generation, and switching to big-endian

kevinlewi:
**Warning: This PR will break existing stored hashes and data because it changes the hashing mechanism in multiple ways!**

To convert a regular label (arbitrary string of bytes) into a [NodeLabel], we compute the output as: H(label || stale || version)

Specifically, we concatenate the following together:
- I2OSP(len(label) as u64, label)
- A single byte encoded as 0u8 if "stale", 1u8 if "fresh"
- A u64 representing the version

This saves on the number of hash computations we do (previously, 3).

In addition, I am adjusting the serialization of things to use big-endian instead of little-endian. This is mainly because the I2OSP serialization follows network byte order and does big-endian, and I'm hoping to keep endianness conversions consistent across the akd library. But let me know if there are reasons why we should avoid big-endian and would be happy to discuss...

Finally, this PR will also add back in the incorporating of the "version" to the commitment generated by `commit_value`. This was previously removed, perhaps accidentally, but should be re-added.

### DIFF
--- a/akd_core/src/ecvrf/traits.rs
+++ b/akd_core/src/ecvrf/traits.rs
@@ -49,41 +49,34 @@ pub trait VRFKeyStorage: Clone + Sync + Send {
         self.get_vrf_private_key().await.map(|key| (&key).into())
     }
 
-    /// Returns the tree nodelabel that corresponds to a version of the uname argument.
-    /// The stale boolean here is to indicate whether we are getting the nodelabel for a fresh version,
+    /// Returns the [NodeLabel] that corresponds to a version of the label argument.
+    /// The stale boolean here is to indicate whether we are getting the [NodeLabel] for a fresh version,
     /// or a version that we are retiring.
     async fn get_node_label(
         &self,
-        uname: &AkdLabel,
+        label: &AkdLabel,
         stale: bool,
         version: u64,
     ) -> Result<NodeLabel, VrfError> {
-        let proof = self.get_label_proof(uname, stale, version).await?;
-        let output: super::ecvrf_impl::Output = (&proof).into();
-        Ok(NodeLabel::new(output.to_truncated_bytes(), 256u32))
+        let proof = self.get_label_proof(label, stale, version).await?;
+        self.get_node_label_from_vrf_proof(proof).await
     }
 
     /// Returns the tree nodelabel that corresponds to a vrf proof.
     async fn get_node_label_from_vrf_proof(&self, proof: Proof) -> Result<NodeLabel, VrfError> {
         let output: super::ecvrf_impl::Output = (&proof).into();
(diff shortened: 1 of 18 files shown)


akd_core/src/ecvrf/traits.rs
```diff
@@ -49,41 +49,35 @@ pub trait VRFKeyStorage: Clone + Sync + Send {
...
-        uname: &AkdLabel,
+        label: &AkdLabel,
         stale: bool,
         version: u64,
     ) -> Result<NodeLabel, VrfError> {
-        let proof = self.get_label_proof(uname, stale, version).await?;
+        let proof = self.get_label_proof(label, stale, version).await?;
         let output: super::ecvrf_impl::Output = (&proof).into();
-        Ok(NodeLabel::new(output.to_truncated_bytes(), 256u32))
+        Ok(NodeLabel::new(output.to_truncated_bytes(), 256))
```
slawlor, line 63: nit: Can we just call `get_node_label_from_vrf_proof` here?

akd_core/src/hash/tests.rs
```diff
@@ -112,7 +112,7 @@ fn test_merge_int_validity() {
     let random_hash = random_hash();
     let merged = merge_with_int(random_hash, random_epoch);
 
-    let data = vec![random_hash.to_vec(), random_epoch.to_le_bytes().to_vec()].concat();
+    let data = vec![random_hash.to_vec(), random_epoch.to_be_bytes().to_vec()].concat();
```
slawlor, line 115: I can't do a find-and-replace on your branch, but you're sure you got all the `le_bytes` instances? Just want to make sure there's nothing dangling left!

slawlor (approved):
Looks good! Thanks!

akd_core/src/ecvrf/traits.rs
kevinlewi: Definitely... nice!

akd_core/src/hash/tests.rs
kevinlewi: I believe so:
```
klewi@klewi-mbp akd % grep -rn 'le_bytes' */src
akd_core/src/utils.rs:75:    let stale_bytes = if stale { &[0u8] } else { &[1u8] };
akd_core/src/utils.rs:79:            stale_bytes,
akd_mysql/src/mysql_storables.rs:639:                (Some(possible_bytes), Some(possible_u32)) => {
akd_mysql/src/mysql_storables.rs:642:                        from_value_opt::<Vec<u8>>(possible_bytes),
```

kevinlewi (commented):
Addressing comments + adding some doc comments to akd_core's lib.rs to describe the updated hashing changes

PR #290, opened 2022-12-16
Removing a lot of needless debug statements

slawlor:
This PR removes a lot of `debug!("BEGIN...` and `debug!("END...` statements which required a lot of string formatting needlessly


eozturk1:
Can we add time measurements for different publishing phases as well?

eozturk1:
Instead of removing them, we may want to hide the logging behind a flag in case we need it later.

dillonrg (approved):
Looks good!

PR #292, opened 2022-12-19
Change direction from Option<usize> to u8 + Sha512/256 support

slawlor:
This PR 

1. Moves direction from an `Option<usize>` to a `u8` formatted enum. (We aren't supporting an arity >= 2 so it's unnecessary to have a larger tree size and due to the risk of encoding problems between usize on architectures, we should be hard-setting a lower-risk type.)
2. Adds sync_fork.sh to help automate fork synchronization with the GitHub CLI
3. Adds support for Sha512/256 hashing (truncated SHA512 hashes to 256 bits)

### DIFF
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -9,7 +9,7 @@
 use crate::storage::manager::StorageManager;
 use crate::storage::types::StorageType;
 use crate::{
-    errors::{AkdError, DirectoryError, TreeNodeError},
+    errors::{AkdError, DirectoryError},
     storage::{Database, Storable},
     tree_node::*,
     AppendOnlyProof, Digest, Direction, LayerProof, MembershipProof, Node, NodeLabel,
@@ -18,11 +18,11 @@ use crate::{
 
 use akd_core::SizeOf;
 use async_recursion::async_recursion;
-use log::info;
-use std::marker::{Send, Sync};
-
 use keyed_priority_queue::{Entry, KeyedPriorityQueue};
+use log::info;
 use std::collections::HashSet;
+use std::convert::TryFrom;
+use std::marker::{Send, Sync};
 
 /// The default azks key
 pub const DEFAULT_AZKS_KEY: u8 = 1u8;
@@ -190,7 +190,7 @@ impl Azks {
                 }
 
                 for dir in 0..ARITY {
-                    let child_label = node.get_child_label(Direction::Some(dir));
+                    let child_label = node.get_child_label(Direction::try_from(dir).unwrap());
 
                     if let Some(child_label) = child_label {
                         current_nodes.push(NodeKey(child_label));
@@ -319,10 +319,11 @@ impl Azks {
         let mut longest_prefix_children = [Node {
             label: EMPTY_LABEL,
             hash: crate::utils::empty_node_hash(),
-        }; ARITY];
+        }; ARITY as usize];
         for i in 0..ARITY {
+            let dir = Direction::try_from(i).unwrap();
             let child = lcp_node
-                .get_child_state(storage, Some(i), self.latest_epoch)
+                .get_child_state(storage, dir, self.latest_epoch)
                 .await?;
             match child {
                 None => {
@@ -335,7 +336,7 @@ impl Azks {
                         self.get_latest_epoch(),
                     )
                     .await?;
-                    longest_prefix_children[i] = Node {
+                    longest_prefix_children[i as usize] = Node {
                         label: unwrapped_child.label,
                         hash: optional_child_state_hash(&Some(unwrapped_child)),
                     };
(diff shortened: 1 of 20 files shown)


akd_core/src/types/mod.rs
```diff
@@ -44,11 +46,40 @@ pub trait SizeOf {
...
-pub type Direction = Option<usize>;
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
+#[cfg_attr(
+    feature = "serde_serialization",
+    derive(serde::Serialize, serde::Deserialize)
+)]
+#[repr(u8)]
+pub enum Direction {
+    /// Left
+    Left = 0u8,
```
kevinlewi, line 57: Perhaps a style thing, but we don't need to specify 0u8, 1u8, 255u8 when we have `#[repr(u8)]` right? Also, perhaps we should use u8::MAX instead of 255 everywhere, or vice versa?

akd_core/src/proto/mod.rs
```diff
@@ -169,7 +171,7 @@ impl From<&crate::LayerProof> for specs::types::LayerProof {
         Self {
             label: MessageField::some((&input.label).into()),
             siblings: input.siblings.iter().map(|s| s.into()).collect::<Vec<_>>(),
-            direction: input.direction.map(|i| i as u32),
+            direction: Some(input.direction as u32),
```
kevinlewi, line 174: Should the .proto instead use a u8 here as well, instead of an Option<u32>?

akd_core/src/lib.rs
```diff
@@ -33,4 +33,4 @@ pub use types::*;
 
 /// The arity of the tree. Should EXACTLY match the ARITY within
 /// the AKD crate (i.e. akd::ARITY)
-pub const ARITY: usize = 2;
+pub const ARITY: u8 = 2;
```
kevinlewi, line 36: We should create a new issue tracking the removal of ARITY

akd/src/tree_node.rs
```diff
@@ -694,18 +697,18 @@ impl TreeNode {
...
                 "Unexpected child index: {:?}",
                 direction
             )));
+        } else {
+            if direction == Direction::Left {
+                self.left_child = Some(child_node.label);
+            }
+            if direction == Direction::Right {
+                self.right_child = Some(child_node.label);
+            }
```
kevinlewi, line 711: Maybe for this, we can just do:
```
match direction {
  Direction::None => ...,
  Direction::Left => ...,
  Direction::Right => ...,
}
```

akd/src/append_only_zks.rs
```diff
@@ -319,10 +319,11 @@ impl Azks {
         let mut longest_prefix_children = [Node {
             label: EMPTY_LABEL,
             hash: crate::utils::empty_node_hash(),
-        }; ARITY];
+        }; ARITY as usize];
         for i in 0..ARITY {
+            let dir = Direction::try_from(i).unwrap();
```
kevinlewi, line 324: This will be an error since Direction::try_from(i).unwrap() when i=0 will yield Direction::None, whereas we mean for it to yield Direction::Left.

Perhaps we can just do:
```
for dir in [Direction::Left, Direction::Right] {
  ...
}
```
and avoid the unwrap() as well.

This doesn't respect the ARITY = 2 requirement, but... perhaps we should also get rid of ARITY as well.

Also, I'm surprised that our tests didn't catch this error. Any idea why that might be happening?

akd/src/append_only_zks.rs
```diff
@@ -190,7 +190,7 @@ impl Azks {
                 }
 
                 for dir in 0..ARITY {
-                    let child_label = node.get_child_label(Direction::Some(dir));
+                    let child_label = node.get_child_label(Direction::try_from(dir).unwrap());
```
kevinlewi, line 193: Same here:

```
for dir in [Direction::Left, Direction::Right] {
  ...
}
```

kevinlewi (changes requested):

akd_core/src/types/mod.rs
slawlor: We do, since by default it would encode 0, 1, 2. In this case, since we're going to encode over the wire I'd like it clear that "None" is really a separate value and not leave it up to the compiler to infer (perhaps incorrectly).

This just removes any risk in case someone re-orders enum fields or something. To the `u8::MAX` point, good catch will adjust!

akd_core/src/proto/mod.rs
slawlor: Protobuf doesn't support u8 unfortunately :(
https://developers.google.com/protocol-buffers/docs/proto#scalar

akd_core/src/types/mod.rs
kevinlewi: Sorry I meant, do we need to say `0u8`, `1u8`, `255u8` instead of just `0`, `1`, `255` / `u8::MAX`?

akd/src/append_only_zks.rs
slawlor: `Direction::Left` is `0u8` so it's valid for a Direction. It's why we map `Direction::None` to `u8::MAX`

akd_core/src/proto/mod.rs
kevinlewi: Maybe we should just make the Direction a native `u32`, instead of `u8`, then? Just to avoid any confusion or potential bugs when dealing with encodings/decodings.

akd/src/append_only_zks.rs
slawlor: So I've removed the loops over ARITY (in upcoming update), but maybe it's helpful to have ARITY stay around since we have some array initializers which point to that const. Even if just treated as a name alias, it removes the mystical `[...;2]` in array inits around the lib?

Anyways the direction array definitely helps clean things up!

akd_core/src/lib.rs
slawlor: I made a comment about this in another comment, since we're removing a lot of the `ARITY` based loops, it's really only used now for array initialization which might be useful just as a name alias tbh. Just something to think about!

akd/src/append_only_zks.rs
kevinlewi: Ohh, I see, good to know that it was my mistake in interpreting it as a bug when it was actually working correctly, and not a lapse in our tests, either. Thanks!

akd_core/src/types/node_label/mod.rs
```diff
@@ -230,6 +230,6 @@ impl NodeLabel {
         if other.get_prefix(self.get_len()) != *self {
             return Direction::None;
         }
-        Direction::Some(other.get_bit_at(self.get_len()) as usize)
+        Direction::try_from(other.get_bit_at(self.get_len())).unwrap()
```
kevinlewi, line 233: Filing an issue here to track and remove this .unwrap() call in a future PR

akd_core/src/lib.rs
kevinlewi: Good point!

kevinlewi (approved):
Excellent!

akd_core/src/types/node_label/mod.rs
slawlor: So I thought about it, but other.get_bit_at only returns a `0` or `1` so it should always be safe (i.e. not panic). It was either this or thread the direction error all the way around everywhere this parent function was called. But yeah not opposed to removing unwraps, just means more error bubbling when it shouldn't ever happen.

PR #295, opened 2022-12-20
Support for parallel calculations of VRF labels

slawlor:
VRFs are relatively expensive to compute especially for the directory's publish operation. They're presently done sequentially as we iterate over the data. Servers which will be processing thousands of new leaves generally will have more than 1 core available to compute VRF labels, and therefore we should be able to parallelize this operation trivially and store the result into a lookup map.

This PR adds support for parallel VRFs via a toggle-able feature flag (`parallel_vrf`) on both `akd` and `akd_core`. Disabled by default, it allows opt-in for this behaviour dependent on deployment configuration.

Additionally a benchmark on parallel VRF calculations was added. Run it with

```bash
cargo bench -p akd_core -F bench
```

Sample output
image

This benchmark is based off 1K labels needing to be computed, which we see already is a large improvement (just with a parallelism level of 4 we see a nearly 4x improvement).

NOTE: Additionally this adds another caching test which should have been added in a previous PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Installation
 Add the following line to the dependencies of your `Cargo.toml`:
 
 ```
-akd = "0.7"
+akd = "0.8"
 ```
 
 ### Minimum Supported Rust Version
--- a/akd_core/benches/parallel_vrfs.rs
+++ b/akd_core/benches/parallel_vrfs.rs
@@ -14,6 +14,7 @@ use akd_core::{ecvrf::VRFKeyStorage, AkdLabel};
 fn bench_parallel_vrfs(c: &mut Criterion) {
     // utilize all cores available
     let runtime = tokio::runtime::Builder::new_multi_thread().build().unwrap();
+    // A runtime which is capped at 4 worker threads (cores)
     let limited_runtime = tokio::runtime::Builder::new_multi_thread()
         .worker_threads(4)
         .build()
@@ -37,8 +38,7 @@ fn bench_parallel_vrfs(c: &mut Criterion) {
             for (label, stale, version) in labels.iter() {
                 akd_core::ecvrf::HardCodedAkdVRF::get_node_label_with_key(
                     &key, label, *stale, *version,
-                )
-                .unwrap();
+                );
             }
         })
     });
--- a/akd_core/benches/parallel_vrfs.rs
+++ b/akd_core/benches/parallel_vrfs.rs
@@ -1,0 +1,66 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+//
+// This source code is licensed under both the MIT license found in the
+// LICENSE-MIT file in the root directory of this source tree and the Apache
+// License, Version 2.0 found in the LICENSE-APACHE file in the root directory
+// of this source tree.
+
+//! Benchmarks for parallel vs sequential VRF calculations
+
+extern crate criterion;
+use self::criterion::*;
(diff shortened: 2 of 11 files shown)


README.md
```diff
@@ -23,7 +23,7 @@ Installation
 Add the following line to the dependencies of your `Cargo.toml`:
 
 ```
-akd = "0.7"
+akd = "0.8"
```
kevinlewi, line 26: It's really hard to remember to update this :P thanks

kevinlewi (approved):
LGTM!

akd_core/src/ecvrf/traits.rs
```diff
@@ -76,10 +97,74 @@ pub trait VRFKeyStorage: Clone + Sync + Send {
...
+        {
+            let mut join_set = tokio::task::JoinSet::new();
+            let labels_vec = labels.to_vec();
+            for (label, stale, version) in labels_vec.into_iter() {
+                let key_ref = key.clone();
+
+                let future = {
+                    async move {
+                        Self::get_node_label_with_key(&key_ref, &label, stale, version)
+                            .map(|ok_result| (label, ok_result))
```
eozturk1, line 138: So we ignore the errors here but early terminate in the version without parallel VRFs. 

This changes where we actually catch the errors in these different versions and potentially we lose the `username` information (in `directory.rs:149`) if we don't use the parallel version. Can we match the implementations?

One way is this function returns a vector and we can recover the username without the VRF label in the map. In that case, we should also check for the size of the map compared to the input to see whether this is a user-level problem or more general, i.e., we couldn't generate any VRFs for users.

eozturk1 (commented):
LGTM overall, thanks for putting this up quickly!

One comment on diverging paths for different features leading to non-matching info in error messages.

akd_core/benches/parallel_vrfs.rs
```diff
@@ -0,0 +1,66 @@
...
+        .map(|i| {
+            let name = format!("user {}", i);
+            (AkdLabel::from_utf8_str(&name), false, i as u64)
+        })
+        .collect::<Vec<_>>();
+    let labels_clone = labels.clone();
+
+    c.bench_function("Sequential VRFs", |b| {
+        b.iter(|| {
+            let key = runtime
```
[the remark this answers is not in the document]
slawlor: yeah it's part of the proper benchmark, because the parallel will do this op too. If we don't it's not a "true' comparison, but I see what you're getting at. But I think it should be measured.

akd_core/src/ecvrf/traits.rs
slawlor: Ah so we're not losing the errors, we're mapping a success to add the label, but the error still propogates. The type of an element in the join set is

```rust
impl Future<Output = Result<(AkdLabel, NodeLabel), VrfError>>
```

So once we capture the output of the future with `join_next()`, we then check if the VRF op was successful or not. However looking at it, it's an infallible operation so I'll remove a lot of the `Result<_,_>`'s. What's fallible is the retrieval of the key from storage.

slawlor: Ahh I believe I see your point actually, we're going to bubble up a vrf err but lose which username it occurred on. I think me removing bits that aren't actually fallible will help clean this up, the only VRF error that can occur in computing node labels is the retrieval of the private key, which isn't dependent on username so shouldn't need to be tied to a specific user acct.

PR #296, opened 2022-12-21
General cleanups

kevinlewi:
An assortment of renaming of variables, simplifying unwraps and complex code segments, and fixing documentation typos. Doesn't change any actual functionality.

### DIFF
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -6,6 +6,7 @@
 // of this source tree.
 
 //! An implementation of an append-only zero knowledge set
+use crate::errors::TreeNodeError;
 use crate::storage::manager::StorageManager;
 use crate::storage::types::StorageType;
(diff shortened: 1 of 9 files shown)


akd/src/append_only_zks.rs
```diff
@@ -573,6 +577,35 @@ impl Azks {
         self.latest_epoch = epoch;
     }
 
+    /// Gets the sibling node of the passed node's child in the "opposite" of the passed direction.
+    async fn get_sibling_node<S: Database + Sync + Send>(
```
slawlor, line 581: Thanks for extracting this logic! Much cleaner :)

slawlor: Now that I think of it, is it worth adding test coverage over this specifically?

slawlor (approved):
Nice work!

PR #298, opened 2022-12-22
Remove Stale Profiling Results + Images

dillonrg:
- Remove images and text files associated with past profiling experiments.


kevinlewi (approved):
Thanks!

PR #299, opened 2022-12-22
Minimize over-the-wire encoding in protobuf formats.

slawlor:
This change removes trailing 0's from the node labels as they go over the wire, since we know the length will ALWAYS be 32 bytes. Therefore if there's a bunch of trailing 0's in the label, we can safely trim them off keeping only the most significant data (this is common with nodes higher up in the tree).

When decoding, we see how many bytes are there and put them in the leading spaces, in-order, with a simple slice copy.

Unit test coverage added in `akd_core/src/proto/tests.rs`

### DIFF
--- a/akd_core/src/proto/mod.rs
+++ b/akd_core/src/proto/mod.rs
@@ -99,11 +99,33 @@ macro_rules! convert_from_vector {
 // NodeLabel
 // ==============================================================
 
+fn minimize_label_bytes(v: &[u8; 32]) -> Vec<u8> {
+    let mut i = 31;
+    for byte in v.iter().rev() {
+        if *byte != 0u8 {
+            break;
+        }
+
+        if i > 0 {
+            i -= 1;
+        } else {
+            return Vec::new();
+        }
+    }
+    v[0..=i].to_vec()
+}
+
+fn parse_min_label(v: &[u8]) -> [u8; 32] {
+    let mut out = [0u8; 32];
+    out[..v.len()].copy_from_slice(v);
+    out
+}
+
 impl From<&crate::NodeLabel> for specs::types::NodeLabel {
     fn from(input: &crate::NodeLabel) -> Self {
         Self {
             label_len: Some(input.label_len),
-            label_val: Some(input.label_val.to_vec()),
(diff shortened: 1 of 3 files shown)


akd_core/src/proto/mod.rs
```diff
@@ -99,11 +99,33 @@ macro_rules! convert_from_vector {
...
+    out[..v.len()].copy_from_slice(v);
+    out
+}
+
 impl From<&crate::NodeLabel> for specs::types::NodeLabel {
     fn from(input: &crate::NodeLabel) -> Self {
         Self {
             label_len: Some(input.label_len),
-            label_val: Some(input.label_val.to_vec()),
+            label_val: Some(minimize_label_bytes(&input.label_val)),
```
kevinlewi, line 128: Rather than iterating and searching for the the beginning of the trailing zeros, we can just use the label_len parameter here. In other words, just take the slice of `&input.label_val[..input.label_len]` and that should work!

akd_core/src/proto/tests.rs
```diff
@@ -278,3 +278,33 @@ fn test_convert_single_append_only_proof() {
     let protobuf: SingleAppendOnlyProof = (&original).into();
     assert_eq!(original, (&protobuf).try_into().unwrap());
 }
+
+#[test]
+fn test_minimum_encoding_label_bytes() {
```
kevinlewi, line 283: Tests might just set an actual NodeLabel with label_val and label_len and make sure it is truncated properly

kevinlewi (changes requested):
Awesome! Just a comment on how the minimization is done

akd_core/src/proto/tests.rs
slawlor: Spoke out-of-band, since label's might allow for additional 0's added, this optimizes for wire-size above all else (nodes in the very left of the tree will have trailing 0's, and the length will point beyond the last sig byte). This "search" is a small computation to minimize the wire-size overall.

Tests with node-label will have no effect then, since we don't utilize the length anyways.

kevinlewi (approved):
Great!

PR #300, opened 2022-12-22
Fix incorrect conditional check in common prefix function

afterdusk:


eozturk1 (commented):
Can we add a test with previously failing values?

slawlor:
Great catch, also want to have @kevinlewi look this over!

kevinlewi:
Amazing.

afterdusk:
Added tests that would have failed prior to this fix

slawlor (approved):
LGTM!

eozturk1 (approved):

PR #301, opened 2022-12-23
Make leaf insertion and hashing recursive

afterdusk:
# Overview
To parallelize leaf insertion and hashing, the process needs to be made fully recursive, such that work can be split into subtasks and performed independently of one another. As an additional plus point, a fully recursive approach is more concise and easier to reason about. 

To this end, this PR does the necessary refactor. I do this by partitioning the set of leaves to be inserted based on the current node, then delegating the partitions to the left and right child nodes to handle recursively. When both child nodes return, the current node is hashed and returned to its parent.

The actual parallelization comes in #302, for ease of review.

# Benchmark

**Main**
Cherry-picked the updated benchmark (6c9f83e, 3ade785) on top of main (see: https://github.com/afterdusk/akd/tree/benchmark) to compare this PR against.

image

**This PR**
image

# Progress
- [x] Rewrite leaf insertion and hashing to be recursive
- [x] Rebase on main branch
- [x] Consider porting some tests from `tree_node.rs` to `append_only_zks.rs`, as tree traversal logic was largely abstracted out of `tree_node.rs` in this PR (will relook in a separate PR)
- [x] Benchmark


integration_tests/src/test_util.rs
```diff
@@ -120,132 +120,6 @@ impl log::Log for FileLogger {
 
 // ================== Test Helpers ================== //
 
-// FIXME: We actually probably want to use this. Figure out later.
-/// The suite of tests to run against a fully-instantated and storage-backed directory.
-/// This will publish 3 epochs of ```num_users``` records and
-/// perform 10 random lookup proofs + 2 random history proofs + and audit proof from epochs 1u64 -> 2u64
-#[allow(unused)]
-pub(crate) async fn directory_test_suite<
```
afterdusk, line 128: This test is already present in `akd_test_tools` and is used from there.

https://github.com/novifinancial/akd/blob/9e2ef7e2518da9b16159351e911c8438b3d0ce00/integration_tests/src/memory_tests.rs#L23

https://github.com/novifinancial/akd/blob/9e2ef7e2518da9b16159351e911c8438b3d0ce00/integration_tests/src/mysql_tests.rs#L51

akd/src/append_only_zks.rs
```diff
@@ -199,88 +206,108 @@ impl Azks {
...
+        let mut current_node;
+        let mut num_inserted;
+
+        match (node_label, &insertion_set[..]) {
+            (Some(node_label), _) => {
+                // Case 1: The node label is not None, meaning that there was an
+                // existing node at this level of the tree.
+                current_node =
+                    TreeNode::get_from_storage(storage, &NodeKey(node_label), epoch).await?;
+                num_inserted = 0;
```
slawlor, line 229: nit: just for readability, can we make this part of the else statement below? Then the match can be and assignment to num_inserted's initial value. Ex

```rust
let mut num_inserted = match (node_label, &insertion_set[..]) {
    (Some(...), ) => {

        if (blah) {
            1
        } else {
            0
        }
    }
    ...
}
```

akd/src/append_only_zks.rs
```diff
@@ -199,88 +206,108 @@ impl Azks {
...
-    /// However, both constructions have very minor differences, and the append_only_usage
-    /// bool keeps track of this.
-    pub async fn batch_insert_leaves_helper<S: Database + Sync + Send>(
-        &mut self,
+    /// Inserts a batch of leaves recursively from a given node label.
+    #[async_recursion]
+    pub async fn recursive_batch_insert_leaves<S: Database + Sync + Send>(
         storage: &StorageManager<S>,
+        node_label: Option<NodeLabel>,
         insertion_set: Vec<Node>,
```
slawlor, line 214: nit: can we make this take a `&[Node]` instead of a Vec? That way we don't need to create vectors eveywhere and probably can help cut-down on cloning.

akd/src/append_only_zks.rs
```diff
@@ -199,88 +206,108 @@ impl Azks {
...
+        epoch: u64,
         append_only_exclude_usage: bool,
```
slawlor, line 216: nit: I'm going to suggest a @dillonrg suggestion here, can we make this something other than a bool? Like 

```rust

enum AzksInsertUsage {
    Auditor,
    Directory
}
```

akd/src/tree_node.rs
```diff
@@ -810,6 +546,30 @@ impl TreeNode {
 
 /////// Helpers //////
 
+pub(crate) fn get_longest_common_prefix(nodes: &[akd_core::Node]) -> NodeLabel {
```
slawlor, line 549: nit, we should be able to write this as 

```rust
nodes: &[crate::Node]
```

without directly referencing `akd_core` due to the re-exports of types. This will be clearer for devs of this specific crate.

akd/src/tree_node.rs
```diff
@@ -810,6 +546,30 @@ impl TreeNode {
...
+    let mut prefix = nodes[0].label;
```
slawlor, line 550: nit: can we use a fold here?

```rust
nodes
  .iter()
  .skip(1)
  .fold(nodes[0].label, |acc, node| node.label.get_longest_common_prefix(acc))
```

akd/src/tree_node.rs
```diff
@@ -810,6 +546,30 @@ impl TreeNode {
...
+        prefix = node.label.get_longest_common_prefix(prefix);
+    }
+    prefix
+}
+
+pub(crate) fn partition_longest_common_prefix(
+    nodes: Vec<akd_core::Node>,
+    lcp_label: NodeLabel,
+) -> (Vec<akd_core::Node>, Vec<akd_core::Node>) {
+    let mut left = vec![];
```
slawlor, line 561: nit: same comment, can't we do this by folding to be cleaner?

```rust
nodes
  .into_iter()
  .fold((vec![], vec![]), |mut (left, right), node| {
        match lcp_label.get_dir(node.label) {
            Direction::Left => left.push(node),
            Direction::Right => right.push(node),
            Direction::None => (),
        }
  })
```

slawlor: Not sure I got the syntax 100% right here, but that's the idea anyways

slawlor (changes requested):
Some styling things, but I think it'll help with readability. However very awesome work and thanks for the great effort!

akd/src/append_only_zks.rs
afterdusk: Ah I think it actually makes sense for the `insertion_set` to be a Vec here. We're handing over ownership of the partitions to the recursive calls, so we're never cloning the data. If we were to make it a reference, we would have to ensure that the original Vec lives long enough until the recursive call returns, which is particularly inconvenient if the recursive call runs in a separate task.

akd/src/append_only_zks.rs
afterdusk: Addressed in #308, will rebase when it's merged 🙂

akd/src/append_only_zks.rs
afterdusk: I've refactored this section and added an `else` block, hopefully it's clearer! I opted to not have the match expression return anything because the `current_node` variable would have to be assigned as well, and I think it's a bit neater to assign them together inside the match arms.

afterdusk:
Rebased on main and addressed review comments. Will work on benchmarking this PR separately from #302, as #302 is currently blocked on some perf issues with parallelization. If this PR does not impact perf negatively, I would like to merge it first.

kevinlewi:
Really loving how clean this is!

Two places where we might be able to improve on perf:
1. When you are getting the longest prefix of all nodes in the insertion set, instead of iterating through the insertion set and calling longest_common_prefix on each element, we might be able to make use of something like this function (https://github.com/novifinancial/akd/blob/main/akd/src/append_only_zks.rs#L165-L175). Basically, start by assuming that the insertion set is a sorted vector of NodeLabels, sorted by their label. Then, start with a candidate prefix derived from the first element of the sorted vec, and check if it is a prefix of all of the others via binary search. Keep going until you find the point at which you have the "longest" prefix (also could be found via binary search on the bits of the prefix).
2. Kindof related to #1, but right now when you compute `partition_longest_common_prefix`, you are iterating over the insertion set to compute the partition. However, if the insertion set is already sorted, you would just need to find the "midpoint" (again via binary search), and then break up the vec into two partitions immediately.

Hope that makes sense! But anyways let's land this first and worry about these perf optimizations later. The structure of this is really nice.

kevinlewi (approved):

afterdusk:
@kevinlewi Thanks for the suggestions, they make sense to me! I'll work on them and should have a PR up by this week :)

kevinlewi:
@afterdusk Btw I did a checkout of this PR and ran it locally just to see what the perf change is right now,

Main branch (without this PR):
```
Batch insertion (100000 initial leaves) took 9666 ms
Batch insertion (100000 initial leaves, 100000 inserted leaves) took 11955 ms
```
This branch (with this PR):
```
Batch insertion (100000 initial leaves) took 7552 ms
Batch insertion (100000 initial leaves, 100000 inserted leaves) took 8571 ms
```

Already a 20-30% speedup, it seems...!

slawlor (approved):
Looks good to merge to me! I agree with @kevinlewi that a binary search on the nodes vec if it's already sorted will give us another perf win!

PR #302, opened 2022-12-24
Parallelize leaf insertion and hashing

afterdusk:
# Overview
This PR builds upon #301 to finally parallelize leaf insertion and hashing. A tokio task is spawned for every recursive call on the left child, while the right child's work is executed in the current task. 

While tokio tasks are [lightweight](https://docs.rs/tokio/latest/tokio/task/#what-are-tasks), benchmarking has revealed that the best performance is had when the number of spawned tasks approximately equals the number of available CPU cores. To restrict the number of tasks spawned, I get the available number of threads and calculate the level by which we have to stop spawning new tasks. This is a rough measure that might not yield the ideal parallelism in all circumstances, but will be reasonably close to it. 

# Benchmark
Ran the `azks.rs` benchmark, which inserts 100K leaves into a tree with 10K leaves.

## Sequential
```console
$ cargo bench -p akd --bench azks --no-default-features -F bench
```
image

## Parallel
```console
$ cargo bench -p akd --bench azks --no-default-features -F bench -F parallel_insert
```
image

A whopping 80% improvement in execution time 🏎💨

# Progress
- [x] Parallelize recursive insert
- [x] Gate parallelism behind feature flag
- [x] Convert transaction to use dashmap
- [x] Restrict tasks spawned
- [x] Benchmark 
- [x] Update GH actions to run tests without feature flag

### DIFF
--- a/akd/benches/azks.rs
+++ b/akd/benches/azks.rs
@@ -55,6 +55,9 @@ fn batch_insertion(c: &mut Criterion) {
                 let db = StorageManager::new_no_cache(&database);
                 let mut azks = runtime.block_on(Azks::new(&db)).unwrap();
 
+                // create transaction object
+                db.begin_transaction();
+
                 // insert initial leaves as part of setup
                 runtime
                     .block_on(azks.batch_insert_nodes(
--- a/akd/benches/azks.rs
+++ b/akd/benches/azks.rs
@@ -52,7 +52,7 @@ fn batch_insertion(c: &mut Criterion) {
         b.iter_batched(
             || {
                 let database = AsyncInMemoryDatabase::new();
-                let db = StorageManager::new_no_cache(&database);
+                let db = StorageManager::new(database, None, None, None);
                 let mut azks = runtime.block_on(Azks::new(&db)).unwrap();
 
                 // create transaction object
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -6,7 +6,6 @@
 // of this source tree.
 
 //! An implementation of an append-only zero knowledge set
-#[cfg(feature = "parallel_insert")]
 use crate::errors::ParallelismError;
 use crate::errors::TreeNodeError;
 use crate::helper_structs::LookupInfo;
@@ -31,6 +30,11 @@ use std::ops::Deref;
 /// The default azks key
 pub const DEFAULT_AZKS_KEY: u8 = 1u8;
 
+/// The default available parallelism for parallel batch insertions, used when
+/// available parallelism cannot be determined at runtime. Should be > 1
+#[cfg(feature = "parallel_insert")]
+pub const DEFAULT_AVAILABLE_PARALLELISM: usize = 32;
+
 async fn tic_toc<T>(f: impl core::future::Future<Output = T>) -> (T, Option<f64>) {
     #[cfg(feature = "runtime_metrics")]
     {
@@ -41,6 +45,26 @@ async fn tic_toc<T>(f: impl core::future::Future<Output = T>) -> (T, Option<f64>
     }
     #[cfg(not(feature = "runtime_metrics"))]
     (f.await, None)
+}
+
+fn get_parallel_levels() -> Option<u8> {
+    #[cfg(not(feature = "parallel_insert"))]
+    None;
+
+    #[cfg(feature = "parallel_insert")]
+    {
+        // Based on profiling results, the best performance is achieved when the
+        // number of spawned tasks is equal to the number of available cores.
+        // We therefore get the number of available threads and calculate the
+        // number of levels that should be executed in parallel to give the
+        // number of tasks closest to the number of threads. While there might
+        // be other tasks that are running on the threads, this is a reasonable
+        // approximation that should yield good performance in most cases.
+        let available_parallelism = std::thread::available_parallelism()
+            .map_or(DEFAULT_AVAILABLE_PARALLELISM, |v| v.into());
+        let levels = (available_parallelism as f32).log2().ceil() as u8;
+        Some(levels)
+    }
 }
 
 /// An azks is built both by the [crate::directory::Directory] and the auditor.
@@ -272,6 +296,7 @@ impl Azks {
                 node_set,
                 self.latest_epoch,
                 insert_mode,
+                get_parallel_levels(),
             )
             .await?;
 
@@ -292,6 +317,7 @@ impl Azks {
         node_set: NodeSet,
         epoch: u64,
         insert_mode: InsertMode,
+        parallel_levels: Option<u8>,
     ) -> Result<(TreeNode, u64), AkdError> {
         // Phase 1: Obtain the current root node of this subtree and count if a
         // node is inserted.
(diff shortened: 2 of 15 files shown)


slawlor:
Should be good to rebase. It'll help see what the changes in this specific PR are

afterdusk:
@slawlor Rebase is done!

akd/src/append_only_zks.rs
```diff
@@ -111,7 +113,7 @@ impl Azks {
     /// An azks is built both by the [crate::directory::Directory] and the
     /// auditor. However, both constructions have very minor differences, and
     /// the append_only_exclude_usage bool keeps track of this.
-    pub async fn batch_insert_leaves<S: Database + Sync + Send>(
+    pub async fn batch_insert_leaves<S: Database + Sync + Send + 'static>(
```
slawlor, line 116: Let's take the opportunity to add derivations to the base Database trait. The rewriting everywhere of sync + send is just getting silly lol. But maybe it makes sense to leave 'static here so the lifetime is only usage based.

akd/src/append_only_zks.rs
```diff
@@ -326,10 +328,32 @@ impl Azks {
         let (left_insertion_set, right_insertion_set) =
             partition_longest_common_prefix(insertion_set, current_node.label);
 
+        // if parallel: spawn a tokio task for the left child
+        #[cfg(feature = "parallel_insert")]
+        let maybe_future = if !left_insertion_set.is_empty() {
+            let storage_clone = storage.clone();
+            let left_child_label = current_node.get_child_label(Direction::Left)?;
+            Some(tokio::task::spawn(async move {
```
slawlor, line 336: Can we try vlocking_spawn and limit the concurrency level to say a few levels of the tree? say like 3 levels? That would limit to 7 dedicated threads each with a decent amount of work

akd/src/storage/transaction.rs
```diff
@@ -13,25 +13,23 @@ use crate::storage::types::ValueState;
...
-
 /// Represents an in-memory transaction, keeping a mutable state
 /// of the changes. When you "commit" this transaction, you return the
 /// collection of values which need to be written to the storage layer
 /// including all mutations. Rollback simply empties the transaction state.
+#[derive(Clone)]
 pub struct Transaction {
-    state: RwLock<TransactionState>,
+    mods: Arc<DashMap<Vec<u8>, DbRecord>>,
+    active: Arc<RwLock<bool>>,
```
slawlor, line 29: Ah maybe the use of a rwlock here is hurting us in parallel mode. Try an atomic bool instead! You could also convert the counters to atomic u64s

afterdusk: Ah I gave it a shot, it only improves execution time marginally, parallel execution is still slower than sequential :( Although since atomic primitives are faster than RwLocks in all circumstances, I'll refactor the RwLocks to atomic primitives either ways!

akd/src/storage/transaction.rs
```diff
@@ -87,99 +82,88 @@ impl Transaction {
 
     /// Start a transaction in the storage layer
     pub async fn begin_transaction(&self) -> bool {
```
slawlor, line 84: can we now refactor out the async signature too since we're using atomics? it'll remove the `Poll` in tokio since there's no async work iirc

akd/src/storage/transaction.rs
```diff
@@ -53,26 +51,23 @@ impl Transaction {
...
-        let lock = self.state.read().await;
-        lock.mods.len()
+        self.mods.len()
     }
 
     /// Log metrics about the current transaction instance. Metrics will be cleared after log call
     pub async fn log_metrics(&self, level: log::Level) {
-        let r = crate::utils::rwlock_swap(&self.num_reads, 0).await;
-        let w = crate::utils::rwlock_swap(&self.num_writes, 0).await;
+        let r = self.num_reads.swap(0, Ordering::Relaxed);
```
slawlor, line 69: Same here, removing the async signature if there's no async work

akd/src/storage/transaction.rs
```diff
@@ -13,25 +13,23 @@ use crate::storage::types::ValueState;
...
+use std::sync::Arc;
 
 /// Represents an in-memory transaction, keeping a mutable state
 /// of the changes. When you "commit" this transaction, you return the
 /// collection of values which need to be written to the storage layer
 /// including all mutations. Rollback simply empties the transaction state.
+#[derive(Clone)]
 pub struct Transaction {
-    state: RwLock<TransactionState>,
+    mods: Arc<DashMap<Vec<u8>, DbRecord>>,
```
slawlor, line 28: Actually looking at it, nothing in this impl should be async anymore. Which is awesome! It'll save a lot of polling all around the place, it's now all sync work

afterdusk: That's a good point, I've pushed the change! I've also made the calling functions sync if they no longer need to be async.

akd/src/append_only_zks.rs
afterdusk: #321 🙂

afterdusk:
Thank you so much @slawlor for unblocking this PR by figuring out the reason why the benchmark was making parallel execution appear much slower than sequential!

akd/benches/azks.rs
```diff
@@ -55,6 +55,9 @@ fn batch_insertion(c: &mut Criterion) {
                 let db = StorageManager::new_no_cache(&database);
                 let mut azks = runtime.block_on(Azks::new(&db)).unwrap();
 
+                // create transaction object
```
slawlor, line 58: so one consideration here, is we are never "committing" the transaction, i.e. paying the price to write it to the underlying DB. This means all the data will effectively be dropped, but imho this is fine for this benchmark since we are not wanting to compare the db contention time for reads/writes, but rather the amount of time spent hashing + interacting in parallel with the transaction struct.

akd/src/append_only_zks.rs
```diff
@@ -41,6 +47,26 @@ async fn tic_toc<T>(f: impl core::future::Future<Output = T>) -> (T, Option<f64>
...
+    #[cfg(feature = "parallel_insert")]
+    {
+        // Based on profiling results, the best performance is achieved when the
+        // number of spawned tasks is equal to the number of available cores.
+        // We therefore get the number of available threads and calculate the
+        // number of levels that should be executed in parallel to give the
+        // number of tasks closest to the number of threads. While there might
+        // be other tasks that are running on the threads, this is a reasonable
+        // approximation that should yield good performance in most cases.
+        let available_parallelism = std::thread::available_parallelism()
```
slawlor, line 63: is there a risk of getting this from std and not through some tokio-specific call? It seems strange to use a default parallelism crate and not tokio.

It seems like there might be a tokio specific way to do this?
https://docs.rs/tokio/latest/tokio/runtime/struct.RuntimeMetrics.html

akd/src/append_only_zks.rs
```diff
@@ -41,6 +47,26 @@ async fn tic_toc<T>(f: impl core::future::Future<Output = T>) -> (T, Option<f64>
...
+            .map_or(DEFAULT_AVAILABLE_PARALLELISM, |v| v.into());
+        let levels = (available_parallelism as f32).log2().ceil() as u8;
```
slawlor, line 65: nit: a not about why a log_2 here would be helpful (something like because it's a binary tree and at each level we spawn 2 threads / parent node

akd/src/append_only_zks.rs
afterdusk: We are getting from std here, as it turns out we need tokio_unstable to use the linked API :/ There's no tokio-specific alternative AFAIK.

afterdusk: i.e. `std::thread::available_parallelism()` is an approximation of the number of CPU cores, which is an approximation of the number of tokio threads. According to the [documentation](https://doc.rust-lang.org/std/thread/fn.available_parallelism.html) available parallelism takes into account CPU cores that have been reserved by the VM/host, which sounds perfect for our use-case.

slawlor (approved):
Looks good, great work!

PR #303, opened 2022-12-26
Minor cleanups

kevinlewi:
- Added a missing assert_eq() in `test_get_sibling_node` that was missed in #296 
- Adding some timing information to directory publish to better debug what the bottlenecks are
- Added a batch insertion test inspired from #302 which we can run and trigger printlns by doing something like: `cargo test -- test_random_batch_insertion --nocapture` to better triage directory publish performance

### DIFF
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -823,6 +823,7 @@ mod tests {
                         .await?
                         .unwrap()
                         .label;
+                    assert_eq!(right_child.label, sibling_label);
                     nodes.push(right_child);
                 }
                 None => {}
--- a/akd/src/directory.rs
+++ b/akd/src/directory.rs
@@ -93,6 +93,12 @@ impl<S: Database + Sync + Send, V: VRFKeyStorage> Directory<S, V> {
             )));
         }
 
+        info!(
+            "Starting a publish operation with {} updates",
+            updates.len()
+        );
+        let publish_start_time = tokio::time::Instant::now();
+
         // The guard will be dropped at the end of the publish
         let _guard = self.cache_lock.read().await;
 
@@ -117,9 +123,10 @@ impl<S: Database + Sync + Send, V: VRFKeyStorage> Directory<S, V> {
             .await?;
 
         info!(
-            "Retrieved {} previous user versions of {} requested",
+            "Retrieved {} previous user versions of {} requested, starting VRF computations ({} seconds elapsed)",
             all_user_versions_retrieved.len(),
-            keys.len()
+            keys.len(),
+            publish_start_time.elapsed().as_secs(),
         );
 
         let vrf_computations = updates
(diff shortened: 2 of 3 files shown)


akd/src/directory.rs
```diff
@@ -117,9 +123,10 @@ impl<S: Database + Sync + Send, V: VRFKeyStorage> Directory<S, V> {
             .await?;
 
         info!(
-            "Retrieved {} previous user versions of {} requested",
+            "Retrieved {} previous user versions of {} requested, starting VRF computations ({} seconds elapsed)",
```
eozturk1, line 126: Can we clarify in the logs elapsed time is since publish start?

akd/src/directory.rs
```diff
@@ -137,6 +144,11 @@ impl<S: Database + Sync + Send, V: VRFKeyStorage> Directory<S, V> {
             .into_iter()
             .collect::<HashMap<_, _>>();
 
+        info!(
+            "Completed VRF label computation ({} seconds elapsed)",
+            publish_start_time.elapsed().as_secs(),
```
eozturk1, line 149: Does it make sense to print the elapsed time for each phase instead of from the publish start? We wouldn't need to subtract the previous phase each time then.

akd/src/append_only_zks.rs
```diff
@@ -823,6 +823,7 @@ mod tests {
                         .await?
                         .unwrap()
                         .label;
+                    assert_eq!(right_child.label, sibling_label);
```
eozturk1, line 826: Nice catch!

eozturk1 (changes requested):
WASM compilation is failing due to missing `time` there. Shall we hide it behind a feature?

Personally I'd prefer printing the elapsed time for each phase to avoid elapsed since confusion -- not a blocker.

slawlor (approved):
I think @eozturk1 got most of the comments, but it lgtm!

slawlor:
> WASM compilation is failing due to missing `time` there. Shall we hide it behind a feature?
> 
> Personally I'd prefer printing the elapsed time for each phase to avoid elapsed since confusion -- not a blocker.

Ah it's because `akd_client` pulls in `akd` as a dependency to TEST in WASM. This actually wouldn't be a problem in prod, but only in the test profile. We could hard-code a proof into the `akd_client` tests to remove the `akd` dependency (i.e. not dynamically generate a directory + proof) as an option?

eozturk1:
> 

Yeah, that's an option. I am wondering whether we would need to hardcode/copy a lot of things in the WASM side and would it be better to just hide unimplemented sections with a flag. @kevinlewi, what do you think?

kevinlewi:
Thanks for the comments! Actually I think these changes aren't really needed anymore, especially since putting the timing info is such a pain. One solution we could do is put the time measurements behind the `runtime_metrics` feature to avoid the wasm issue.

Perhaps a note for the future if we decide to do a similar change like this in the future. But I'm closing this PR for now!

PR #306, opened 2022-12-26
Optimizing the BFS preloading prefix computation

kevinlewi:
Previously we computed the preloading of nodes by first taking the insertion set of labels, computing all prefixes, and then loading the nodes in BFS manner and checking each node's label against the prefix set. For N labels to insert, this was a set of size `256 * N`, which resulted in about `256 * N` work.

Now, we are using a BinaryHeap to insert the labels in sorted order (sorted by label value), and then instead of building a prefix set, we do a binary search to check if the current node during the BFS traversal is contained as a prefix in the sorted insertion set of labels. This is roughly `N * log(N)` work to build the sorted list, and then about `log^2(N)` work for prefix-checking during insertion. This should be much faster -- see benchmarks below.

Before this change:
Inserting 100k nodes into an empty tree:
```
[00:02:50.423] INFO   Preload of tree (1 objects loaded), took 31.261086083 s (append_only_zks:217)
```
Inserting another 100k nodes into the tree subsequently:
```
[00:06:14.654] INFO   Preload of tree (150813 objects loaded), took 34.730958 s (append_only_zks:217)
```

After this change:

Inserting 100k nodes into an empty tree:
```
[00:02:18.657] INFO   Preload of tree (1 objects loaded), took 0.005568666 s (append_only_zks:265)
```
Inserting another 100k nodes into the tree subsequently:
```
[00:05:06.851] INFO   Preload of tree (151035 objects loaded), took 1.7127710409999999 s (append_only_zks:265)
```

-----------

**Edit**: Those above numbers are run with `--profile = dev` mode. In `--profile = release` mode, the numbers are less drastic but about the same in terms of scaling:

Before:
```
[00:00:21.253] INFO   Preload of tree (1 objects loaded), took 4.403417458 s (append_only_zks:217)
Batch insertion (100000 initial leaves) took 22996 ms

[00:00:45.232] INFO   Preload of tree (150755 objects loaded), took 4.546898041 s (append_only_zks:217)
Batch insertion (100000 initial leaves, 100000 inserted leaves) took 25390 ms
```

After:
```
[00:00:16.759] INFO   Preload of tree (1 objects loaded), took 0.000585958 s (append_only_zks:265)
Batch insertion (100000 initial leaves) took 17196 ms

[00:00:34.034] INFO   Preload of tree (151191 objects loaded), took 0.199543458 s (append_only_zks:265)
Batch insertion (100000 initial leaves, 100000 inserted leaves) took 19388 ms
```

### DIFF
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -21,6 +21,7 @@ use akd_core::SizeOf;
 use async_recursion::async_recursion;
 use keyed_priority_queue::{Entry, KeyedPriorityQueue};
 use log::info;
+use std::cmp::Ordering;
 use std::collections::HashSet;
 use std::marker::{Send, Sync};
 
@@ -152,15 +153,62 @@ impl Azks {
         storage: &StorageManager<S>,
         insertion_set: &[Node],
     ) -> Result<u64, AkdError> {
-        let prefixes_set = crate::utils::build_prefixes_set(
-            insertion_set
-                .iter()
-                .map(|n| n.label)
-                .collect::<Vec<NodeLabel>>()
-                .as_ref(),
-        );
+        let insertion_labels = insertion_set
+            .iter()
+            .map(|n| n.label)
+            .collect::<Vec<NodeLabel>>();
 
-        self.bfs_preload_nodes::<S>(storage, prefixes_set).await
+        self.bfs_preload_nodes_sorted::<S>(storage, &insertion_labels)
+            .await
+    }
+
+    /// Given a sorted list of [NodeLabel]s, determine if a given [NodeLabel] is a prefix of any in the list.
+    fn insertion_set_contains(insertion_labels: &[NodeLabel], label: &NodeLabel) -> bool {
+        insertion_labels
+            .binary_search_by(|candidate| {
+                match label.label_len == 0 || label.is_prefix_of(candidate) {
+                    true => Ordering::Equal,
+                    false => candidate.label_val.cmp(&label.label_val),
+                }
+            })
+            .is_ok()
+    }
+
+    /// Preloads given nodes using breadth-first search.
+    pub async fn bfs_preload_nodes_sorted<S: Database + Sync + Send>(
+        &self,
+        storage: &StorageManager<S>,
+        insertion_labels: &[NodeLabel],
+    ) -> Result<u64, AkdError> {
+        let mut load_count: u64 = 0;
+        let mut current_nodes = vec![NodeKey(NodeLabel::root())];
+
+        while !current_nodes.is_empty() {
+            let nodes =
+                TreeNode::batch_get_from_storage(storage, &current_nodes, self.get_latest_epoch())
+                    .await?;
+            load_count += nodes.len() as u64;
+
+            current_nodes = Vec::<NodeKey>::new();
+
+            // Now that states are loaded in the cache, you can read and access them.
+            // Note, the two for loops are needed because otherwise, you'd be accessing remote storage
+            // individually for each node's state.
+            for node in &nodes {
+                if !Self::insertion_set_contains(insertion_labels, &node.label) {
+                    // Only continue to traverse nodes which are relevant prefixes to insertion_set
+                    continue;
+                }
+
+                for dir in crate::DIRECTIONS {
+                    let child_label = node.get_child_label(dir)?;
+                    if let Some(child_label) = child_label {
+                        current_nodes.push(NodeKey(child_label));
+                    }
+                }
+            }
+        }
+        Ok(load_count)
     }
 
     /// Preloads given nodes using breadth-first search.
@@ -823,6 +871,7 @@ mod tests {
                         .await?
                         .unwrap()
                         .label;
+                    assert_eq!(right_child.label, sibling_label);
                     nodes.push(right_child);
                 }
                 None => {}
--- a/akd/src/utils.rs
+++ b/akd/src/utils.rs
@@ -9,37 +9,7 @@
 // 2. For each node in current_nodes set, check if each child is in prefix hashmap
 // 3. If so, add child label to batch set
 
-use crate::{NodeLabel, EMPTY_LABEL, EMPTY_VALUE};
-use std::collections::HashSet;
-
-// Builds a set of all prefixes of the input labels
-pub(crate) fn build_prefixes_set(labels: &[NodeLabel]) -> HashSet<NodeLabel> {
-    let mut prefixes_set = HashSet::new();
-    for label in labels {
-        for len in 0..(label.get_len() + 1) {
-            prefixes_set.insert(label.get_prefix(len));
-        }
-    }
-    prefixes_set
-}
-
-pub(crate) fn build_lookup_prefixes_set(labels: &[NodeLabel]) -> HashSet<NodeLabel> {
-    let mut lookup_prefixes_set = HashSet::new();
-    for label in labels {
-        // We need the actual node for lookup too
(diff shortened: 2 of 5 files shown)


slawlor:
OK so the benchmarking here is against an in-memory DB right? So we're only measuring the computational work, not any of the actual DB timing, correct?

akd/src/append_only_zks.rs
```diff
@@ -152,15 +153,62 @@ impl Azks {
...
-                .map(|n| n.label)
-                .collect::<Vec<NodeLabel>>()
-                .as_ref(),
-        );
+        let insertion_labels = insertion_set
+            .iter()
+            .map(|n| n.label)
+            .collect::<Vec<NodeLabel>>();
+
+        self.bfs_preload_nodes_sorted::<S>(storage, &insertion_labels)
```
slawlor, line 161: q: Hmm I feel like there's got to be a way to not have to collect this set and then reference it again no?

akd/src/utils.rs
```diff
@@ -9,36 +9,42 @@
...
-            prefixes_set.insert(label.get_prefix(len));
-        }
+/// A node with node.label.label_len = 256. Used a wrapper around a [Node] in order to
+/// help with preloading nodes during directory publish.
+#[derive(Eq)]
+pub(crate) struct MaxLengthNode(Node);
+
+impl PartialEq for MaxLengthNode {
+    fn eq(&self, other: &Self) -> bool {
+        self.0.label.label_val == other.0.label.label_val
```
slawlor, line 21: nit: do we not support eq on a `NodeLabel` directly? Do we need to compare the inner values rather than just 
```rust
self.0 == other.0
```?

**Comment by kevinlewi** [comment]
Yep, that's correct, these benchmarks were against an in-memory DB.

The actual loading of nodes from the DB in the before and after should be the same -- we aren't changing anything there, since we are still loading the same # of nodes and in the same manner. The only thing that is changing is a more efficient way to compute which nodes we should be loading in.

File: akd/src/utils.rs [hunk 3]
```diff
@@ -9,36 +9,42 @@
...
-        for len in 0..(label.get_len() + 1) {
-            // Sibling prefixes unfortunately do not cover all the nodes we will need for
-            // a lookup proof. Although we can figure out which nodes are exactly needed
-            // this will require basically doing a pre-lookup.
-            // Instead here we load the prefixes as well.
-            // This combination (sibling- + self-prefixes) covers all the nodes we need.
-            lookup_prefixes_set.insert(label.get_prefix(len));
-            lookup_prefixes_set.insert(label.get_sibling_prefix(len));
-        }
+impl Ord for MaxLengthNode {
```
slawlor, line 32: If we wanted to get rid of `.into()` calls, we could support the `Deref` trait on `MaxLengthNode` to be able to silently cast it back to a Node. (Not sure it's super helpful)

slawlor: For an example, `AkdLabel` and `AkdValue` support this to get the underlying bytes without much headache

akd_core/src/types/node_label/mod.rs
```diff
@@ -87,6 +87,19 @@ impl NodeLabel {
         [&self.label_len.to_be_bytes(), &self.label_val[..]].concat()
     }
 
+    /// Outputs whether or not self is a prefix of the other [NodeLabel]
+    pub fn is_prefix_of(&self, other: &Self) -> bool {
+        if self.label_len > other.label_len {
+            return false;
+        }
+        for i in 0..self.label_len {
```
slawlor, line 95: nit: Can be replaced with a find if I'm not mistaken

```rust
(0..self.label_len).into_iter().find(|i| self.get_bit_at(i) != other.get_bit_at(i)).is_none()
```

slawlor (approved):
Sweet!

akd/src/append_only_zks.rs
kevinlewi: True, we could avoid this, but I think it comes at some additional complexity to the code. I benchmarked this operation and for inserting 100k nodes, it only takes 6 milliseconds to do this collect. So maybe not worth the optimization for now? Unless you see a simple way to address it.

akd/src/utils.rs
kevinlewi: Good point! Actually not relevant anymore since I decided to get rid of `MaxLengthNode` altogether.

akd/src/utils.rs
kevinlewi: Ah, actually, turns out I can just get rid of the `MaxLengthNode` struct altogether and re-use the Ord defined on Node / NodeLabel. Thanks!

kevinlewi (commented):
Getting rid of `MaxLengthNode`

akd/src/append_only_zks.rs
```diff
@@ -152,15 +153,62 @@ impl Azks {
...
 
-        self.bfs_preload_nodes::<S>(storage, prefixes_set).await
+    /// Given a sorted list of [NodeLabel]s, determine if a given [NodeLabel] is a prefix of any in the list.
+    fn insertion_set_contains(insertion_labels: &[NodeLabel], label: &NodeLabel) -> bool {
+        insertion_labels
+            .binary_search_by(|candidate| {
+                match label.label_len == 0 || label.is_prefix_of(candidate) {
+                    true => Ordering::Equal,
+                    false => candidate.label_val.cmp(&label.label_val),
+                }
```
afterdusk, line 172: Hey @kevinlewi , would we properly handle scenarios like the following?

Say we have this sorted list of labels:

[.., 01, 000, 010, ..]

Say we're looking for 000 and currently at at 01, we would see that 01 is *not* a prefix of 000, then compare the values of 01 and 000. Would we see that 000 is smaller than 01 instead of larger?

afterdusk: Ah, insertion sets are labels of the same length, so the above scenario won't happen. That's the assertion here, right?

kevinlewi: This function assumes that the list of labels passed into it are all of the same length. Indeed, this is the case when using it for the preloading of nodes because the labels are all corresponding to leaf nodes with `label_len = 256`.

So I think the case you mentioned would not occur. Actually, we should add a comment to this function to explicitly clarify this assumption...

afterdusk: Ah gotcha. I'm doing some light refactoring to make it clear where inputs need to be sorted (e.g. for `batch_insert_leaves`, many callsites are not sorting the `insertion_set` even though the function assumes it is sorted), I'll take care of adding the comment. Thanks for clarifying!

afterdusk: @kevinlewi  Is this assertion true even in auditor directory construction? It seems like the `unchanged_nodes` field in an append only proof contains labels of varying lengths:

https://github.com/novifinancial/akd/blob/50ac0043f1d8b601060310e5768ab560da6d987f/akd/src/auditor.rs

You can pull this branch and run tests, there's one test that fails: https://github.com/afterdusk/akd/tree/test-assertion

kevinlewi: Ahh, good point, that assumption is not true for the auditor. However, the auditor starts with an empty AZKS and calls batch_insert_leaves, which means that no pre-loading needs to happen (which is why the function isn't broken for auditors).

Hmm... yeah, not entirely sure what to do about this here. It seems like the perf optimization I made might be a little harder to generalize...

afterdusk: Ah, that makes sense. Appreciate the clarification! I have an idea for how to encode the sorting + equal length requirements for preloading, I'll take a stab at it and we can discuss on the PR 🙂

PR #307, opened 2022-12-26
Optimizing VRF computation by saving expanded private key and public key

kevinlewi:
One more VRF computation optimization that we can do:

Rather than re-expanding the VRF private key into a VRFExpandedPrivateKey and recomputing the public key every time we call evaluate, we can just compute the expanded private key and public key once, and then reuse them across VRF evaluate calls.

Before this change:
```
Batch insertion (100000 initial leaves) took 14172 ms
Batch insertion (100000 initial leaves, 100000 inserted leaves) took 16920 ms
```

After this change:
```
Batch insertion (100000 initial leaves) took 13712 ms
Batch insertion (100000 initial leaves, 100000 inserted leaves) took 15315 ms
```

### DIFF
--- a/akd_core/benches/parallel_vrfs.rs
+++ b/akd_core/benches/parallel_vrfs.rs
@@ -9,6 +9,7 @@
 
 extern crate criterion;
 use self::criterion::*;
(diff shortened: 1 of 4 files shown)


slawlor (approved):
Looks good to me! Anything to help perf wins!

PR #308, opened 2022-12-27
Use enums to distinguish regular and auditor directory constructions

afterdusk:
Per suggestion from @slawlor in #301 , this PR replaces the use of booleans with enums to distinguish between regular and auditor directory constructions/hashing. This is clearer and less error-prone.

Note that there were actually two spots where the parameter of `include_ep` was passed to a function expecting `exclude_ep` without negating the variable (see in-line comment). I believe these were just function naming errors and not actual bugs. This PR will help avoid such mistakes in future.

### DIFF
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -44,10 +44,10 @@ async fn tic_toc<T>(f: impl core::future::Future<Output = T>) -> (T, Option<f64>
 /// mode enum is used to differentiate between the two.
 #[derive(Debug, Clone, Copy)]
 pub(crate) enum InsertMode {
-    /// The regular construction of the the tree
+    /// The regular construction of the the tree.
     Directory,
-    /// The auditor's mode of constructing the tree - last modified epochs of
-    /// leaves are not included in node hashes
+    /// The auditor's mode of constructing the tree - last epochs of leaves are
+    /// not included in node hashes.
     Auditor,
 }
 
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -44,24 +44,18 @@ async fn tic_toc<T>(f: impl core::future::Future<Output = T>) -> (T, Option<f64>
 /// mode enum is used to differentiate between the two.
 #[derive(Debug, Clone, Copy)]
 pub(crate) enum InsertMode {
-    /// The default mode of constructing the tree
+    /// The regular construction of the the tree
     Directory,
     /// The auditor's mode of constructing the tree - last modified epochs of
     /// leaves are not included in node hashes
     Auditor,
 }
 
-impl Default for InsertMode {
-    fn default() -> Self {
-        InsertMode::Directory
-    }
-}
-
-impl From<InsertMode> for HashMode {
+impl From<InsertMode> for NodeHashingMode {
     fn from(mode: InsertMode) -> Self {
         match mode {
-            InsertMode::Directory => HashMode::WithLeafEpoch,
-            InsertMode::Auditor => HashMode::NoLeafEpoch,
+            InsertMode::Directory => NodeHashingMode::WithLeafEpoch,
+            InsertMode::Auditor => NodeHashingMode::NoLeafEpoch,
         }
     }
 }
@@ -140,7 +134,7 @@ impl Azks {
         &mut self,
         storage: &StorageManager<S>,
         insertion_set: Vec<Node>,
-        insert_usage: InsertMode,
+        insert_mode: InsertMode,
     ) -> Result<(), AkdError> {
         // preload the nodes that we will visit during the insertion
         let (fallable_load_count, time_s) =
@@ -166,7 +160,7 @@ impl Azks {
                 Some(NodeLabel::root()),
                 insertion_set,
                 self.latest_epoch,
-                insert_usage,
+                insert_mode,
             )
             .await?;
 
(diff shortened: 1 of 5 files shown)


akd/src/tree_node.rs
```diff
@@ -412,10 +412,17 @@ impl TreeNode {
         new_leaf: Self,
         epoch: u64,
         num_nodes: &mut u64,
-        include_ep: Option<bool>,
+        leaf_hash_mode: HashMode,
     ) -> Result<(), AkdError> {
-        self.insert_single_leaf_helper::<_>(storage, new_leaf, epoch, num_nodes, true, include_ep)
```
afterdusk, line 417: `insert_single_leaf_helper` expects `exclude_ep`, the `include_ep` variable was passed without negation:

https://github.com/novifinancial/akd/blob/9ce9f872885833e8b98d4e3c5925679bb93c7d26/akd/src/tree_node.rs#L443-L451

As mentioned in the PR description I don't believe this is a bug but rather a naming error, would be good if someone else can confirm.

akd/src/tree_node.rs
```diff
@@ -429,10 +436,17 @@ impl TreeNode {
         new_leaf: Self,
         epoch: u64,
         num_nodes: &mut u64,
-        include_ep: Option<bool>,
+        leaf_hash_mode: HashMode,
     ) -> Result<(), AkdError> {
-        self.insert_single_leaf_helper::<_>(storage, new_leaf, epoch, num_nodes, false, include_ep)
```
afterdusk, line 434: Same as above

slawlor:
Looks great to me, but I want @kevinlewi to check the logic to be sure. I'll let him approve

akd/src/tree_node.rs
afterdusk: Ah I believe it was an option simply because there was a default value (false):

https://github.com/novifinancial/akd/blob/eb23bc7aff83c11d0c78416ff0aa8d811a32a0fe/akd/src/tree_node.rs#L413

akd/src/append_only_zks.rs
```diff
@@ -40,6 +40,33 @@ async fn tic_toc<T>(f: impl core::future::Future<Output = T>) -> (T, Option<f64>
...
+    /// The auditor's mode of constructing the tree - last modified epochs of
+    /// leaves are not included in node hashes
+    Auditor,
+}
+
+impl Default for InsertMode {
+    fn default() -> Self {
+        InsertMode::Directory
+    }
+}
```
dillonrg, line 59: We can actually derive Default as well. In addition adding `Default` on the enum itself, we also need to specify which enum value is the default by annotating it with `#[default]`. See here for reference: https://doc.rust-lang.org/std/default/trait.Default.html#enums.

dillonrg (approved):
@afterdusk assuming you've gotten resolution on the question(s) you raised in the PR, this looks good to me! Thank you so, so much for doing this! This makes the code much cleaner and readable. Cheers!

akd/src/tree_node.rs
```diff
@@ -800,6 +799,14 @@ impl TreeNode {
 
 /////// Helpers //////
 
+#[derive(Debug, Clone, Copy)]
+pub(crate) enum HashMode {
```
kevinlewi, line 803: naming nits:
- `HashMode` -> `LeafHashingMode`
- `WithLeafEpoch` -> `WithEpoch`
- `NoLeafEpoch` -> `WithoutEpoch`

akd/src/tree_node.rs
kevinlewi: The best person to ask would be @Jasleen1 , but after looking at the callsites, it does indeed look like just a naming error. Thanks for fixing this!

akd/src/append_only_zks.rs
```diff
@@ -144,7 +177,7 @@ impl Azks {
         storage: &StorageManager<S>,
         insertion_set: Vec<Node>,
     ) -> Result<(), AkdError> {
-        self.batch_insert_leaves_helper::<_>(storage, insertion_set, false)
+        self.batch_insert_leaves_helper::<_>(storage, insertion_set, InsertMode::default())
```
kevinlewi, line 180: Why not just `InsertMode::Directory` here?

akd/src/append_only_zks.rs
kevinlewi: Actually I think we can just get rid of the `impl Default` entirely, and just refer to `InsertMode::Directory()` since it's a purely internal enum and it would be great to be explicit each time it is used.

kevinlewi (approved):
So good!

akd/src/tree_node.rs
afterdusk: Technically the hashes that are affected are the parents of leaf nodes, not the leaf nodes themselves. Would it be confusing to call it `LeafHashingMode`?

akd/src/tree_node.rs
eozturk1: One of the main ideas of oZKS is to include the epoch in the leaf (hash) so that we have a proof of when the leaf was inserted. Looks like a naming error to me.

akd/src/tree_node.rs
kevinlewi: Ah, fair enough! How about `NodeHashingMode` in that case? I basically wanted to stay away from `HashMode` because it sounds a bit too generic and easily-confusable with some other hashing-related stuff. :)

kevinlewi:
LGTM!

Jasleen1:
This looks great, thanks so much!

PR #309, opened 2022-12-27
Parallelizing CI workflow

kevinlewi:
Should allow CI tests to be run in parallel

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,13 +8,67 @@ on:
 
 jobs:
   test:
-    name: Test Rust ${{matrix.toolchain}} on ${{matrix.os}}
-    runs-on: ${{matrix.os}}-latest
+    name: ${{matrix.name}}
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -154,34 +130,33 @@ jobs:
...
-      - name: Test default profile (blake3)
-        working-directory: ./akd_client
-        run: wasm-pack test --node --features wasm
-
-      - name: Test SHA3_256 and no default features
+      - name: Run test
         working-directory: ./akd_client
-        run: wasm-pack test --node --no-default-features --features wasm,sha3_256
\ No newline at end of file
+        run: ${{matrix.command}}
```
slawlor, line 162: nit: could this be something like 

```
wasm-pack ${{matrix.command}} ${{matrix.args}}
```

then the strategy would be a little more standardized with the previous block?

slawlor (approved):
teeny comment, but looks awesome!

.github/workflows/ci.yml
slawlor: whoops, merged without thinking. Maybe i'll clean this up later, sorry!

PR #312, opened 2022-12-28
Reduce Duplicated Code in BFS Preload Paths

dillonrg:
- Prior to this patch, the code for preloading nodes via BFS in both for both the sorted and unsorted variants contained a fair bit of duplicated code. This change introduces a private method which reduces duplication between the two public methods exposed by `Azks`.
- Various other code duplication reductions in tests.

### DIFF
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -196,35 +196,10 @@ impl Azks {
         storage: &StorageManager<S>,
         insertion_labels: &[NodeLabel],
     ) -> Result<u64, AkdError> {
-        let mut load_count: u64 = 0;
-        let mut current_nodes = vec![NodeKey(NodeLabel::root())];
-
-        while !current_nodes.is_empty() {
-            let nodes =
-                TreeNode::batch_get_from_storage(storage, &current_nodes, self.get_latest_epoch())
-                    .await?;
-            load_count += nodes.len() as u64;
-
-            current_nodes = Vec::<NodeKey>::new();
-
-            // Now that states are loaded in the cache, you can read and access them.
-            // Note, the two for loops are needed because otherwise, you'd be accessing remote storage
-            // individually for each node's state.
-            for node in &nodes {
-                if !Self::insertion_set_contains(insertion_labels, &node.label) {
-                    // Only continue to traverse nodes which are relevant prefixes to insertion_set
-                    continue;
-                }
-
-                for dir in crate::DIRECTIONS {
-                    let child_label = node.get_child_label(dir)?;
-                    if let Some(child_label) = child_label {
-                        current_nodes.push(NodeKey(child_label));
-                    }
-                }
-            }
-        }
-        Ok(load_count)
+        self.preload_filtered_nodes(storage, |node| {
+            !Self::insertion_set_contains(insertion_labels, &node.label)
+        })
+        .await
     }
 
     /// Preloads given nodes using breadth-first search.
@@ -233,6 +208,19 @@ impl Azks {
         storage: &StorageManager<S>,
         nodes_to_load: HashSet<NodeLabel>,
     ) -> Result<u64, AkdError> {
+        self.preload_filtered_nodes(storage, |node| !nodes_to_load.contains(&node.label))
+            .await
+    }
+
+    async fn preload_filtered_nodes<S, P>(
+        &self,
+        storage: &StorageManager<S>,
+        mut filter_predicate: P,
+    ) -> Result<u64, AkdError>
+    where
+        S: Database + Send + Sync,
+        P: FnMut(&TreeNode) -> bool,
+    {
         let mut load_count: u64 = 0;
         let mut current_nodes = vec![NodeKey(NodeLabel::root())];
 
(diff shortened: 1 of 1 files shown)


akd/src/append_only_zks.rs
```diff
@@ -1325,4 +1255,17 @@ mod tests {
         ));
         Ok(())
     }
+
+    fn gen_nodes(num_nodes: usize) -> Vec<Node> {
```
afterdusk, line 1259: Maybe this can go into `utils.rs` and be only included in test builds?

akd/src/append_only_zks.rs
```diff
@@ -242,25 +230,28 @@ impl Azks {
...
-            }
+            current_nodes = nodes
+                .iter()
+                .filter(|node| filter_predicate(node))
+                .flat_map(|node| {
+                    DIRECTIONS
+                        .iter()
+                        .filter_map(|dir| {
+                            node.get_child_label(*dir)
+                                .expect(&format!(
```
kevinlewi, line 244: Oh oops, can we instead throw an error here instead of panicking if we encounter this?

akd/src/append_only_zks.rs
```diff
@@ -242,25 +230,28 @@ impl Azks {
...
-                    continue;
-                }
-
-                for dir in crate::DIRECTIONS {
-                    let child_label = node.get_child_label(dir)?;
-                    if let Some(child_label) = child_label {
-                        current_nodes.push(NodeKey(child_label));
-                    }
-                }
-            }
```
afterdusk, line 262: Might be an unpopular opinion, but I actually find this imperative approach easier to understand than the functional one. I think the functional approach can be simplified further though (see clippy lints), maybe that would help readability :)

akd/src/append_only_zks.rs
dillonrg: I actually think we may want to panic here. Reason being that the `DIRECTIONS` const *should only ever contain valid* directions: https://github.com/facebook/akd/blob/main/akd_core/src/types/mod.rs#L36-L37. I opted for a panic, as if we attempt to move in an invalid direction here, that seems like it would mean we may have a bug in lib.

Let me know what you think!

akd/src/append_only_zks.rs
dillonrg: Good suggestion! I think we may have a few other spots floating around where this is copy/pasted.

akd/src/append_only_zks.rs
dillonrg: Fair enough! I debated between the two, as I found the iteration over `DIRECTIONS` a bit "meh". I have incorporated the clippy lints as this point, and that does help a bit, so let me know what you think! I am happy to go with whatever the majority finds most readable on this one.

afterdusk: Ah yes this is better! If we want to avoid writing the panic statement, we can also do:
```rust
current_nodes = nodes
    .iter()
    .filter(|node| filter_predicate(node))
    .flat_map(|node| {
        DIRECTIONS
            .iter()
            .filter_map(|dir| node.get_child_label(*dir).ok().flatten().map(NodeKey))
            .collect::<Vec<NodeKey>>()
    })
    .collect();
```
But if we want an explicit panic that's fine too 🙂

dillonrg: I personally prefer the explicit panic, as I find it more explicit. That said, I'd like to get a confirmation from @kevinlewi around a panic at all, as this used to just result in an error. I questioned that, as `DIRECTIONS` is only left and right + the method we're calling is OK with left and right.

kevinlewi: Main thing is that a panic will result in a server crashing, right? The `get_child_label()` call could throw a StorageError due to storage being inaccessible for some reason, and that error would need to be handled

akd/src/append_only_zks.rs
kevinlewi: See other comment about the function potentially throwing a StorageError. That said, in general, if we did have a situation where something should never happen, it would be better to see if we can make it enforced at compile-time rather than panicking at runtime.

akd/src/append_only_zks.rs
dillonrg: Yes, that's right! I think the `get_child_label()` doesn't throw errors unless `Direction::None` is passed (see: https://github.com/facebook/akd/blob/main/akd/src/tree_node.rs#L503-L514). Since `DIRECTIONS` is composed of only `Left` and `Right`, it seemed like we could safely assume that this call should never result in an error at this callsite. Please do let me know if I am overlooking something!

kevinlewi: Oh sorry, I was looking at `get_child_node()` and not `get_child_label()`, my bad. I see what you mean.

Still, it feels like maybe the right thing to do would be to get rid of Direction::None, or create a DirectionWithoutNone kind of enum and have get_child_label() function take that as an argument, since get_child_label() should ideally not throw an error there either. But this is of course a much larger refactor...

kevinlewi: Perhaps we can put keep the panic for now, but add a TODO to remove it? I just am not a fan of having a possibility for the server to completely crash at run-time (as opposed to a catch-able error).

dillonrg: @kevinlewi definitely down for that! I agree that this should be a compile-time error, if possible. I also agree that seems like a slightly larger refactor, so I am happy to tackle that in a separate change since it'll probably sprawl out a bit more.

If you're most comfortable with maintaining existing behavior as opposed to potentially introducing a server crash, I am fine with that as well. I honestly opted for the panic here to make the error more apparent in the event that it does happen considering that it really shouldn't and may be hard to find if it did.

Let me know what you think :D

kevinlewi: Sure, if you can file an issue to track this, and keep the expect there, that sounds good to me!

dillonrg: @kevinlewi created https://github.com/facebook/akd/issues/314 to track the issue. I'll add a FIXME/TODO to this section of code to track the issue.

kevinlewi (approved):
Thanks!!

akd/src/append_only_zks.rs
```diff
@@ -196,35 +196,10 @@ impl Azks {
...
-        }
-        Ok(load_count)
+        self.preload_filtered_nodes(storage, |node| {
+            !Self::insertion_set_contains(insertion_labels, &node.label)
```
dillonrg, line 200: This check is backwards. Given that this isn't used to "continue" any longer, and instead filters, then this shouldn't be a NOT.

akd/src/append_only_zks.rs
```diff
@@ -233,6 +208,19 @@ impl Azks {
         storage: &StorageManager<S>,
         nodes_to_load: HashSet<NodeLabel>,
     ) -> Result<u64, AkdError> {
+        self.preload_filtered_nodes(storage, |node| !nodes_to_load.contains(&node.label))
```
dillonrg, line 211: Same point about this filter.

PR #313, opened 2022-12-28
Add abstraction for insertion sets and optimize recursive insert

afterdusk:
# Overview
1. Adding a `NodeSet` enum that encodes information about whether a vector of nodes is binary searchable (i.e. nodes have the same label length AND are sorted). Our functions can then decide whether to apply binary search based on the enum variant.
    - This addresses the [minor issue in #306](https://github.com/facebook/akd/pull/306/files#r1057887531) where insertion sets are assumed to be binary searchable when they may not necessarily be.

2. Adding binary search to the partitioning of insertion sets + more efficient computation of LCP in an insertion set, as [suggested in #301](https://github.com/facebook/akd/pull/301#issuecomment-1365706006)

# Benchmark
Used 5fdad25 to represent main branch (it includes the fix for the benchmark I am running). Ran:
```console
```

## Inserting 10K Nodes

Main:
image

This PR:
image

No improvements observed at this scale.

## Inserting 100K Nodes
Main:
image

This PR:
image

Minor perf improvements observed.

### DIFF
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -135,15 +135,10 @@ impl InsertionSet {
 
     pub(crate) fn get_longest_common_prefix(&self) -> NodeLabel {
         match self {
-            InsertionSet::BinarySearchable(nodes) => {
-                // TODO: Implement binary search
-                if nodes.is_empty() {
-                    return EMPTY_LABEL;
-                }
-                nodes.iter().skip(1).fold(nodes[0].label, |acc, node| {
-                    node.label.get_longest_common_prefix(acc)
-                })
-            }
+            InsertionSet::BinarySearchable(nodes) => match (nodes.first(), nodes.last()) {
+                (Some(first), Some(last)) => first.label.get_longest_common_prefix(last.label),
+                _ => EMPTY_LABEL,
+            },
             InsertionSet::Unsorted(nodes) => {
                 if nodes.is_empty() {
                     return EMPTY_LABEL;
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -96,22 +96,27 @@ impl From<Vec<Node>> for InsertionSet {
 }
 
 impl InsertionSet {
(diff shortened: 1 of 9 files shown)


akd/src/append_only_zks.rs
```diff
@@ -60,6 +60,106 @@ impl From<InsertMode> for NodeHashingMode {
...
+}
+
+impl From<Vec<Node>> for InsertionSet {
+    fn from(mut nodes: Vec<Node>) -> Self {
+        if !nodes.is_empty()
+            && nodes
+                .iter()
+                .all(|node| node.label.label_len == nodes[0].label.label_len)
+        {
+            nodes.sort();
```
slawlor, line 90: since we know we can't have duplicates, unstable sort is faster here, we should use it

afterdusk: I think we would have been fine even with duplicates, thanks for the suggestion!

akd/src/storage/cache/high_parallelism.rs
```diff
@@ -34,6 +34,8 @@ pub struct TimedCache {
     can_clean: Arc<AtomicBool>,
     item_lifetime: Duration,
     memory_limit_bytes: Option<usize>,
+
+    #[cfg(feature = "runtime_metrics")]
```
slawlor, line 38: Doing this, you need to do it everywhere it's referenced which breaks a lot of stuff. Essentially what I had done, was to create the field but only populate/update it when the feature is enabled. That way it's just a non-accessed 0-value.

afterdusk: Ah yeah, but there was a 'is never read' warning and I was hoping to address it 

image

slawlor:
Are you good to properly review this? I think it looks good to me!

afterdusk:
@slawlor It's ready now! Was just finishing up some final tweaks + the benchmarks 🙂

slawlor (approved):
looks good to me!

kevinlewi (approved):
This is a great restructuring! Thanks!

PR #315, opened 2022-12-29
Fixing VRF map for already-existing labels and introducing VersionFreshness enum

### LINKED ISSUE #304: Add usage of VRF map for already-existing labels in Directory::publish()
In #295, we introduced a VRF map in directory publish that precomputes the VRF labels in parallel.

This map is used for the case where a label has not been encountered before, like so: https://github.com/novifinancial/akd/blob/main/akd/src/directory.rs#L147-L151

However, it is not used for the case where a label has already been encountered, see: https://github.com/novifinancial/akd/blob/main/akd/src/directory.rs#L169-L176

In this second case, note that we actually do two VRF computations, one for the stale label and one for the fresh label. We should add usage of the VRF map for this case as well.

cc: @slawlor

kevinlewi:
Closes #304 by using a flat_map in publish when computing the VRF map.

Also, instead of using a boolean stale value passed around, I am introducing a `VersionFreshness` enum to track whether or not a node is marked as stale or fresh.

### DIFF
--- a/akd_core/src/utils.rs
+++ b/akd_core/src/utils.rs
@@ -8,7 +8,7 @@
 //! Utility functions
 
 use crate::hash::Digest;
-use crate::{AkdLabel, AkdValue, NodeLabel};
+use crate::{AkdLabel, AkdValue, NodeLabel, VersionFreshness};
 
 #[cfg(feature = "nostd")]
 use alloc::vec::Vec;
@@ -63,7 +63,7 @@ pub fn get_commitment_nonce(
 }
 
 /// To convert a regular label (arbitrary string of bytes) into a [NodeLabel], we compute the
-/// output as: H(label || stale || version)
+/// output as: H(label || liveness || version)
 ///
 /// Specifically, we concatenate the following together:
 /// - I2OSP(len(label) as u64, label)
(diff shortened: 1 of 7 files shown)


slawlor:
Are you missing a commit to push? I have like a ton of failing signals because of a bad reference to a struct instead of the struct itself.

kevinlewi:
Sorry, should be ready to review now.

akd_core/src/utils.rs
```diff
@@ -63,20 +63,27 @@ pub fn get_commitment_nonce(
...
 /// These are all interpreted as a single byte array and hashed together, with the output
 /// of the hash returned.
-pub(crate) fn get_hash_from_label_input(label: &AkdLabel, stale: bool, version: u64) -> Vec<u8> {
-    let stale_bytes = if stale { &[0u8] } else { &[1u8] };
+pub(crate) fn get_hash_from_label_input(
+    label: &AkdLabel,
+    freshness: VersionFreshness,
+    version: u64,
+) -> Vec<u8> {
+    let freshness_bytes = match freshness {
```
slawlor, line 79: should this be part of the struct? Like a deref trait impl on the struct? Just to avoid hardcoding here and keep the logic together. Or add a `to_bytes()` method perhaps?

slawlor: Or perhaps use the actual `#[repr(u8)]` functionality? We can probably cast it so we don't have just hardcoded bytes outside of the struct def

slawlor (approved):
Looks good overall, great cleanup, but there's  1 small adjustment i think then we can merge it!

akd_core/src/utils.rs
slawlor: Feel free to merge!

Issue #314, opened 2022-12-28
Convert Invalid Directional Movements During Preloading to Compile-time Errors


dillonrg:
In https://github.com/facebook/akd/pull/312, we discovered that during preloading in the `Azks` struct there exists a codepath which "moves" in valid directions (e.g., left and right) as specified by the [`DIRECTIONS` constant within `akd_core`](https://github.com/facebook/akd/blob/main/akd_core/src/types/mod.rs#L36-L37). As part of these valid moves, we invoke the [`get_child_label` method defined within `tree_node`, which never throws in the event valid directions are passed to it as arguments](https://github.com/facebook/akd/blob/main/akd/src/tree_node.rs#L503-L514).

Considering that we are only ever moving in "valid" directions as defined by the constant, we `panic` if we attempt to move in an "invalid" direction according to the `get_child_label` method. The rationale is that panicking is valid in this case, as we really should never attempt to move in an invalid direction and that may indicate a bug within `akd`. Panicking here is done in an attempt to better surface "shouldn't happen" scenarios. 

Overall, we'd prefer any attempts to move in an invalid direction to be a compile-time error. That is, instead of panicking at runtime and crashing the server, we'd like to prevent `akd` from compiling altogether. Potential solutions include encoding only valid directions as their own type, but further evaluation is required before a concrete solution is reached.

PR #316, opened 2022-12-29
Add TODO for Issue 314

dillonrg:
- Add a TODO linking a panic that we'd like to convert into a compile-time error instead. For more information, please see: https://github.com/facebook/akd/issues/314.


dillonrg:
I goofed and thought I had amended my original commit (with actual changes) with the added TODO. Looks like I was wrong 🤦 .

slawlor (approved):
LGTM

Issue #318, opened 2022-12-29
Issue with preload logic


kevinlewi:
This check is backwards. Given that this isn't used to "continue" any longer, and instead filters, then this shouldn't be a NOT.

_Originally posted by @dillonrg in https://github.com/facebook/akd/pull/312#discussion_r1059143656_

dillonrg:
The core bug has been identified and fixed with #320. However, verification of the fix was done manually. We'd also like to implement some automated tests to ensure that this issue does not pop up again.

PR #320, opened 2022-12-29
Invert Filter on Preload to Ensure Appropriate Nodes are Preloaded

dillonrg:
- Prior to a more functional approach to preloading being adopted in #312, the "filter" that was being applied was used to continue within a loop. However, now that we're filtering in an iterator, the check should be inverted such that we continue iterating over the appropriate nodes.

- [x] Manually test to verify that preload is preloading more than one value
- [x] Add unit tests to catch this case

### DIFF
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -861,6 +861,7 @@ type AppendOnlyHelper = (Vec<Node>, Vec<Node>);
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::storage::types::DbRecord;
     use crate::utils::byte_arr_from_u64;
     use crate::{
         auditor::audit_verify,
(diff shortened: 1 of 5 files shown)


slawlor:
Just checking, is this #318?

dillonrg:
> Just checking, is this #318?

Yes, yes it is! Let me take that issue and link this.

slawlor:
Change looks fine, but it'd be good to cover it with a test prior to merging. Thanks!

dillonrg:
> Change looks fine, but it'd be good to cover it with a test prior to merging. Thanks!

Test has been added! There are two commits as part of this PR. The first introduces a failing test (disabled for CI purposes) to demonstrate the bug. The second then fixes the bug and enables the test to verify that preload is loading up the expected value.

slawlor (approved):
Awesome!

PR #321, opened 2022-12-29
Add Send and Sync trait bounds to Database trait

afterdusk:
Per suggestion in https://github.com/facebook/akd/pull/302#discussion_r1057996150, this simplifies the use of the `Database` trait. It seems reasonable to expect that all implementors of the trait will be thread-safe.


dillonrg (approved):
Thanks for making this change, @afterdusk! I literally made this change locally to make testing for #320 a little bit easier. This looks good to me :)

slawlor (approved):
There were a lot of those lol. Nice work

PR #322, opened 2022-12-30
Preload nodes in lookups

eozturk1:
In `batch_lookups` we utilize node preloading to minimize individual `GET` operations. We can do the same optimization for regular lookups -- which this PR adds. Performance results are shown below. 

In summary, instead of 40 individual `GET` operations, with this PR, we only have 2 `GET` + 8 `BATCH GET` operations which gives a total of 10 compared to 40. This should improve the performance of lookups quite a bit considering database access delays.

---
Without this PR:

```
Post look-up metrics:
Cache hit since last: 47, cached size: 40 items

    SET 2, 
    BATCH SET 3, 
    GET 40, 
    BATCH GET 0
    TOMBSTONE 0
    GET USER STATE 1
    GET USER DATA 0
    GET USER STATE VERSIONS 3
============ Database operation timing ============
    TIME READ 98 ms
    TIME WRITE 1380 ms
```

With this PR:
```
Post look-up metrics:
Cache hit since last: 80, cached size: 40 items

    SET 2, 
    BATCH SET 3, 
    GET 2, 
    BATCH GET 8
    TOMBSTONE 0
    GET USER STATE 1
    GET USER DATA 0
    GET USER STATE VERSIONS 3
============ Database operation timing ============
    TIME READ 143 ms
    TIME WRITE 1379 ms
 ```
  
 ---
> **NOTE:** Tests done after patching #320. A single lookup operation costs are shown only for post-lookup. Generated with the diff below:
 
 ```diff
+            mysql_db.flush_cache().await;
+            println!("Pre look-up metrics:");
+            mysql_db.log_metrics(log::Level::Trace).await;
             // Perform 10 random lookup proofs on the published users
             for user in users.iter().choose_multiple(&mut rng, 10) {
                 let key = AkdLabel::from_utf8_str(user);
                 match dir.lookup(key.clone()).await {
                     Err(error) => panic!("Error looking up user information {:?}", error),
                     Ok((proof, root_hash)) => {
+                        println!("Post look-up metrics:");
+                        mysql_db.log_metrics(log::Level::Trace).await;
+                        return;
```
---

### DIFF
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -18,6 +18,7 @@ use crate::{
 };
 
 use akd_core::SizeOf;
+use akd_core::hash::EMPTY_DIGEST;
 use async_recursion::async_recursion;
 use log::info;
 use std::cmp::Ordering;
@@ -287,6 +288,22 @@ impl Azks {
         }
 
         Ok(())
+    }
+
+    pub(crate) async fn preload_lookup_nodes<S: Database + Send + Sync>(&self, storage: &StorageManager<S>, lookup_labels: &[NodeLabel]) -> Result<u64, AkdError> {
+        // Create nodes for labels.
+        let lookup_nodes: Vec<Node> = lookup_labels
+            .into_iter()
+            .map(|&l| Node {
+                label: l,
+                hash: EMPTY_DIGEST,
+            })
+            .collect();
+
+        // Load nodes. Note NodeSet will sort these nodes for efficient preloading.
+        self
+            .preload_nodes(storage, &NodeSet::from(lookup_nodes))
+            .await
     }
 
     /// Preloads given nodes using breadth-first search.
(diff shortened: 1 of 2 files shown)


slawlor:
Lints and rustfmt? Change looks good

akd/src/append_only_zks.rs
```diff
@@ -289,6 +291,26 @@ impl Azks {
...
+        let lookup_nodes: Vec<Node> = lookup_infos
+            .iter()
+            .flat_map(|li| vec![li.existent_label, li.marker_label, li.non_existent_label])
+            .map(|l| Node {
+                label: l,
+                hash: EMPTY_DIGEST,
+            })
+            .collect();
+
+        // Load nodes. Note NodeSet will sort these nodes for efficient preloading.
```
afterdusk, line 309: I think we can drop the sorting comment, how `NodeSet` represents the data is an internal detail.

afterdusk (approved):
LGTM as well 👍

kevinlewi:
Question: should we do something similar like this for history proof generation?

eozturk1:
> Question: should we do something similar like this for history proof generation?

I need to look into how we load the nodes but I think we can. I'll add it next week.

Another possibility is batch history proofs similar to batch lookups.

PR #323, opened 2022-12-30
Skip preloading when cache is not present

afterdusk:
Preloading nodes only makes sense when a cache is present. This PR adds a check to skip preloading if the storage manager does not have a cache.

Also moved the `recursive_batch_insert_nodes` function up next to `batch_insert_nodes`.


slawlor (approved):
Looks good!

kevinlewi (approved):

PR #324, opened 2022-12-30
Parallelism for hash insertion and recursive tree walk

slawlor:
Eventually a controllable levle is available.

~610ms with level of parallelism = 2
~ 1.25s with no parallelism


slawlor:
Was an example for @afterdusk 's PR, so closing since irrelevant

PR #325, opened 2022-12-31
Remove reference passing in CTORs

slawlor:
There's a few instances where we take a reference to a struct an nearly immediately clone it. This should be left to the caller, since if they are only say constructing a database to pass it to the storage manager, we may be inducing a clone for no real reason.


dillonrg (approved):
Thanks, @slawlor! This looks great :)

PR #327, opened 2022-12-31
Remove Pub Modifier on StorageManager DB

dillonrg:
- Remove the pub access modifier to the Database impl field in the `StorageManager` struct. Database access should only be performed through an instance of `StorageManager`.


slawlor:
Can we safely close this? I think this was merged in from another PR

dillonrg:
> Can we safely close this? I think this was merged in from another PR

Yes, we can close this out given that it was effectively completed with #333! Thanks, @slawlor!

PR #328, opened 2022-12-31
Reduce and cleanup writes to storage during insertion

afterdusk:
# Overview
Now that node insertion is fully recursive, it's easy to identify the redundant calls writing nodes to storage. While we have a transaction object that make these writes fairly fast, they still result in some contention under parallelism. Furthermore, the inconsistent callsites make it unclear when objects have already been written to storage. This PR cuts out the unnecessary calls.

# Benchmark
Ran the `azks.rs` benchmark, which inserts 100K leaves into a tree with 10K leaves.:

```console
```

Note that the benchmark does use a transaction object (as a write cache) as opposed to writing to the DB directly, which is reflective of real-world usage.

## Main

image

## This PR

image

That's... quite a huge improvement 😀

### DIFF
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -296,7 +296,7 @@ impl Azks {
 
         if !node_set.is_empty() {
             // call recursive batch insert on the root
-            let (_, num_inserted) = Self::recursive_batch_insert_nodes(
+            let (root_node, num_inserted) = Self::recursive_batch_insert_nodes(
                 storage,
                 Some(NodeLabel::root()),
                 node_set,
@@ -305,6 +305,7 @@ impl Azks {
                 get_parallel_levels(),
             )
             .await?;
+            root_node.write_to_storage(storage).await?;
 
             // update the number of nodes
             self.num_nodes += num_inserted;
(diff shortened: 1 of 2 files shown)


kevinlewi:
Thanks! Probably optional, but, is there a way in which we can add a test for this, to ensure that we are never overwriting a node that we have already written to? Perhaps by implementing a custom storage layer and testing for this specifically?

Otherwise, looks good!

afterdusk:
@kevinlewi Ah, I think a custom storage implementation or using a mocking library should work! 

However I just found a way to reduce the writes even further, and now all the non-test calls to `write_to_storage` happen inside `append_only_azks.rs`. Tracing the code, it should be pretty clear that we're only writing nodes in these places:
1. Writing the root node when creating a new `Azks`
2. In batch insert:
  a. When a node gets pushed down
  b. After a node has been returned from the recursive call

I'm more confident we're not overwriting data now than before, so I'll go ahead and merge this PR first if that's fine with you. I'll work on testing in a separate PR!

Btw, I also updated the benchmark after my latest change, it looks like the execution time reduction is now close to 40%!

kevinlewi:
Looks good to me, thank you!

To clarify, this PR doesn't actually affect how long it takes the do the final transaction commit, right? It only speeds up the transaction-writing step while we are figuring out which nodes to update hashes for and which new nodes to write?

kevinlewi (approved):

afterdusk:
@kevinlewi That's right! This only affects how many times we are writing to the transaction object (an in-memory sharded hashmap). Given that we now have multiple threads writing to it after #302 , reducing the number of writes reduces contention between threads.

The final transaction commit step is unchanged!

PR #329, opened 2023-01-01
Specify blake3 feature flag when publishing

afterdusk:
The GH actions run to publish the v0.8.3 crate was failing: https://github.com/facebook/akd/actions/runs/3816632953

In #292 we dropped the `blake3` feature flag from the `akd_core` crate's default features. Therefore we now need to specify the feature flag when publishing so that a hash function is included in the build. 

Running the new command in dry-run mode succeeds locally:
```console
$ cargo publish --dry-run --manifest-path Cargo.toml -p akd_core -F blake3
```


slawlor (approved):
Looks good

PR #330, opened 2023-01-01
Bump version to 0.8.4 to re-attempt publish

afterdusk:
Unfortunately, the workflow to publish the akd v0.8.3 failed after akd_core and akd_client v0.8.3 was published: https://github.com/facebook/akd/actions/runs/3817658498/jobs/6494059254

We also have to bump the version number for all crates to attempt a republish, since we cannot overwrite already published crates.

### DIFF
--- a/akd/Cargo.toml
+++ b/akd/Cargo.toml
@@ -62,7 +62,6 @@ once_cell = { version = "1" }
 ctor = "0.1"
 tokio-test = "0.4"
 tokio = { version = "1.21", features = ["rt", "sync", "time", "macros"] }
-akd = { path = ".", features = ["public-tests"], version = "0.8.3", default-features = false }
 
 [[bench]]
 name = "azks"
--- a/akd/Cargo.toml
+++ b/akd/Cargo.toml
@@ -63,6 +63,9 @@ ctor = "0.1"
 tokio-test = "0.4"
 tokio = { version = "1.21", features = ["rt", "sync", "time", "macros"] }
 
+# To enable the public-test feature in tests
+akd = { path = ".", features = ["public-tests"], default-features = false }
+
 [[bench]]
 name = "azks"
 harness = false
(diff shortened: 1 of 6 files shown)


akd/Cargo.toml
```diff
@@ -62,7 +62,6 @@ once_cell = { version = "1" }
 ctor = "0.1"
 tokio-test = "0.4"
 tokio = { version = "1.21", features = ["rt", "sync", "time", "macros"] }
-akd = { path = ".", features = ["public-tests"], version = "0.8.3", default-features = false }
```
afterdusk, line 65: @slawlor Is this fine to remove?

slawlor: No just remove the version piece of it. In a test profile, we want to reference the latest code we have so we don't specify a version number and therefore it won't try to ever pull from crates.io but will build from local source code

slawlor (changes requested):
Small change

akd/Cargo.toml
slawlor: We don't want to include the public tests feature always, hence why we re-import the library under the Dev profile saying the default in Dev, also known as test, is to enable the public tests feature. This way you can build, and have a default set of features, which is different from the test profile

afterdusk: Ah that makes sense! The reason why we have this line is also manifesting in CI. I wonder why we included a version number in the first place...

slawlor: Iirc we never had a version number did we? Anyways this should fix it

slawlor (approved):

akd/Cargo.toml
afterdusk: Looks like we didn't have a version number for a while, then 0.8.0 was added in #292. It might have been an accident, let's revert it to version-less!

PR #331, opened 2023-01-03
Avoid looking up previous version of nodes that are new

afterdusk:
# Overview

When writing tree nodes to storage, we look up previous versions of the node to create the `TreeNodeWithPreviousValue` struct. However, this is not necessary when we know a node is new and does not have a previous value. By keeping track of whether a node is new, we save on incurring an unnecessary round trip to the database.

This PR performs the requisite change.

I would have liked to add a test for this change, however I cannot think of a straightforward way to assert that the database is not read beyond the initial preload. I believe this is something that can be addressed by bringing in a mocking library to mock out the `Database` trait, or implementing an in-memory database which we can mock. Filed an issue regarding this: #332

# Manual Test
Added the following test to `append_only_zks.rs`, which batch inserts 10 nodes on an empty tree and prints database access metrics:
```rust
#[tokio::test]
async fn test_get_count() -> Result<(), AkdError> {
    let num_nodes = 10;
    let node_set = gen_nodes(num_nodes);

    let database = AsyncInMemoryDatabase::new();
    let db = StorageManager::new(database.clone(), None, None, None);
    let mut azks = Azks::new(&db).await?;

    db.begin_transaction();
    azks.batch_insert_nodes(&db, node_set, InsertMode::Directory)
        .await?;
    db.log_metrics(log::Level::Trace).await;

    Ok(())
}
```
The test was ran with the following command:
```console
$ cargo test test_get_count -- --nocapture
```
If the change works, no gets are expected as the insert was performed on an empty tree with nothing to preload.
## Before
On the `main` branch, with 7878bcb cherry-picked:
```
    SET 1, 
    BATCH SET 0, 
    GET 19, 
    BATCH GET 0
    TOMBSTONE 0
    GET USER STATE 0
    GET USER DATA 0
    GET USER STATE VERSIONS 0
```
19 gets are performed.
## After
On this PR's branch:
```
    SET 1, 
    BATCH SET 0, 
    GET 0, 
    BATCH GET 0
    TOMBSTONE 0
    GET USER STATE 0
    GET USER DATA 0
    GET USER STATE VERSIONS 0
```

Indeed the number of gets are now 0.

### DIFF
--- a/akd/Cargo.toml
+++ b/akd/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "akd"
-version = "0.8.4"
+version = "0.8.5"
 authors = ["Harjasleen Malvai <hmalvai@fb.com>", "Kevin Lewi <klewi@fb.com>", "Sean Lawlor <seanlawlor@fb.com>"]
 description = "An implementation of an auditable key directory"
 license = "MIT OR Apache-2.0"
@@ -34,7 +34,7 @@ default = ["blake3", "public_auditing", "parallel_vrf", "parallel_insert"]
 
 [dependencies]
 ## Required dependencies ##
-akd_core = { path = "../akd_core", version = "0.8.4", default-features = false, features = ["vrf"] }
+akd_core = { path = "../akd_core", version = "0.8.5", default-features = false, features = ["vrf"] }
 async-recursion = "0.3"
 async-trait = "0.1"
 curve25519-dalek = "3"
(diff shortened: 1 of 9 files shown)


eozturk1:
Can we get the statistics on an integration test suite to see whether we are using any `GET` calls that hits the database? Example in #223.

slawlor (approved):
We'll follow up with automated tests once we have a storage mock imported

eozturk1 (approved):
Thanks for sharing the metrics, looks great!

Issue #332, opened 2023-01-03
Implement mockable in-memory database


afterdusk:
This would be helpful in tests to assert expected database access patterns. For example, if unexpected reads or writes are performed. See:
- https://github.com/facebook/akd/pull/328#issuecomment-1368307195
- https://github.com/facebook/akd/pull/331#issue-1517967955

slawlor:
@afterdusk I think we now support the mocks, but I guess do we want to leave this task open to implement the test adjustments we want now?

PR #333, opened 2023-01-03
Adding mock support for the Database trait in AKD

slawlor:
Related to: #332

Tests to come

### DIFF
--- a/akd/src/storage/manager/mod.rs
+++ b/akd/src/storage/manager/mod.rs
@@ -51,14 +51,24 @@ mod tests;
 
 /// Represents the manager of the storage mediums, including caching
 /// and transactional operations (creating the transaction, committing it, etc)
-#[derive(Clone)]
 pub struct StorageManager<Db: Database> {
     cache: Option<TimedCache>,
     transaction: Transaction,
     /// The underlying database managed by this storage manager
-    pub db: Db,
+    pub db: Arc<Db>,
 
     metrics: [Arc<AtomicU64>; NUM_METRICS],
+}
+
+impl<Db: Database> Clone for StorageManager<Db> {
+    fn clone(&self) -> Self {
+        Self {
+            cache: self.cache.clone(),
+            transaction: self.transaction.clone(),
+            db: self.db.clone(),
+            metrics: self.metrics.clone(),
+        }
+    }
 }
 
 unsafe impl<Db: Database> Sync for StorageManager<Db> {}
@@ -66,7 +76,7 @@ unsafe impl<Db: Database> Send for StorageManager<Db> {}
 
 impl<Db: Database> StorageManager<Db> {
     /// Create a new storage manager with NO CACHE
-    pub fn new_no_cache(db: Db) -> Self {
+    pub fn new_no_cache(db: Arc<Db>) -> Self {
         Self {
             cache: None,
             transaction: Transaction::new(),
@@ -77,7 +87,7 @@ impl<Db: Database> StorageManager<Db> {
 
     /// Create a new storage manager with a cache utilizing the options provided (or defaults)
     pub fn new(
-        db: Db,
+        db: Arc<Db>,
         cache_item_lifetime: Option<Duration>,
         cache_limit_bytes: Option<usize>,
         cache_clean_frequency: Option<Duration>,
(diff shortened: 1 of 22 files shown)


akd/src/tests.rs
```diff
@@ -8,16 +8,78 @@
...
+            flag: ValueStateRetrievalFlag,
+        ) -> Result<ValueState, StorageError>;
+
+        /// Retrieve the user -> state version mapping in bulk. This is the same as get_user_states but with less data retrieved from the storage layer
+        async fn get_user_state_versions(
+            &self,
+            usernames: &[AkdLabel],
+            flag: ValueStateRetrievalFlag,
+        ) -> Result<HashMap<AkdLabel, (u64, AkdValue)>, StorageError>;
+    }
```
dillonrg, line 80: I suspect that this was just "grabbed" from the Database trait, but I think we can do without the rustdoc comments on the mock.

akd/src/tests.rs
```diff
@@ -8,16 +8,78 @@
...
+#[derive(Clone)]
+pub struct LocalDatabase;
+
+unsafe impl Send for LocalDatabase {}
+unsafe impl Sync for LocalDatabase {}
+
+mockall::mock! {
+    pub LocalDatabase {
+
+    }
```
dillonrg, line 37: Nit: Any reason not to use `MockDatabase`? I find it a bit more apt than the "Local" convention.

dillonrg (approved):
Just some minor suggestions, but looks fine otherwise! Thanks for getting this up, @slawlor!

akd/src/tests.rs
slawlor: The "Mock" prepend is added by the `mock!` macro, so it'll be MockLocalDatabase. I didn't want to define a `Database` struct to risk clashing with the trait.

dillonrg: Ahh, that's right! I totally forgot that the "mock" prefix is added with the `mock!` macro. I thought only the `#[automock]` attribute added it for some reason. Thanks!

akd/src/storage/manager/mod.rs
```diff
@@ -51,16 +51,26 @@ mod tests;
...
+    fn clone(&self) -> Self {
+        Self {
+            cache: self.cache.clone(),
+            transaction: self.transaction.clone(),
+            db: self.db.clone(),
+            metrics: self.metrics.clone(),
+        }
+    }
+}
+
```
dillonrg, line 73: Can't we just derive Clone instead?

akd/src/tests.rs
```diff
@@ -1057,6 +1165,67 @@ async fn test_tombstoned_key_history() -> Result<(), AkdError> {
...
+    // Create a set with 2 updates, (label, value) pairs
+    // ("hello10", "hello10")
+    // ("hello11", "hello11")
+    let mut updates = vec![];
+    for i in 0..1 {
+        updates.push((
+            AkdLabel(format!("hello1{}", i).as_bytes().to_vec()),
+            AkdValue(format!("hello1{}", i).as_bytes().to_vec()),
+        ));
+    }
```
dillonrg, line 1192: Small nit: I think this can be written out as
```rust
let updates = (0..1).map(|i| (AkdLabel(...), AkdVaue(...))).collect();
```

akd/src/tests.rs
```diff
@@ -1057,6 +1165,67 @@ async fn test_tombstoned_key_history() -> Result<(), AkdError> {
...
+        .expect("Failed to create directory");
+
+    // create more updates
+    let mut updates = vec![];
+    for i in 0..1 {
+        updates.push((
+            AkdLabel(format!("hello1{}", i).as_bytes().to_vec()),
+            AkdValue(format!("hello1{}", i + 1).as_bytes().to_vec()),
+        ));
+    }
```
dillonrg, line 1220: Same comment as above

afterdusk (commented):
I can't recall off-hand if we have any AKD user code that creates multiple `StorageManager`s with the same database struct. If we do, we might run into issues because:
1.  the database trait no longer inherits `Clone`
2. there's no means of creating a `StorageManager` by passing in an `Arc` of a DB.

Either ways we can address the issue if we run into it - thanks for adding mocking so quickly @slawlor !

PR #334, opened 2023-01-04
Skip metric logging if runtime_metrics feature is not enabled

afterdusk:
Currently, non-populated transaction metrics and database metrics are printed when the `runtime_metrics` feature is not enabled. We can skip these for cleaner logs.

This PR also adds the number of records to the transaction commit log statement.

### DIFF
--- a/akd/src/storage/tests.rs
+++ b/akd/src/storage/tests.rs
@@ -381,7 +381,7 @@ async fn test_transactions<S: Database>(storage: &StorageManager<S>) {
     let tic = Instant::now();
     assert!(storage.begin_transaction());
     assert_eq!(Ok(()), storage.batch_set(new_data).await);
-    assert_eq!(Ok(()), storage.commit_transaction().await);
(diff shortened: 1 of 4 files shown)


akd/src/storage/tests.rs
```diff
@@ -381,7 +381,7 @@ async fn test_transactions<S: Database>(storage: &StorageManager<S>) {
     let tic = Instant::now();
     assert!(storage.begin_transaction());
     assert_eq!(Ok(()), storage.batch_set(new_data).await);
-    assert_eq!(Ok(()), storage.commit_transaction().await);
+    assert!(matches!(storage.commit_transaction().await, Ok(_)));
```
eozturk1, line 384: Maybe `assert!(storage.commit.transaction().await.is_ok()`?

slawlor (approved):
LGTM

eozturk1 (approved):
LGTM, thanks for the clean up!

PR #335, opened 2023-01-05
Renaming of functions / fields (part 1)

kevinlewi:
Renaming of:
- `struct Node` -> `struct AzksElement`
- field `hash` of `AzksElement` to be called `value`
- `NodeSet` -> `AzksElementSet`
- `NodeType` -> `TreeNodeType`
- `commitment_proof` -> `commitment_nonce`
- `next_few_vrf_proofs` -> `until_marker_vrf_proofs`
- `non_existence_of_next_few` -> `non_existence_until_marker_proofs`
- `non_existence_of_future_markers` -> `non_existence_of_future_marker_proofs`
- `previous_version_stale_at_ep` -> `previous_version_proof`
- `existence_at_ep` -> `existence_proof`
- `plaintext_value` -> `value`
- `LayerProof` -> `SiblingProof`
- `get_marker_version()` -> `get_marker_version_log2()`
- `random_node()` -> `random_azks_element()`

Note that this will require a bump in version since it changes the protobuf files.

### DIFF
--- a/akd/benches/azks.rs
+++ b/akd/benches/azks.rs
@@ -11,7 +11,7 @@ extern crate criterion;
 use akd::append_only_zks::InsertMode;
 use akd::storage::manager::StorageManager;
 use akd::storage::memory::AsyncInMemoryDatabase;
-use akd::{Azks, Node, NodeLabel};
+use akd::{Azks, AzksElement, NodeLabel};
 use criterion::{BatchSize, Criterion};
 use rand::rngs::StdRng;
 use rand::{RngCore, SeedableRng};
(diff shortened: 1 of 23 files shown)


afterdusk:
We might need to rename variables as well e.g. `node_set` -> `azks_element_set` or `element_set`. If we're taking care of variables in a later part that's fine too!

slawlor:
Just asking the question, but do we really want to change the protobuf spec's? It kind of defeats the purpose of "backwards compat", and we're decoding to structs which can have whatever name we want. This just means changes of logic everywhere that these spec's appear (servers, clients, etc)

slawlor:
Discussed offline, it's the best time to do this for now, so let's do it now

slawlor (approved):

PR #338, opened 2023-01-06
Fixing history proof verification

### LINKED ISSUE #336: Bug in history proof verification
For key history proofs, the server should generate non-membership proofs for the epochs up until the next power of 2, and then for each power of two until the current epoch. Right now, the existing logic has some off-by-one errors which need to be fixed.

Example 1: Let’s say a user has a value that is on version 11, and the current epoch is 65.

Expected behavior: A key history proof for this value should include:
- For each i in the range [1, 11], a membership proof for version #i being marked as “fresh”
- For each i in the range [2, 11], a membership proof for version #i-1 being marked as “stale”
- For each i in the range [12, 15], a non-membership proof for version #i being marked as “fresh”
- For each i in the set {16, 32, 64}, a non-membership proof for version #i being marked as “fresh” (as a marker node)

Actual behavior: The current logic fails to provide a non-membership proof for version 16 due to an off-by-one error.

Example 2: Let’s say a user has a value that is on version 8, and the current epoch is 65.

Expected behavior: A key history proof for this value should include:
- For each i in the range [1, 8], a membership proof for version #i being marked as “fresh”
- For each i in the range [2, 8], a membership proof for version #i-1 being marked as “stale”
- For each i in the range [9, 15], a non-membership proof for version #i being marked as “fresh”
- For each i in the set {16, 32, 64}, a non-membership proof for version #i being marked as “fresh” (as a marker node)

Actual behavior: Actually, I think this one is correctly handled, but we should add a test case to double-check.

kevinlewi:
This will fix an error in the key history verification logic that was not caught before. I also am adding a test for these more complicated scenarios with multiple versions and epochs which would have caught this issue.

Closes #336

### DIFF
--- a/akd_core/src/verify/history.rs
+++ b/akd_core/src/verify/history.rs
@@ -126,7 +126,7 @@ pub fn key_history_verify(
     }
 
     // Verify the VRFs and non-membership proofs for future markers
-    for (i, pow) in (next_marker + 1..final_marker).enumerate() {
+    for (i, pow) in (next_marker..final_marker + 1).enumerate() {
         let ver = 1 << pow;
         let pf = &proof.non_existence_of_future_marker_proofs[i];
         let vrf_pf = &proof.future_marker_vrf_proofs[i];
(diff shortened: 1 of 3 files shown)


akd_core/src/verify/history.rs
```diff
@@ -126,7 +126,7 @@ pub fn key_history_verify(
     }
 
     // Verify the VRFs and non-membership proofs for future markers
-    for (i, pow) in (next_marker + 1..final_marker).enumerate() {
+    for (i, pow) in (next_marker..final_marker + 1).enumerate() {
```
slawlor, line 129: Was this the bug?

akd_core/src/verify/history.rs
```diff
@@ -191,54 +191,57 @@ fn verify_single_update_proof(
...
+    let previous_version_proof = proof.previous_version_proof.as_ref().ok_or_else(|| {
+        VerificationError::HistoryProof(format!(
+            "Staleness proof of user {:?}'s version {:?} at epoch {:?} is None",
+            uname,
+            (version - 1),
+            epoch
+        ))
+    })?;
+    // Check that the correct value is included in the previous stale proof
+    if merge_with_int(hash(&crate::EMPTY_VALUE), epoch) != previous_version_proof.hash_val {
```
slawlor, line 217: I see your point about pre-computing the hash of an empty value... lol

slawlor (approved):
<3!

akd_core/src/verify/history.rs
kevinlewi: Yep!

PR #340, opened 2023-01-09
Use dashmap for in-memory DB to improve audit verification perf

afterdusk:
# Overview
The audit verify flow relies on an in-memory DB without a transaction object to build up a tree. As the in-memory DB is a RwLocked hashmap, this results in significant contention during the newly parallelized batch insert process. By moving the in-memory DB to a dashmap, we minimize this contention and unlock significant perf gains for the audit verify flow.

# Benchmark

Ran the `azks.rs` benchmark and observed the `audit_verify` function, which performs a verification of a publish from epoch 1 (10K leaves) to epoch 2 (10K leaves).

```console
```

## Before
image

## After
image

### DIFF
--- a/akd/benches/azks.rs
+++ b/akd/benches/azks.rs
@@ -65,7 +65,7 @@ fn batch_insertion(c: &mut Criterion) {
     });
 }
 
-fn audit(c: &mut Criterion) {
+fn audit_verify(c: &mut Criterion) {
     let num_initial_leaves = 10000;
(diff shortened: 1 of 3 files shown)


kevinlewi:
Holy cow! Can we do the same for improving the performance of `Directory::audit` as well? Or is this limited to improving `audit_verify`?

kevinlewi (approved):

PR #341, opened 2023-01-10
Adding code coverage CI

kevinlewi:
Adding a new CI pipeline for measuring code coverage. Followed this guide here: https://blog.rng0.io/how-to-do-code-coverage-in-rust

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 .DS_Store
 .vs
 
+debug/
+coverage/
+
 .vscode/**
 !.vscode/launch.json
 !.vscode/settings.json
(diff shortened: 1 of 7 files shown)


.gitignore
```diff
@@ -1,6 +1,10 @@
 .DS_Store
 .vs
 
+debug/
+target/
```
slawlor, line 5: This won't be excluded already with **/target?

xtask/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
```
slawlor, line 1: This looks like a binary, do we need a `default-run` statement?

slawlor (approved):
Looks good, just a couple comments for running the coverage but this is helpful thanks!

.gitignore
kevinlewi: Ah, thanks!

PR #342, opened 2023-01-11
General cleanups and reorganization (part 2)

kevinlewi:
- Reorganizing Directory::publish VRF computations to iterate over the VRF map instead of the update set
- Refactored (but didn't change) logic for generating membership proofs, fixing a flaky get_sibling_node() test that I had previously introduced
- Made CI run benches with `--no-run` parameter so they just check for compilation errors

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -148,7 +148,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: bench
-          args: -F bench
+          args: -F bench --no-run
 
   docs:
     name: docs
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,7 @@
+ignore:
+  - "akd_local_auditor"
+  - "akd_mysql"
+  - "akd_test_tools"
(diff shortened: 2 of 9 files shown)


slawlor:
I think you need to tweak the benchmark by passing a slice of the vec (probably just adding a `&`). Looks like the rest of the CI is happy.

.github/workflows/ci.yml
```diff
@@ -148,7 +148,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: bench
-          args: -F bench
+          args: -F bench --no-run
```
slawlor, line 151: why are we disabling the benchmark running? I guess we just want to make sure they compile, but not actually measure in the CI?

kevinlewi: Right... just to try to improve the overall running time of the CI (so that it doesn't take forever)

slawlor: .. that doesn't seem right lol. 13min sounds like it's running the full benchmarks not just compiling the bench's. Let's see how the CI does on my pr

Issue #294, opened 2022-12-19
Remove all usages of .unwrap() in non-test code


kevinlewi:
_Originally posted by @kevinlewi in https://github.com/novifinancial/akd/pull/292#discussion_r1052487581_

slawlor:
Ah we actually want this! Returning a "Result" from test functions does not give you the line or stack of what failed in the test. We really want the `unwrap()` or ideally `expect()` statements to show us exactly where a test failed. _Ideally_ we shouldn't have any tests which actually return a `Result<_,_>` type and unwrap everything that should not fail.

kevinlewi:
@slawlor: To double-check, this issue is for removing usages of unwrap in **non-test** (production) code. Having unwraps in test code sounds fine by me!

slawlor:
> @slawlor: To double-check, this issue is for removing usages of unwrap in **non-test** (production) code, because they could cause un-catchable panics that abort the server. Having unwraps in test code sounds fine by me!

Ah wow.. My bad, I misread that!

slawlor:
A precursory search found a few

a. (1) in `akd/src/append_only_zks.rs`
b. (2) in `akd/src/ecvrf_impl.rs`
c. (4) in `akd_core/src/types/node_label/mod.rs`
d. (2) in `akd_mysql/src/mysql_storables.rs`
e. (3) in `akd_mysql/src/mysql.rs`

The rest I think are in tests from the looks of it. Looking at the usages a few are kind of logical, so i'm not sure we want to get rid of all of them. For example 
```rust
let bin = St::get_full_binary_key_id(key);
// Since these are constructed from a safe key, they should never fail
// so we'll leave the unwrap to simplify
let back: NodeKey =
        TreeNodeWithPreviousValue::key_from_full_binary(&bin).unwrap();
```

so like unless we screwed up our conversions, it shouldn't be possible... which if we screwed them up we probably want to panic here :p. I'll put up a PR cleaning up some of them however.

PR #343, opened 2023-01-12
Removing `unwrap()` calls in production code

slawlor:
This change removes a lot of calls to `unwrap()` in runtime code in order to minimize the risk of panic. We generally don't want to risk panicking in runtime, and want to return a graceful error when possible.

Related to #294

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,6 @@ jobs:
 
           - name: Test the base library, with default features
             package: akd
-            flags:
 
           - name: Test the base library, without parallelism
             package: akd
@@ -52,11 +51,9 @@ jobs:
 
           - name: Test the local auditor, with default features
             package: akd_local_auditor
-            flags:
 
           - name: Test the base client library, with default features
             package: akd_client
-            flags:
 
           - name: Test the client for wasm and SHA3-256 hashing
             package: akd_client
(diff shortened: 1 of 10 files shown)


akd_core/src/ecvrf/tests.rs
```diff
@@ -27,6 +27,14 @@ use rand::rngs::StdRng;
 use rand::{CryptoRng, RngCore};
 use serde::{Deserialize, Serialize};
 
+impl Clone for VRFPrivateKey {
+    fn clone(&self) -> Self {
+        // In theory, creating a key from bytes could be a DecodingError, except
+        // we just copied these bytes out of the source key, so ...
+        Self(ed25519_PrivateKey::from_bytes(self.0.as_bytes()).unwrap())
```
ragansa, line 34: Feels like an `expect("Copied bytes from source key should be valid")` would be appropriate?

slawlor:
@kevinlewi I think I found out what was up with the benchmarks. We are compiling all the libs into benchmark binaries even if they have no benchmarks. I tweaked the CI to just have 2 runs, benchmarks on `akd_core` and `akd` specifically, ignoring the other libs. That should speed things up a lot.

I think the bulk of it was all the UI libs included due to using `iced` in the auditor app 💀

.github/workflows/ci.yml
```diff
@@ -144,11 +141,17 @@ jobs:
...
         with:
           command: bench
-          args: -F bench --no-run
+          args: -p akd_core -F bench --no-run
+
+      - name: Run akd benches
+        uses: actions-rs/cargo@v1
+        with:
+          command: bench
+          args: -p akd -F bench --no-run
```
kevinlewi, line 154: Can we use a [matrix strategy](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) here to parallelize this? Similar to the matrix strategy used for the main tests.

kevinlewi (approved):
Awesome!

PR #346, opened 2023-01-13
Deterministic fixture generation

kevinlewi:
Making fixture generation be deterministic so that we can more easily test for deltas when re-running fixture generation.

This involves adding Ord to the objects that can be stored in a DbRecord, as well as using StdRng with a seed instead of OsRng.

Also moved to using StdRng for a bunch of other tests, and added a more complex publish test.


slawlor (approved):
LGTM

PR #347, opened 2023-01-14
Using AzksValue as node value instead of Digest

kevinlewi:
- Defines a wrapper struct `AzksValue` which is used to hold the node's value. Previously it was just a raw `Digest`
- `AzksElement` now has two fields, a label `NodeLabel` and a value `AzksValue`
- Also defines `AzksValueWithEpoch` to make the distinction of when the value is hashed together with the epoch for leaf nodes
- Also adds cargo doc build tests for each package individually (since akd_core was not building, see https://docs.rs/crate/akd_core/latest)
- Doesn't change any protobuf generation or the fixture

Btw, given this change, I am thinking of renaming the `NodeLabel` struct to `AzksLabel`. Thoughts? This would potentially affect the protobuf files.

### DIFF
--- a/akd/src/tree_node.rs
+++ b/akd/src/tree_node.rs
@@ -11,14 +11,14 @@ use crate::crypto::{
     compute_parent_hash_from_children, empty_node_hash, empty_root_value, hash_leaf_with_commitment,
 };
 use crate::errors::{AkdError, StorageError, TreeNodeError};
+use crate::hash::EMPTY_DIGEST;
 use crate::storage::manager::StorageManager;
 use crate::storage::types::{DbRecord, StorageType};
 use crate::storage::{Database, Storable};
-use crate::Digest;
+use crate::AzksValue;
 use crate::{node_label::*, Direction, EMPTY_LABEL};
-use akd_core::hash::EMPTY_DIGEST;
 #[cfg(feature = "serde_serialization")]
-use akd_core::utils::serde_helpers::{bytes_deserialize_hex, bytes_serialize_hex};
+use akd_core::utils::serde_helpers::{azks_value_hex_deserialize, azks_value_hex_serialize};
 use std::cmp::{max, min};
 use std::convert::TryInto;
 use std::marker::Sync;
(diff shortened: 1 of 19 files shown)


akd_core/Cargo.toml
```diff
@@ -33,8 +33,8 @@ parallel_vrf = ["tokio"]
 
 bench = ["parallel_vrf", "blake3", "vrf", "tokio/rt-multi-thread"]
 
-# Default features mix (no hash function is selected at this point)
-default = ["vrf"]
+# Default features mix (with blake3 as the default hash function)
+default = ["vrf", "blake3"]
```
slawlor, line 37: Hmm we used to have this, and i remember it breaking something (like running `akd` without blake3 hashing or something). But the CI is happy so clearly i'm wrong....

But perhaps some other updates fixed it. I think it had to do with like `akd` pulling in a different mix of functionality in the `dev` profile, but that would break the ci

akd_core/Cargo.toml
```diff
@@ -69,4 +69,4 @@ lto = true
 [[bench]]
 name = "parallel_vrfs"
 harness = false
-required-features = ["bench"]
\ No newline at end of file
+required-features = ["bench"]
```
slawlor, line 72: whoopsie :)

slawlor (approved):
LGTM

akd/src/tree_node.rs
```diff
@@ -505,23 +505,26 @@ pub(crate) enum NodeHashingMode {
...
+) -> AzksValue {
     match input {
         Some(child_state) => {
-            let mut hash = child_state.hash;
+            let mut value = child_state.hash;
             if let (TreeNodeType::Leaf, NodeHashingMode::WithLeafEpoch) =
                 (child_state.node_type, hash_mode)
             {
-                hash = hash_leaf_with_commitment(hash, child_state.last_epoch);
+                value = AzksValue(hash_leaf_with_commitment(value, child_state.last_epoch).0);
```
afterdusk, line 525: Semantically, this should be an `AzksValueWithEpoch` I think. If so, we have to make `AzksValue` an enum with two variants `WithEpoch` and `WithoutEpoch` (or something along the lines)

kevinlewi: True, I haven't decided what the best way to use the distinction is, yet. I mainly wanted to make a distinction between AzksValues that have had their epochs incorporated (which only happens for leaf nodes) and AzksValues that have not had their epochs incorporated (happens for non-leaf nodes).

PR #348, opened 2023-01-16
Using PrefixOrdering to replace invalid Directions

### LINKED ISSUE #314: Convert Invalid Directional Movements During Preloading to Compile-time Errors

kevinlewi:
Closes #314.

We introduce a `PrefixOrdering` which is like a `Direction` that can be `None`, and then remove the `None` option from a `Direction`. That way, we isolate the places where we check for invalid prefix ordering, and deal with only valid directions in tree-manipulating code.

### DIFF
--- a/akd_core/src/types/mod.rs
+++ b/akd_core/src/types/mod.rs
@@ -21,9 +21,9 @@ use crate::utils::serde_helpers::{
 use crate::ARITY;
 
 #[cfg(feature = "nostd")]
-use alloc::vec::Vec;
+use alloc::string::{String, ToString};
 #[cfg(feature = "nostd")]
-use alloc::{format, string::String};
+use alloc::vec::Vec;
 #[cfg(feature = "nostd")]
 use core::cmp::{Ord, Ordering, PartialOrd};
 #[cfg(feature = "rand")]
@@ -33,9 +33,6 @@ use std::cmp::{Ord, Ordering, PartialOrd};
 
 pub mod node_label;
 pub use node_label::*;
-
-/// The possible VALID child directions
-pub const DIRECTIONS: [Direction; 2] = [Direction::Left, Direction::Right];
 
 // ============================================
 // Traits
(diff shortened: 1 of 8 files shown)


akd_core/src/types/node_label/mod.rs
```diff
@@ -193,59 +203,15 @@ impl NodeLabel {
...
-    pub fn get_longest_common_prefix_and_dirs(&self, other: Self) -> (Self, Direction, Direction) {
-        let lcp_label = self.get_longest_common_prefix(other);
-        let dir_other = lcp_label.get_dir(other);
-        let dir_self = lcp_label.get_dir(*self);
-        (lcp_label, dir_other, dir_self)
-    }
-
-    /// Gets the direction of other with respect to self, if self is a prefix of other.
+    /// Gets the prefix ordering of other with respect to self, if self is a prefix of other.
     /// If self is not a prefix of other, then returns None.
```
slawlor, line 207: nit: comments clearer with `then returns [PrefixOrdering::Invalid]`

akd_core/src/types/mod.rs
```diff
@@ -77,40 +113,45 @@ pub enum Direction {
...
+            Bit::Zero => Self::Left,
+            Bit::One => Self::Right,
         }
     }
 }
 
+impl core::convert::TryFrom<PrefixOrdering> for Direction {
+    type Error = String;
+    fn try_from(prefix_ordering: PrefixOrdering) -> Result<Self, Self::Error> {
+        let direction = match prefix_ordering {
```
slawlor, line 138: nit: Generally early-return clauses are helpful in really long fn's but in this case it might be clearer as

```rust
match prefix_ordering {
    PrefixOrdering::WithZero => Ok(Direction::Left),
    PrefixOrdering::WithOne => Ok(Direction::Right),
    _ => Err("Could not convert from PrefixOrdering to Direction".to_string())
}
```

slawlor (approved):
A couple tiny nits, but the rest looks good to me

slawlor (approved):

PR #351, opened 2023-01-17
Optimizations to preload nodes in history proofs

slawlor:
We were doing a LOT of direct get requests to the data layer to retrieve the various nodes needed to do proof generation. This can be demonstrated with the benchmark added in this PR, and additionally artifically slowing the database's `get` call to simulate some network latency.

```bash
$ cargo bench -p akd --bench directory -F bench,slow_internal_db
...
Benchmark key history proof generation on a small tree
                        time:   [4.2915 s 4.3005 s 4.3096 s]
```

Then if we enable the support for preloading nodes in `akd/src/directory.rs` that this PR adds, we see that we are just calling batch_get which greatly improves our overall performance, even if batch-get has similar latency

```bash
$ cargo bench -p akd --bench directory -F bench,slow_internal_db,preload_history
...
Benchmark key history proof generation on a small tree
                        time:   [1.7625 s 1.7663 s 1.7700 s]
                        change: [-59.051% -58.929% -58.804%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Without any artifical db slowing we have the following:

No preload:
```bash
Benchmark key history proof generation on a small tree
                        time:   [23.138 ms 23.557 ms 23.977 ms]
```

With preload:
```bash
Benchmark key history proof generation on a small tree
                        time:   [23.798 ms 24.204 ms 24.608 ms]
```

### DIFF
--- a/akd/Cargo.toml
+++ b/akd/Cargo.toml
@@ -28,6 +28,10 @@ runtime_metrics = []
 parallel_vrf = ["akd_core/parallel_vrf"]
 # Parallelize node insertion during publish
 parallel_insert = []
+# Enable pre-loading of the nodes when generating history proofs
+preload_history = []
+# TESTING ONLY: Artifically slow the in-memory database (for benchmarking)
+slow_internal_db = []
 
 # Default features mix (blake3 + audit-proof protobuf mgmt support)
 default = ["blake3", "public_auditing", "parallel_vrf", "parallel_insert"]
@@ -71,5 +75,10 @@ akd = { path = ".", features = ["public-tests"], default-features = false }
 
 [[bench]]
 name = "azks"
+harness = false
+required-features = ["bench"]
+
+[[bench]]
+name = "directory"
 harness = false
 required-features = ["bench"]
--- a/akd/Cargo.toml
+++ b/akd/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "akd"
-version = "0.8.6"
+version = "0.8.7"
 authors = ["Harjasleen Malvai <hmalvai@fb.com>", "Kevin Lewi <klewi@fb.com>", "Sean Lawlor <seanlawlor@fb.com>"]
 description = "An implementation of an auditable key directory"
 license = "MIT OR Apache-2.0"
@@ -38,7 +38,7 @@ default = ["blake3", "public_auditing", "parallel_vrf", "parallel_insert"]
 
 [dependencies]
 ## Required dependencies ##
-akd_core = { path = "../akd_core", version = "0.8.5", default-features = false, features = ["vrf"] }
+akd_core = { path = "../akd_core", version = "0.8.7", default-features = false, features = ["vrf"] }
 async-recursion = "0.3"
 async-trait = "0.1"
 curve25519-dalek = "3"
--- a/akd/benches/directory.rs
+++ b/akd/benches/directory.rs
@@ -1,0 +1,86 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+//
+// This source code is licensed under both the MIT license found in the
+// LICENSE-MIT file in the root directory of this source tree and the Apache
+// License, Version 2.0 found in the LICENSE-APACHE file in the root directory
+// of this source tree.
+
+#[macro_use]
+extern crate criterion;
+
+use akd::ecvrf::HardCodedAkdVRF;
+use akd::storage::manager::StorageManager;
+use akd::storage::memory::AsyncInMemoryDatabase;
+use akd::{AkdLabel, AkdValue, Directory};
+use criterion::{BatchSize, Criterion};
+use rand::distributions::Alphanumeric;
+use rand::rngs::StdRng;
+use rand::{Rng, SeedableRng};
+
+fn history_generation(c: &mut Criterion) {
+    let num_users = 1000;
+    let num_updates = 10;
+    let runtime = tokio::runtime::Builder::new_multi_thread()
+        .enable_time()
+        .build()
+        .unwrap();
+
+    let idata = (1..num_users)
+        .into_iter()
+        .map(|i| {
+            let user = format!("User {}", i);
+            AkdLabel::from(&user)
+        })
+        .collect::<Vec<_>>();
+
+    let id = "Benchmark key history proof generation on a small tree".to_string();
+
+    c.bench_function(&id, move |b| {
+        b.iter_batched(
+            || {
+                let mut rng = StdRng::seed_from_u64(42);
+                let database = AsyncInMemoryDatabase::new();
+                let vrf = HardCodedAkdVRF {};
+                let db = StorageManager::new(
+                    database,
+                    Some(std::time::Duration::from_secs(60)),
+                    None,
+                    Some(std::time::Duration::from_secs(60)),
+                );
+                let db_clone = db.clone();
+                let directory = runtime
+                    .block_on(async move { Directory::new(db, vrf, false).await })
+                    .unwrap();
+
+                for _epoch in 1..num_updates {
+                    let value: String = (0..rng.gen_range(10, 20))
+                        .map(|_| rng.sample(&Alphanumeric))
+                        .map(char::from)
+                        .collect();
+                    let data = idata
+                        .iter()
+                        .map(|k| (k.clone(), AkdValue::from(&value)))
+                        .collect::<Vec<_>>();
+                    runtime.block_on(directory.publish(data)).unwrap();
+                }
+
+                (directory, db_clone)
+            },
+            |(directory, db)| {
+                // flush the cache prior to each generation to get fresh results
+                runtime.block_on(db.flush_cache());
+
+                // generate for the most recent 10 updates
+                let label = AkdLabel::from("User 1");
+                let params = akd::HistoryParams::MostRecent(5);
+                runtime
+                    .block_on(directory.key_history(&label, params))
+                    .unwrap();
+            },
+            BatchSize::PerIteration,
+        );
+    });
+}
+
+criterion_group!(directory_benches, history_generation);
+criterion_main!(directory_benches);
--- a/akd/src/directory.rs
+++ b/akd/src/directory.rs
@@ -369,6 +369,33 @@ impl<S: Database + 'static, V: VRFKeyStorage> Directory<S, V> {
         Ok((lookup_proofs, root_hash))
     }
 
+    async fn build_lookup_info(&self, latest_st: &ValueState) -> Result<LookupInfo, AkdError> {
+        let akd_label = &latest_st.username;
+        // Need to account for the case where the latest state is
+        // added but the database is in the middle of an update
+        let version = latest_st.version;
+        let marker_version = 1 << get_marker_version(version);
+        let existent_label = self
+            .vrf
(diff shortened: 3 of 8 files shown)


akd/src/directory.rs
```diff
@@ -464,6 +467,19 @@ impl<S: Database + 'static, V: VRFKeyStorage> Directory<S, V> {
             return Err(AkdError::Storage(StorageError::NotFound(msg)));
         }
 
+        #[cfg(feature = "preload_history")]
+        {
+            let mut lookup_infos = vec![];
+            for ud in user_data.iter() {
+                if let Ok(lo) = self.build_lookup_info(ud).await {
```
slawlor, line 474: @kevinlewi this *should* cover all of the nodes we'd need to walk in a history proof right? This already has the `ValueState`s so we should just be able to translate that to a `LookupInfo` which then lets us do a preload of the nodes we'd need to walk iiuc.

akd/src/directory.rs
```diff
@@ -838,7 +854,7 @@ impl<S: Database + 'static, V: VRFKeyStorage> Directory<S, V> {
             .map(|(akd_label, _val)| akd_label.clone())
             .collect();
         // sort the keys, as inserting in primary-key order is more efficient for MySQL
-        keys.sort_by(|a, b| a.cmp(b));
+        keys.sort();
```
afterdusk, line 857: Actually, do we still need this? This seems to be a database specific optimization, perhaps the sort could be moved into mysql data layer's `get_user_state_versions` (if we want to keep it at all)

akd/benches/directory.rs
```diff
@@ -0,0 +1,86 @@
...
+                        .collect::<Vec<_>>();
+                    runtime.block_on(directory.publish(data)).unwrap();
+                }
+
+                (directory, db_clone)
+            },
+            |(directory, db)| {
+                // flush the cache prior to each generation to get fresh results
+                runtime.block_on(db.flush_cache());
+
```
afterdusk, line 72: Ah we don't have to do this, the closure above is a setup function that is run once for each benchmark run. This means that a new directory and DB is created for each run of this closure.

akd/benches/directory.rs
```diff
@@ -0,0 +1,86 @@
...
+                for _epoch in 1..num_updates {
+                    let value: String = (0..rng.gen_range(10, 20))
+                        .map(|_| rng.sample(&Alphanumeric))
+                        .map(char::from)
+                        .collect();
+                    let data = idata
+                        .iter()
+                        .map(|k| (k.clone(), AkdValue::from(&value)))
+                        .collect::<Vec<_>>();
```
afterdusk, line 63: (Not a blocking suggestion) The benchmark might be slightly more performant if we precompute the updates outside of the setup closure

afterdusk (commented):
Overall LGTM, this is a very nice improvement! I'll leave this for Kevin to stamp :)

akd/Cargo.toml
```diff
@@ -73,3 +77,8 @@ akd = { path = ".", features = ["public-tests"], default-features = false }
 name = "azks"
 harness = false
 required-features = ["bench"]
+
+[[bench]]
+name = "directory"
+harness = false
+required-features = ["bench"]
```
kevinlewi, line 84: EOF newline

akd/Cargo.toml
```diff
@@ -28,13 +28,17 @@ runtime_metrics = []
 parallel_vrf = ["akd_core/parallel_vrf"]
 # Parallelize node insertion during publish
 parallel_insert = []
+# Enable pre-loading of the nodes when generating history proofs
+preload_history = []
+# TESTING ONLY: Artifically slow the in-memory database (for benchmarking)
+slow_internal_db = []
```
kevinlewi, line 34: Interesting idea! Do we want to keep this in here long-term as a toggle, or was this meant to be removed later?

kevinlewi (approved):
LGTM!

I think another thing that could greatly reduce the number of GETs would be to preload all of the nodes that need to be pulled in order to generate the history proof, by making functions like get_non_membership_proof() and get_membership_proof() to do preloading as well.

akd/src/directory.rs
slawlor: actually with the transaction ordering and everything i think this is defunct, it was a clippy lint update, but might as well just clean it up. Thanks!

akd/Cargo.toml
```diff
@@ -28,13 +28,17 @@ runtime_metrics = []
 parallel_vrf = ["akd_core/parallel_vrf"]
 # Parallelize node insertion during publish
 parallel_insert = []
+# Enable pre-loading of the nodes when generating history proofs
+preload_history = []
```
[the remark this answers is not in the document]
slawlor: remnant from benchmarking! will cleanup

akd/benches/directory.rs
slawlor: ah it didn't seem like it actually... without it it clearly looked like caching was taking over. I was getting nearly equal numbers for both benchmarks with and without preloading. Since it wasn't hitting my artificial slow-down

akd/Cargo.toml
slawlor: I debated actually, if we're doing optimizations it might be helpful to say minimize the number of db calls overall if they have a non-zero cost compared to caching and transaction I/O. I feel like this might be helpful in the future....

slawlor:
Current metric with "slow db" is 1.72s / proof. updated benchmark for faster speedup in data generation

eozturk1:
Could we get the database metrics to see whether we still have other single get requests so that we have a sense of there is more optimizations to be done here? (Feel free to merge this one. If there is room for improvement, we can create a separate task for it.)

slawlor:
> Could we get the database metrics to see whether we still have other single get requests so that we have a sense of there is more optimizations to be done here? (Feel free to merge this one. If there is room for improvement, we can create a separate task for it.)

Yeah so it's a bit of a hacky test, but i threw together something by editing the test `test_simple_key_history` which is generating a series of key history proofs. We start with

1. Editing the storage manager to have a cache
2. Flushing the cache after the writes (to simulate an empty cache)
3. Generating our proofs
4. Adding a `storage.log_metrics` at the end of the function then panicking lol

There's 2 GET requests which are constant from the Directory's azks retrieval and first root node retrieval ops, but the rest are batch-get's with 0 additional gets.

```text
00:00:00.034] INFO   Preload of tree (31 nodes) completed (append_only_zks:539)
[00:00:00.034] INFO   Key history preloaded 31 nodes (directory:477)
[00:00:00.077] INFO   Preload of tree (25 nodes) completed (append_only_zks:539)
[00:00:00.077] INFO   Key history preloaded 25 nodes (directory:477)
[00:00:00.101] INFO   Preload of tree (25 nodes) completed (append_only_zks:539)
[00:00:00.101] INFO   Key history preloaded 25 nodes (directory:477)
[00:00:00.120] INFO   Preload of tree (17 nodes) completed (append_only_zks:539)
[00:00:00.120] INFO   Key history preloaded 17 nodes (directory:477)
[00:00:00.136] ERROR  Cache hit since last: 604, cached size: 35 items (high_parallelism:62)
[00:00:00.136] ERROR  Transaction writes: 74, Transaction reads: 63 (transaction:88)
[00:00:00.136] ERROR  
    SET 2, 
    BATCH SET 6, 
    GET 2, 
    BATCH GET 8
    TOMBSTONE 0
    GET USER STATE 0
    GET USER DATA 4
    GET USER STATE VERSIONS 6
============ Database operation timing ============
    TIME READ 0 ms
    TIME WRITE 0 ms (manager:175)
```

Without preloads we have: 
```text
[00:00:00.107] ERROR  Cache hit since last: 506, cached size: 35 items (high_parallelism:62)
[00:00:00.107] ERROR  Transaction writes: 74, Transaction reads: 63 (transaction:88)
[00:00:00.107] ERROR  
    SET 2, 
    BATCH SET 6, 
    GET 37, 
    BATCH GET 0
    TOMBSTONE 0
    GET USER STATE 0
    GET USER DATA 4
    GET USER STATE VERSIONS 6
============ Database operation timing ============
    TIME READ 0 ms
    TIME WRITE 0 ms (manager:175)
```

eozturk1:
> > Could we get the database metrics to see whether we still have other single get requests so that we have a sense of there is more optimizations to be done here? (Feel free to merge this one. If there is room for improvement, we can create a separate task for it.)
> 
> Yeah so it's a bit of a hacky test, but i threw together something by editing the test `test_simple_key_history` which is generating a series of key history proofs. We start with
> 
> 1. Editing the storage manager to have a cache
> 2. Flushing the cache after the writes (to simulate an empty cache)
> 3. Generating our proofs
> 4. Adding a `storage.log_metrics` at the end of the function then panicking lol
> 
> There's 2 GET requests which are constant from the Directory's azks retrieval and first root node retrieval ops, but the rest are batch-get's with 0 additional gets.
> 
> ```
> 00:00:00.034] INFO   Preload of tree (31 nodes) completed (append_only_zks:539)
> [00:00:00.034] INFO   Key history preloaded 31 nodes (directory:477)
> [00:00:00.077] INFO   Preload of tree (25 nodes) completed (append_only_zks:539)
> [00:00:00.077] INFO   Key history preloaded 25 nodes (directory:477)
> [00:00:00.101] INFO   Preload of tree (25 nodes) completed (append_only_zks:539)
> [00:00:00.101] INFO   Key history preloaded 25 nodes (directory:477)
> [00:00:00.120] INFO   Preload of tree (17 nodes) completed (append_only_zks:539)
> [00:00:00.120] INFO   Key history preloaded 17 nodes (directory:477)
> [00:00:00.136] ERROR  Cache hit since last: 604, cached size: 35 items (high_parallelism:62)
> [00:00:00.136] ERROR  Transaction writes: 74, Transaction reads: 63 (transaction:88)
> [00:00:00.136] ERROR  
> ===================================================
> ============ Database operation counts ============
> ===================================================
>     SET 2, 
>     BATCH SET 6, 
>     GET 2, 
>     BATCH GET 8
>     TOMBSTONE 0
>     GET USER STATE 0
>     GET USER DATA 4
>     GET USER STATE VERSIONS 6
> ===================================================
> ============ Database operation timing ============
> ===================================================
>     TIME READ 0 ms
>     TIME WRITE 0 ms (manager:175)
> ```
> 
> Without preloads we have:
> 
> ```
> [00:00:00.107] ERROR  Cache hit since last: 506, cached size: 35 items (high_parallelism:62)
> [00:00:00.107] ERROR  Transaction writes: 74, Transaction reads: 63 (transaction:88)
> [00:00:00.107] ERROR  
> ===================================================
> ============ Database operation counts ============
> ===================================================
>     SET 2, 
>     BATCH SET 6, 
>     GET 37, 
>     BATCH GET 0
>     TOMBSTONE 0
>     GET USER STATE 0
>     GET USER DATA 4
>     GET USER STATE VERSIONS 6
> ===================================================
> ============ Database operation timing ============
> ===================================================
>     TIME READ 0 ms
>     TIME WRITE 0 ms (manager:175)
> ```

Methodology SGTM! Hmm.. it is odd that we don't have any GETs for the future proofs -- as in until the next marker. See https://github.com/facebook/akd/blob/main/akd/src/directory.rs#L511. Maybe we the last version is equal to the next marker in the test code? IIRC, the latest version is the number of times the key has been changed, so the proofs until the next marker isn't maybe too many?

PR #353, opened 2023-01-27
Adding parallelism to audit proof generation

slawlor:
Audit proofs don't currently support parallelization, and they're essentially a recursive tree-walk down the Merkle tree so are ideal candidates for parallel operations.

This change introduces parallel spawning of the audit proof walk by spawning a task for each `left_child` of the current node then recursing.

Additionally a (long) benchmark was added to assess this benefit, which will get progressively more important as the tree gets deep.

We've added a benchmark to demonstrate the functionality (with and without the `parallel_insert` feature enabled)

Without parallelism
```
Audit proof generation. 10000 leaves over 100 epochs
                        time:   [267.28 ms 271.38 ms 275.88 ms]
```

With:
```
Audit proof generation. 10000 leaves over 100 epochs
                        time:   [190.02 ms 192.30 ms 195.04 ms]
                        change: [-30.581% -29.141% -27.623%] (p = 0.00 < 0.05)
                        Performance has improved.
```

But again, these are tiny-scale tests compared to realistic data, which should continually get deeper and more complex as time goes on

### DIFF
--- a/akd/Cargo.toml
+++ b/akd/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "akd"
-version = "0.8.7"
+version = "0.8.8"
 authors = ["Harjasleen Malvai <hmalvai@fb.com>", "Kevin Lewi <klewi@fb.com>", "Sean Lawlor <seanlawlor@fb.com>"]
 description = "An implementation of an auditable key directory"
 license = "MIT OR Apache-2.0"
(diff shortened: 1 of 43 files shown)


slawlor:
New clippy lints required a LOT of changes... so that's the second commit lol

kevinlewi (approved):
LGTM!

PR #354, opened 2023-02-22
Add Convenience Methods to AuditBlobName

dillonrg:
- Adds an associated `new` fn to the `AuditBlobName` struct, along with a method which will construct an AuditBlob from an `AuditBlobName`, along with a reference to the appropriate proof. These methods aren't adding any new functionality to the struct, but are more so intended to serve as quality-of-life improvements for components of AKD.

### DIFF
--- a/akd/src/local_auditing.rs
+++ b/akd/src/local_auditing.rs
@@ -112,6 +112,32 @@ impl TryFrom<&str> for AuditBlobName {
     }
 }
 
+impl AuditBlobName {
+    /// Construct a new [AuditBlobName] from the arguments provided.
+    pub fn new(epoch: u64, previous_hash: Digest, current_hash: Digest) -> Self {
+        Self {
+            epoch,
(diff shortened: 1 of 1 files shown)


akd/src/local_auditing.rs
```diff
@@ -112,6 +112,32 @@ impl TryFrom<&str> for AuditBlobName {
...
+    /// Construct a new [AuditBlobName] from the arguments provided.
+    pub fn new(epoch: u64, previous_hash: Digest, current_hash: Digest) -> Self {
+        Self {
+            epoch,
+            previous_hash,
+            current_hash,
+        }
+    }
+
+    /// Consumes the [AuditBlobName] and transforms it into an [AuditBlob]
```
slawlor, line 125: nit: Consumes seems like a loaded description since it doesn't take ownership?

kevinlewi:
Closing as this is now fairly outdated

PR #355, opened 2023-03-14
Lookup proof optional optimization

slawlor:
In the case of a lookup proof, generally we're normally doing up to `N = depth(Merkle Tree)` requests to preload the nodes for the proof. However this is still like 30+ queries on moderately sized trees (e.g. ~2B nodes)

In this PR, we introduce an optional optimization which will build all 256 labels which could be on the path from the root node to the intended leaf and try a single batch-get with all of them.

A lookup proof's `LookupInfo` needs 3 target labels to retrieve, so we're actually doing 3 * 256 node retrievals, but in a single query which is still often more optimal due to DB round-trip accesses than 32 targeted node requests.

After the first direct-path load for the 3 necessary targets, we then do a second batch-get of all of those node's children (which will only load the found node's children that aren't already in-cache) which gives us all of the sibling nodes to generate the necessary proof.

Additionally this PR fixed a bug in batch-loading tree nodes, where instead of just not adding not-found results to the result set, it was crashing the batch-get with a `TreeNode not found` error. Our normal batch-get operations don't error on not-found results, so this was causing some flows to fail (incl a local test, which I'm surprised slipped through).

Lastly a small optimization as well, where in bulk lookup proofs, we were preloading multiple times. Once in the batch-operation, and then again for each individual proof which is just wasted cycles since it should all be cached anyways. This removed the second loop of preloads by passing a flag to the internal operation to skip the internal preload.

### DIFF
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -27,6 +27,7 @@ use crate::{
 use async_recursion::async_recursion;
 use log::info;
 use std::cmp::Ordering;
+use std::collections::HashSet;
 use std::convert::TryFrom;
 use std::marker::Sync;
 use std::ops::Deref;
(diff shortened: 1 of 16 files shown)


slawlor:
Added another optimization, where we can check the cache for nodes in the path which are already loaded, further reducing the number of nodes we're requesting to fetch. In the original greedy request, you can see we're trying to regularly load ~800+ nodes in our toy sample, where with this optimization it is only that value in the worst-case, where in actuality we're loading far less (just the full-path on the bottom of the tree)

Screen cap from (left) this new optimization and (right) just the original greedy fetch. This shows that as the proofs generate more and more proofs, we end up having to fetch less nodes meaning only lower-levels of the tree.

image

akd/src/append_only_zks.rs
```diff
@@ -1038,6 +1172,45 @@ mod tests {
...
+        };
+
+        let max_set = azks1
+            .build_lookup_maximal_node_set(&db, lookup_info)
+            .await
+            .expect("Failed to build maximal set");
+
+        // since the label is there 3 times, it should all resolve to the same data
+        assert_eq!(256, max_set.len());
+        println!("{max_set:?}");
```
kevinlewi, line 1211: nit: Remove this?

kevinlewi (approved):
Overall, LGTM!

Perhaps we should file an issue to keep track of the task to do this same optimization for history proofs?

Also, can this technique apply to building the audit proofs as well?

akd/src/append_only_zks.rs
slawlor: debugging :p

Issue #260, opened 2022-11-11
Change how read_only flag is exposed to consumer of a Directory


kevinlewi:
Right now, the read_only flag is specified in the constructor for a `Directory`, and the `publish` function returns an error when called on a read-only directory. A slightly cleaner way to organize this would be to create a `ReadOnlyDirectory` that is missing the publish function. That way we could get rid of the flag in the constructor, to slightly simplify the API for a consumer.

Edit: on second thought, let's just create a separate constructor (something like `new_read_only`) for creating read-only directories. That way we don't have to pass the bool into the main constructor.

PR #358, opened 2023-04-17
Change access mode parameter from bool to custom AccessMode enum

HiddenToad:
a common design pattern in rust is to replace boolean config parameters with named enums so that the intent of the configuration is clear at the call site.

previous api:
```
//writeable, can't tell what the bool means from code
Directory::new(db, vrf, false);

//read-only, same problem
Directory::new(db, vrf, true);
```
new api:
```
//intent now clear from code
Directory::new(db, vrf, AccessMode::Writeable);
Directory::new(db, vrf, AccessMode::ReadOnly);
```

this type is also re-exported in lib.rs, and can be `use`d if the user doesn't feel like typing the whole scoping prefix.

### DIFF
--- a/akd/benches/directory.rs
+++ b/akd/benches/directory.rs
@@ -11,7 +11,7 @@ extern crate criterion;
 use akd::ecvrf::HardCodedAkdVRF;
 use akd::storage::manager::StorageManager;
 use akd::storage::memory::AsyncInMemoryDatabase;
-use akd::{AkdLabel, AkdValue, Directory};
+use akd::{AkdLabel, AkdValue, Directory, AccessMode};
 use criterion::{BatchSize, Criterion};
 use rand::distributions::Alphanumeric;
 use rand::rngs::StdRng;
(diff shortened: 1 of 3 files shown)


kevinlewi:
Thanks for putting up this PR! This is related to https://github.com/facebook/akd/issues/260 -- thoughts on the approach described there, so that we can avoid having this third parameter bool / enum in the first place?

HiddenToad:
i definitely think creating a new type works if that's what you guys want! the enum is a good alternative if you want it in the form of a parameter so i came up with that, either works imo ¯\_ (ツ)_/¯

kevinlewi:
@HiddenToad let's go with the pseudo-overload for the constructor (imo the new type is probably overkill). That way we can get rid of the extra parameter entirely. Thanks in advance!

kevinlewi:
@HiddenToad I believe this is being addressed by #359 from @dillonrg -- can you take a look there and see if that solution resolves this issue? If so, we can land that one and I will close this PR.

PR #359, opened 2023-04-20
Add a Newtype Representative of a Read-only Directory

dillonrg:
- Prior to this patch, we were reliant upon a boolean parameter passed to the associated `new` fn on the `Directory` type to indicate whether the instance it produced should be treated as a read-only entity. To more explicitly represent a `Directory` which should be treated as a read-only entity, this patch introduces a [newtype](https://doc.rust-lang.org/rust-by-example/generics/new_types.html) known as `ReadOnlyDirectory` which acts as a simple wrapper for a `Directory`.
- To avoid confusion around whether or not a directory has the ability to publish, the `ReadOnlyDirectory` and `Directory` types do not share any traits. Instead, the `ReadOnlyDirectory` tpye wraps a `Directory` instance and essenitally calls into the `Directory` for read-only operations. As such, we get compile time guarantees that a `publish` operation is never possible via a `ReadOnlyDirectory` since the method simply doesn't exist.

### DIFF
--- a/akd/Cargo.toml
+++ b/akd/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "akd"
-version = "0.8.9"
+version = "0.9.0"
 authors = ["Harjasleen Malvai <hmalvai@fb.com>", "Kevin Lewi <klewi@fb.com>", "Sean Lawlor <seanlawlor@fb.com>"]
 description = "An implementation of an auditable key directory"
 license = "MIT OR Apache-2.0"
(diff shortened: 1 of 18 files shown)


slawlor:
Were you really out of date? Looks like a big rebase is necessary...

dillonrg:
> Were you really out of date? Looks like a big rebase is necessary...

Looks like it! I am rebasing now. I thought I had just pulled the latest before making changes, but looks like I didn't! Fixing now and will update.

slawlor (approved):
Just ignore the lint? Add an explicit exclusion

Change LGTM, but please bump the version of the crates in the Toml specs

PR #360, opened 2023-04-24
Addressing some FIXMEs and simplifying get_root_hash API

kevinlewi:
This PR goes through some of the remaining FIXMEs (except the ones in crypto.rs) and addresses them.
- Adding an error for `get_append_only_proof`
- Renaming the "hash" field to "value" on a TreeNode
- FIXME saying to get rid of a clone in verify_consecutive_append_only -- actually it's OK to clone here.
- Changed the empty label value to be a 1 followed by 31 0s... this allows for label compression to take effect and slightly shorten proofs (previously it was an array of 1s).
- Updated documentation for the TreeNode struct
- Instead of computing the hash for a NodeLabel, just use the raw value of the NodeLabel instead
- Removed the `serial` decorator for `test_mysql_db`

This also simplifies the API slightly for getting the root hash of a directory. The main function to call should be `Directory::get_root_hash_and_epoch()`, since the other functions require supplying an AZKS.

Eliminates the `test_empty_tree_root_hash()` test, because it is quite cumbersome to change this for every feature manually once the tree structure updates. And I don't think we really need this test anymore (previously it was introduced for consistency reasons).


kevinlewi:
FYI @afterdusk : In https://github.com/facebook/akd/pull/271 you introduced the `get_root_hash_safe` function, and in this PR I am getting rid of it in the public API and just encouraging the use of `Directory::get_epoch_hash()`. Does this sound OK to you?

afterdusk:
@kevinlewi That's ok, there's no use-case for the API at this time 🙂

akd/src/directory.rs
```diff
@@ -562,8 +571,8 @@ where
     /// the "latest" epoch. There is a "special" flow in the storage layer
     /// to do a storage-layer retrieval which ignores the cache
     ///
-    /// NOTE: Due to the use of std::thread::sleep(.) this will BLOCK
-    /// the polling thread, and should be allocated it's own thread since it won't
+    /// NOTE: Due to the use of [std::thread::sleep] this will BLOCK
```
slawlor, line 574: This is no longer accurate right? We're using tokio's sleep which does NOT block the thread, and therefore this can just be `spawn`'d

kevinlewi: Excellent! Removed the note.

akd/src/append_only_zks.rs
```diff
@@ -761,6 +760,16 @@ impl Azks {
...
     ) -> Result<AppendOnlyProof, AkdError> {
+        if self.get_latest_epoch() < end_epoch || end_epoch <= start_epoch {
+            return Err(AkdError::Directory(DirectoryError::InvalidEpoch(format!(
+                "Start epoch must be less than end epoch, and end epoch must be at most the latest epoch. \
+                Start epoch: {}, end epoch: {}, latest_epoch: {}.",
+                start_epoch,
+                end_epoch,
+                self.get_latest_epoch(),
+            ))));
+        }
```
dillonrg, line 771: To avoid future clippy warnings, we should use inline these. For example:

```rust
        let latest_epoch = self.get_latest_epoch();
        if latest_epoch < end_epoch || end_epoch <= start_epoch {
            return Err(AkdError::Directory(DirectoryError::InvalidEpoch(format!(
                "Start epoch must be less than end epoch, and end epoch must be at most the latest epoch. \
                Start epoch: {start_epoch}, end epoch: {end_epoch}, latest_epoch: {latest_epoch}."
            ))));
        }
```

kevinlewi: Ah, cool -- thanks!

slawlor (approved):
LGTM

dillonrg (approved):

PR #362, opened 2023-04-25
Clarifying the dual-license

### LINKED ISSUE #357: Lincense terms are unclear
The various `Cargo.toml` files throughout this tree give a license of "MIT OR Apache-2.0". However, the source files themselves give terms of:
```
// This source code is licensed under both the MIT license found in the
// LICENSE-MIT file in the root directory of this source tree and the Apache
// License, Version 2.0 found in the LICENSE-APACHE file in the root directory
// of this source tree.
```
listing that both licenses apply, an "AND" clause. If your intent is to allow either license to be used, I would suggest wording it as such in the individual-file licenses. With the AND clause, both licenses would apply, and given the additional restrictions on the Apache license, the terms of the MIT license would largely be irrelevant.

In addition, the `CONTRIBUTING.md` states that contributions must comply with the non-existent `LICENSE` file in the project. I suspect this is from a template and not corrected for this project.

kevinlewi:
Closes #357

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,5 +49,7 @@ disclosure of security bugs. In those cases, please go through the process
 outlined on that page and do not file a public issue.
 
 ## License
+
 By contributing to akd, you agree that your contributions will be
-licensed under the LICENSE file in the root directory of this source tree.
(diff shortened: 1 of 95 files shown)


slawlor (approved):
Thanks!

Issue #368, opened 2023-05-16
Improving the "Overview" documentation to specify what AKD does not support 


divyatalesra-2001:
Will expand upon the second paragraph under "Overview" in https://docs.rs/akd/latest/akd/. One, we aren’t just querying entries, we are querying mappings from labels to values in the database. And two, that this isn’t enough for a full key transparency solution. We still need a way for clients to know that they got the same commitment for the database, but this is outside of the scope of the AKD library and must be handled separately.

PR #369, opened 2023-05-17
Updating overview documentation

divyatalesra-2001:
Closes https://github.com/facebook/akd/issues/368

### DIFF
--- a/akd/src/lib.rs
+++ b/akd/src/lib.rs
@@ -18,8 +18,14 @@
 //! Keys can also be updated to be associated with different values. Each batch of updates to these key-value
 //! mappings are associated with an epoch along with a commitment to the database of entries at that point in time.
 //! The server that controls the database can use this library to generate proofs of inclusion to clients that wish
-//! to query entries in the database. These proofs can be _verified_ by a client against the corresponding
-//! commitment to the database. We can think of this data structure intuitively as a _verifiable dictionary_.
+//! to query for keys in the database to retrieve their associated values. These proofs can be _verified_ by a client
+//! against the corresponding commitment to the database. We can think of this data structure intuitively as a
+//! _verifiable dictionary_.
+//!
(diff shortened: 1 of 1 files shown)


akd/src/lib.rs
```diff
@@ -18,8 +18,12 @@
 //! Keys can also be updated to be associated with different values. Each batch of updates to these key-value
 //! mappings are associated with an epoch along with a commitment to the database of entries at that point in time.
 //! The server that controls the database can use this library to generate proofs of inclusion to clients that wish
-//! to query entries in the database. These proofs can be _verified_ by a client against the corresponding
+//! to query entries in the database. Note that we are not just querying entries; we are querying mappings from labels
+//! to values in the database. These proofs can be _verified_ by a client against the corresponding
```
kevinlewi, line 22:
```suggestion
//! These proofs can be _verified_ by a client against the corresponding
```

akd/src/lib.rs
```diff
@@ -18,8 +18,12 @@
...
 //! commitment to the database. We can think of this data structure intuitively as a _verifiable dictionary_.
+//! Also, this isn’t enough for a full key transparency solution. We still need a way for clients to know
```
kevinlewi, line 24: First, let's leave a newline between the previous text and this text, so as to represent a logical separator of content.

Then, let's change the text to say:

"This library can be used as part of a key transparency system to generate commitments and serve proofs for a database of public keys. However, note that a full key transparency solution still needs to provide a way for clients to ensure that they are receiving the same commitment for each database epoch. This is outside of the scope of this library and must be handled separately."

kevinlewi (changes requested):
Overall looks good! Let's link to the created issue by saying "Closes #368" in the description of the PR. And here is some alternate text that I would suggest...

kevinlewi (approved):
LGTM!

Issue #371, opened 2023-05-30
Added tests for auditor.rs


kevinlewi:
According to https://app.codecov.io/gh/facebook/akd/blob/main/akd/src/auditor.rs, there are three error cases that we don't have test coverage for. We should add them in.

PR #373, opened 2023-06-05
Adding invalid_audit_verification test cases for increased coverage

divyatalesra-2001:


divyatalesra-2001:
Closing #371

kevinlewi:
Thanks! There is a third error condition to trigger in tests: https://github.com/facebook/akd/blob/main/akd/src/auditor.rs#L74-L76 can you add a test for that one as well?

kevinlewi (changes requested):
Otherwise looks good

kevinlewi (approved):
Great, thanks!

Issue #363, opened 2023-04-29
Crypto Configuration / CipherSuite Trait


kevinlewi:
Right now, we have VRFs which can be instantiated and supplied as a generic for a `Directory`, and we use feature-gating for selecting the underlying hash function.

In the future, I'd like to also introduce more configurability for the cryptographic operations in AKD -- for instance, the function which derives a parent node's root hash from its children, how we produce commitments to the AKD values, etc.

Most of these operations are contained in this file for now: https://github.com/facebook/akd/blob/main/akd_core/src/crypto.rs but in order to support more flexibility for other implementations of these functions, I'm envisioning that we can try to bundle these implementations into some sort of crypto configuration trait or ciphersuite syntax.

Examples:

Introduce a new trait, something like:

```
trait CipherSuite { 
  type VRF: VRF,
  type Hash: Hash,
  type TreeHash: TreeHash, // would contain a bunch of functions that are currently in akd_core/src/crypto.rs right now, probably needs a better name
}
```

And then, an instantiation of this trait could be specified when initializing a directory, rather than passing in a VRF and feature-gating with a hash function like we do now. So for instance, we could define

```
struct WhatsAppConfiguration;
impl CipherSuite for WhatsAppConfiguration {
  type VRF: HardCodedAkdVrf,
  type Hash: SHA256,
  type TreeHash: ...,
}
```

with the idea that consumers which would like to re-use these configurations could do so with ease (we could make WhatsAppConfiguration the default for instance). But also if they want to mix and match other configurations, they could simply define their own structs and `impl`s on the `CipherSuite` trait.

That way, we also wouldn't have to feature-gate on the hash function, and we can essentially consolidate all of these configurations into one bundle-able trait.

Thoughts? ccing @slawlor and @dillonrg for feedback

PR #375, opened 2023-06-07
Adding a tree configuration for cryptographic operations

kevinlewi:
Addresses https://github.com/facebook/akd/issues/363

Also updates Rust edition from 2018 -> 2021

This introduces the `Configuration` trait, along with two configurations: `WhatsAppV1Configuration` and `ExperimentalConfiguration`. The idea is that new updates to the configuration should be bundled into `ExperimentalConfiguration`, in a way that does not interfere with the existing functionality that we support.

All tests are run on both configurations. If we want to support more configurations in the future, we can grep for the string `Note(new_config)` which denotes places where code needs to be modified to support new configurations.

This configuration trait also generalizes the hash function. As a result, I am removing the specifying of hash functions via features to simplify things.


akd/Cargo.toml
```diff
@@ -4,24 +4,20 @@ version = "0.9.0-pre.1"
...
+whatsapp_v1 = ["akd_core/whatsapp_v1"]
+experimental = ["akd_core/experimental"]
 
-bench = ["blake3", "public_tests","tokio/rt-multi-thread"]
-public_tests = ["rand", "bincode", "colored", "once_cell", "serde_serialization", "akd_core/rand"]
-public_auditing = ["protobuf", "akd_core/protobuf"]
-serde_serialization = ["serde", "ed25519-dalek/serde", "akd_core/serde_serialization"]
+bench = ["experimental", "public_tests", "tokio/rt-multi-thread", "dep:paste"]
+public_tests = ["dep:rand", "dep:bincode", "dep:colored", "dep:once_cell", "serde_serialization", "akd_core/rand", "dep:paste"]
+public_auditing = ["dep:protobuf", "akd_core/protobuf"]
```
slawlor, line 19: Do you have documentation on this `dep:` syntax? I've never seen this

akd_core/src/lib.rs
```diff
@@ -182,12 +182,21 @@ extern crate alloc;
...
 pub mod hash;
 pub mod utils;
 pub mod verify;
 
+pub mod configuration;
+pub use configuration::Configuration;
+
+// Note(new_config): Update this when adding a new configuration
+
+#[cfg(feature = "experimental")]
```
slawlor, line 195: rather than importing it as a macro or something, do we just want to do

```rust
#[cfg(feature = "experimental")]
pub use configuration::experimental::ExperimentalConfiguration as TheConfiguration;
#[cfg(feature = "whatsapp_v1")]
pub use configuration::whatsapp_v1::WhatsAppV1Configuration as TheConfiguration;
```

Then we can just refer to `akd_core::TheConfiguration` everywhere, which will map to the right type

akd_core/src/configuration/mod.rs
```diff
@@ -0,0 +1,23 @@
...
+// of this source tree. You may select, at your option, one of the above-listed licenses.
+
+//! Defines the configuration trait and implementations for various configurations
+
+mod traits;
+pub use traits::{Configuration, NamedConfiguration};
+
+// Note(new_config): Update this when adding a new configuration
+
+#[cfg(feature = "whatsapp_v1")]
```
slawlor, line 15: Even better than doing it at the crate root, import it under the common name here with the `as` syntax so people only need to update the configuration in 1 place and it'll take effect everywhere.

akd_core/src/hash/mod.rs
```diff
@@ -17,36 +17,8 @@ use alloc::string::String;
...
-// ========== Sha3 settings ===============
-// =========================================
-#[cfg(feature = "sha3")]
-pub mod sha3;
-#[cfg(feature = "sha3")]
-pub use crate::hash::sha3::hash;
-#[cfg(feature = "sha3")]
-pub use crate::hash::sha3::DIGEST_BYTES;
+/// The number of bytes in a digest
+pub const DIGEST_BYTES: usize = 32;
```
slawlor, line 21: OK so we're excluding 64-byte (or custom byte length) hashes with this? We're cool with that? (I'm find with it, just want to be sure)

akd_core/src/configuration/traits.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    /// Computes the node label value from the bytes of the label
+    fn compute_node_label_value(bytes: &[u8]) -> Vec<u8>;
+
+    /// Returns the representation of the empty label
+    fn empty_label() -> NodeLabel;
+}
+
+/// For fixture generation / testing purposes only
+pub trait NamedConfiguration: Configuration {
```
slawlor, line 91: if the comment is true, do we want to gate it under a feature specification? (`public-tests`?)

akd_core/benches/parallel_vrfs.rs
```diff
@@ -9,13 +9,68 @@
...
 
-fn bench_single_vrf(c: &mut Criterion) {
+macro_rules! bench_config {
+    ( $x:ident ) => {
+        paste::paste! {
+            // NOTE(new_config): Add a new configuration here
+
+            #[cfg(feature = "whatsapp_v1")]
+            fn [<$x _ whatsapp_v1_config>](c: &mut Criterion) {
+                $x::<akd_core::WhatsAppV1Configuration>(c)
```
slawlor, line 26: yeah so here we could get rid of the features bit (I think) if you export a common `akd_core::TheConfiguration` or something

akd_client/src/wasm.rs
```diff
@@ -137,10 +139,51 @@ pub fn lookup_verify(
     }
 }
 
+/// NOTE(new_config): Add a new configuration here
+
+/// Verify a lookup proof in WebAssembly for WhatsAppV1Configuration,
+/// utilizing serde serialized structure for the proof
+#[cfg(feature = "whatsapp_v1")]
```
slawlor, line 146: same all this can be cleaned up. I would still add a feature onto the `akd_client` crate, but it would just enable the downstream feature on `akd_core` without needing to be checked anywhere in this crate.

akd/src/local_auditing.rs
```diff
@@ -212,7 +212,7 @@ mod tests {
 
         let blob_name = AuditBlobName {
             current_hash: crate::hash::EMPTY_DIGEST,
-            previous_hash: [1u8; crate::hash::DIGEST_BYTES],
+            previous_hash: [1u8; crate::hash::EMPTY_DIGEST.len()],
```
slawlor, line 215: nit: why not use the `DIGEST_BYTES`? I see you removed the import, but you could still read it from `akd_core`. It seems an odd change to now rely on the output of a fn which isn't a const to build an array.

slawlor (changes requested):
OK so I guess there's an open question here.

Do we expect a user of this crate, which wants to do their own "configuration", to put it into `akd_core`? Technically since everything's a generic, you can rely on `akd_core` without enabling any of the features (except for "vrf"). I.e.

```
$ cargo build -p akd_core --no-default-features --features vrf
```
compiles. 

That begs the question of should we have these configuration elements defined in `akd_core` at all (the traits yes, but the actual implementations should maybe live elsewhere?) If we otherwise expect people to include them in `akd_core` and re-compile the crate, then I'd lobby for making a common export from `akd_core` based on the featues enabled. I.e. something like

```rust
#[cfg(feature = "experimental")]
pub use configuration::experimental::ExperimentalConfiguration as AkdConfiguration;
#[cfg(feature = "whatsapp_v1")]
pub use configuration::whatsapp_v1::WhatsAppV1Configuration as AkdConfiguration;
#[cfg(feature = "fancy")]
pub use configuration::fancy::FancyConfiguration as AkdConfiguration;
```

so then everywhere you can just use ```akd_core::AkdConfiguration``` based on the compilation flags, and you'd have one. This should make the need for all the test macros etc be removable. If we go this route, you could ALSO remove the generics from `akd::Directory`, since you'd have a compile-time guarantee on the configuration type (since `akd_core` could only export 1). 

Happy to have a longer discussion on this, but we should identify the path forward and adjust this PR imho.

akd/Cargo.toml
kevinlewi: See https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies -- without the dep, any dependency marked as "optional" is automatically a toggle-able feature. By using dep, we are saying that it is an optional dependency that should not be a feature.

akd_core/src/lib.rs
kevinlewi: See my above comment on features needing to be additive (so that enabling any combination of features should always compile / not break things)

akd_core/src/hash/mod.rs
kevinlewi: Yes, I think we don't have a use-case for 64 bytes at the moment, so I opted to make this simplification.

akd_core/src/configuration/mod.rs
kevinlewi: See my above comment on features needing to be additive (so that enabling any combination of features should always compile / not break things)

akd_core/src/configuration/traits.rs
kevinlewi: Good idea!

akd/src/local_auditing.rs
kevinlewi: Ah, good point, I think this was something I failed to clean up.

kevinlewi:
That's a good question. I am ultimately aiming for us to have a set of configurations that we "officially" support in AKD, and consumers of the crate that would like to use alternative configurations can do so by implementing the trait.

From the rust book on features (https://doc.rust-lang.org/cargo/reference/features.html):
> A consequence of this is that features should be additive. That is, enabling a feature should not disable functionality, and it should usually be safe to enable any combination of features. 

So I think that with your proposal of using features to toggle what ends up being `AkdConfiguration`, it would violate this additive rule.

slawlor:
Fair point, it just seems like a lot of bloat but yes features would no longer be additive. Oh well, but this looks good then

PR #382, opened 2023-07-02
Updating some dependencies

kevinlewi:
Just going through and updating some outdated dependencies. Updating `serde_yaml` from `0.8` to `0.9` caused some changes in serialization, so I re-ran the fixture generation.


slawlor (approved):
<3

PR #386, opened 2023-07-21
Folding akd_mysql, integration_tests, and akd_test_tools into examples

kevinlewi:
This moves all of the code from akd_mysql, integration_tests, and akd_test_tools into the examples directory.

Now, we have three utilities in the examples/ directory: whatsapp_kt_auditor, mysql_demo, and fixture_generator

Once this is landed, I will yank the akd_mysql crate from crates.io


slawlor (approved):
LGTM

PR #387, opened 2023-07-23
Adding domain separation for ExperimentalConfiguration

kevinlewi:
Typically it is considered safer to have "domain separation" for hashing functions -- that way, hashes for one instance of AKD don't conflict with another application's AKD usage.

This should be mostly avoided by the fact that each application should sample their own VRF secrets, so this might not be super necessary for security. But in case applications reuse VRF secrets, or there are some other subtle attacks that can still arise despite having different VRF secrets, domain separation for the hash function should address all of these.

To set the domain separator label, an application can define a struct that implements the DomainLabel trait (we have an example called ExampleLabel). Then, the configuration can be used as ExperimentalLabel<ExampleLabel> which will ensure that the domain label string is embedded into each hash computation.

Depends on #389, and I will rebase on top of that once it is landed.


slawlor (approved):
This would be considered a breaking change right? Since we're changing how hashing works (from what WA is currently using)

kevinlewi:
Yep, technically a breaking change. But the changes are limited to `ExperimentalConfiguration` and not `WhatsAppV1Configuration`, so as long as we stay on the ``WhatsAppV1Configuration` version, then actually it would not be breaking.

I rebased on top of #389 after it landed.

slawlor:
LGTM, feel free to merge :)

Issue #388, opened 2023-07-23
Consider dropping support for the akd_client library


kevinlewi:
Or at least moving it under akd_core/ (without the wasm bindings). The reason being that the library is lightweight enough on its own (really just a cosmetic wrapper around akd_core), and once we start adding more customizations like domain separation based on application, the likelihood that someone will want to use akd_client as-is diminishes.

slawlor:
Yeah we could drop it, we'll just need to move some documentation around on how to use `akd_core` as it sits now, it's meant to not be publicly interact-able (i.e. it was supposed to be pretty internal). `akd_client` however does seem to re-export everything, so I agree it's value is limited. We could also move the wasm bindings to an example (unpublished) crate.

PR #389, opened 2023-07-27
Moving akd_client into examples/

kevinlewi:
Closes https://github.com/facebook/akd/issues/388

Note that we have renamed `akd_client` to `wasm_client`, and that this is not a typical "runnable" example, instead it is just meant to be a library that shows a developer which functions to expose to a client and how the wasm bindings could look.


slawlor (approved):
LGTM

PR #392, opened 2023-08-15
Adding lookup verify check for version number exceeding epoch

kevinlewi:
Closes https://github.com/facebook/akd/issues/391.

This check is necessary, since key history proofs only verify versions up until the current epoch, and so it is up to the corresponding lookup proof verifications to ensure that version numbers don't exceed the current epoch when returned.


slawlor (approved):
LGTM

PR #394, opened 2023-08-21
Prepare for 0.9.0 release

slawlor:
Prepares versions for 0.9.0 release. Also removes the version number on the examples crate since it's a non-published crate, it doesn't need to be versioned

### DIFF
--- a/akd/Cargo.toml
+++ b/akd/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "akd"
-version = "0.9.0-pre.1"
+version = "0.9.0"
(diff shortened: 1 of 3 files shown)


kevinlewi:
Replacing this with #395 

Also, note sure why the whatsapp-kt-auditor test failed. Ran it locally:

```
[00:00:10] Loaded epochs between (1) and (47140), inclusively.
[00:00:11] Successfully downloaded proof for epoch 47140.
[00:02:20] Audit proof for epoch 47140 has verified successfully!
```
Maybe a temporary network issue?

PR #396, opened 2023-08-21
Fixing publish script hardcoding blake3 feature

kevinlewi:
Publish action was failing due to no-longer-existing blake3 feature that I forgot to delete in the publish script


slawlor (approved):
Lgtm

PR #400, opened 2023-09-02
Adding duplicate entries check in publish

kevinlewi:
Ensures that duplicate label entries cannot be submitted to Directory::publish successfully

### DIFF
--- a/akd/src/directory.rs
+++ b/akd/src/directory.rs
@@ -22,7 +22,7 @@ use crate::{
 use crate::VersionFreshness;
 use akd_core::configuration::Configuration;
 use log::{error, info};
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::marker::PhantomData;
 use std::sync::Arc;
 use tokio::sync::RwLock;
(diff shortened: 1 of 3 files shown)


akd/src/directory.rs
```diff
@@ -92,11 +95,21 @@ where
         let current_epoch = current_azks.get_latest_epoch();
         let next_epoch = current_epoch + 1;
 
-        let mut keys: Vec<AkdLabel> = updates
-            .iter()
-            .map(|(akd_label, _val)| akd_label.clone())
-            .collect();
+        let updates_iter = updates.iter().map(|(akd_label, _val)| akd_label.clone());
```
slawlor, line 98: I think it might be more idiomatic to do

```rust
let distinct_set = updates
    .iter()
    .map(|(label, _)| label.clone())
    .collect::<HashSet<_>>()

if distinct_set.len() != updates.len() {
    return Err(...);
}
```

akd/src/tests.rs
```diff
@@ -1278,6 +1278,38 @@ async fn test_lookup_verify_invalid_version_number<TC: Configuration>() -> Resul
...
+    }
+
+    // Now add a duplicate entry
+    updates.push(updates[0].clone());
+
+    // Attempt to publish -- this should throw an error because of the duplicate entry
+    let akd_result = akd.publish(updates).await;
+
+    // Check that the result is a verification error
+    match akd_result {
```
slawlor, line 1305: since it's a match of a single element, if-let-else?

kevinlewi: Syntax looks a little strange to me:
```
    if let Err(AkdError::Directory(DirectoryError::Publish(_))) = akd.publish(updates).await {
        ()
    } else {
        panic!("Expected a directory publish error");
    }
```
is this what you meant? Fine by me but just wanted to double-check.

slawlor: Ah ya I see what you mean. You could also do let-else so something like.

let _ = ...match... else {
  panic!()
};

slawlor:
Lgtm

slawlor (approved):
LGTM

PR #401, opened 2023-09-02
Updating VRF encode_to_curve checks

kevinlewi:
- Renames hash_to_curve to encode_to_curve
- Checks for identity point before returning encode_to_curve output
- Also checks for point being >= p
- Updates VRF draft references to point to RFC9381


kevinlewi:
Clippy errors got fixed in https://github.com/facebook/akd/pull/401 and should disappear after rebasing on top of that PR

akd_core/src/ecvrf/ecvrf_impl.rs
```diff
@@ -231,18 +233,41 @@ impl VRFPublicKey {
...
+/// As defined in [Section 5.1.3 of RFC8032](https://www.rfc-editor.org/rfc/rfc8032#section-5.1.3)
+///
+/// Will return Some(point) if the hash value can be interpreted as a point, and None otherwise.
+fn interpret_hash_value_as_a_point(hash: [u8; 32]) -> Option<EdwardsPoint> {
+    // If the input bytes are such that bytes 1 to 30 have value 255, byte 31 has value 255 or 127,
+    // and byte 0 has value 256 - i for value i in the (1, 3, 4, 5, 9, 10, 13, 14, 15, 16) list, then
+    // the encoding is invalid.
+    let is_invalid = hash[1..=30].iter().all(|b| *b == 255)
+        && (hash[31] == 255 || hash[31] == 127)
+        && [1, 3, 4, 5, 9, 10, 13, 14, 15, 16]
```
slawlor, line 260: nit: instead of having to build an iterator, is it cleaner as

```rust
[1u8,3,4,5,9,10,13,14,15,16].contains(&(256-hash[0]))
```
?

akd_core/src/ecvrf/ecvrf_impl.rs
```diff
@@ -231,18 +233,41 @@ impl VRFPublicKey {
...
+            .iter()
+            .any(|i| hash[0] == (256 - i) as u8);
+    if is_invalid {
```
slawlor, line 263: nit: Given it's the last statement of the function, not sure the early return makes it easier to read? Feel free to ignore this one, might just be me

```rust
if is_invalid {
  None
} else {
  CompressedEdwardsY::from_slice(&hash).expect("...").decompress()
}
```

slawlor (approved):
Just a couple nits, but looks good!

akd_core/src/ecvrf/ecvrf_impl.rs
kevinlewi: Hmm, yeah, style-wise I prefer not typing out the "else" and instead typing out the "return", I guess

akd_core/src/ecvrf/ecvrf_impl.rs
kevinlewi: Nice, good idea!

PR #402, opened 2023-09-03
Fixing clippy errors

kevinlewi:
Clippy errors getting fixed by this commit:
```
error: private item shadows public glob re-export
Error:    --> akd/src/lib.rs:486:1
    |
486 | mod utils;
    | ^^^^^^^^^^
    |
note: the name `utils` in the type namespace is supposed to be publicly re-exported here
   --> akd/src/lib.rs:483:9
    |
483 | pub use akd_core::*;
    |         ^^^^^^^^^^^
note: but the private item here shadows it
   --> akd/src/lib.rs:486:1
    |
486 | mod utils;
    | ^^^^^^^^^^
    = note: `-D hidden-glob-reexports` implied by `-D warnings`
```

### DIFF
--- a/akd/src/directory.rs
+++ b/akd/src/directory.rs
@@ -255,7 +255,7 @@ where
 
     /// Generate a lookup proof with the provided target information
     ///
-    /// * `current_azks`: The current [Akzs] element
+    /// * `current_azks`: The current [Azks] element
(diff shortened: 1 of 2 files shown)


slawlor (approved):
hmm odd lint... but sure clippy :)

PR #409, opened 2023-09-06
Reducing visibility of membership verification functions

kevinlewi:
Puts the verify_membership and verify_nonmembership functions behind the "public_tests" feature and also annotates them to make it more clear that they should only be called directly in test code.


slawlor (approved):
LGTM

PR #410, opened 2023-09-06
Adding public key validation documentation for ecvrf

kevinlewi:
Added docs to the verify() function for clarifying that public key validation occurs either when deriving from a VRFPrivateKey, or in the TryFrom for VRFPublicKey.


slawlor (approved):
LGTM

PR #415, opened 2023-09-21
Fixing publish workflow

kevinlewi:
Seems like https://github.com/facebook/akd/pull/406 needed to also specify the rust toolchain in the workflow


slawlor (approved):

Shadow666gamer:
> Seems like #406 needed to also specify the rust toolchain in the workflow

PR #417, opened 2023-10-12
Adding key history validation checks

kevinlewi:
Adds error-handling for various edge-cases when performing key history verification


slawlor (approved):
LGTM!

PR #422, opened 2024-01-14
Extending limited history to properly check for the existence of past values

kevinlewi:
In general, this change improves the security properties that are supposed to be attained from a limited history proof. A (full) key history proof provides a list of all past versions for a query key. A limited history proof allows for some past version to be left out of the proof. Previously, the way this was implemented was to literally truncate the older versions.

Now, we are instead properly adding the existence proofs for past marker versions and non-existence proofs for future marker versions.

Specifically in this change, it:

- Removes `HistoryParams::SinceEpoch`, since this is not really supportable by the existing construction. The only way for specifying a non-default parameter now is with `HistoryParams::MostRecent`.
- Parameters in the `HistoryProof` struct were repurposed and renamed to support past and future marker versions
- Added a new `get_marker_versions()` utility function which determines the past and future version numbers to check as part of the history proof generation and verification
- Moved `HistoryParams` out from `akd` and into `akd_core` since it is also used by verification
- Added a new `InvalidVersion` error type
- Added tests for the new history proof verification behavior, and updated docs as well

I am also bumping the version to 0.12.0-pre.1, since the introduced changes are incompatible with the previous version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Installation
 Add the following line to the dependencies of your `Cargo.toml`:
 
 ```
-akd = "0.11"
+akd = "0.12.0-pre.1"
 ```
 
 ### Minimum Supported Rust Version
--- a/akd/Cargo.toml
+++ b/akd/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "akd"
-version = "0.11.0"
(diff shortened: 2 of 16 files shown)


kevinlewi:
Subsumed by #433, closing this one.

PR #425, opened 2024-03-07
Implemented AsyncInMemoryDatabase::set in terms of ::batch_set

cryo28:
The implementation of AsyncInMemoryDatabase::set and batch_set contain a lot of duplicate code.  

Tested with:

```
cargo test test_in_memory_db
```

### DIFF
--- a/akd/src/storage/memory.rs
+++ b/akd/src/storage/memory.rs
@@ -79,23 +79,8 @@ impl AsyncInMemoryDatabase {
 #[async_trait]
 impl Database for AsyncInMemoryDatabase {
     async fn set(&self, record: DbRecord) -> Result<(), StorageError> {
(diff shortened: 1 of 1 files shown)


kevinlewi (approved):
Awesome, thanks and LGTM. However, seems like there is a clippy error that is on a line which is unrelated to your changes, but if you can fix it so that CI tests pass, that would be great!

https://github.com/facebook/akd/actions/runs/8195091568/job/22412621676?pr=425#step:4:556

kevinlewi:
Sorry, I landed #426 first which also fixed the clippy error -- can you rebase on main and force-push again? Should fix the CI issue

cryo28:
It seems like force updates don't re-trigger workflows. Not sure what I should do? Should I recreate the PR or add some dummy commit to the existing one?

PR #426, opened 2024-03-15
Small fixes to the READMEs and CLI help messages

haochenuw:
In this PR I
- updated the README in the example directory to include the MySQL docker instance spin-up command.
- updated the help message for the CLI in mysql_demo. Since it is no longer supported to get root_hash of an older epoch, I modified the description of root_hash command to remove the optional epoch parameter from the CLI help message.


kevinlewi:
Looks good, thanks! I know it's unrelated, but mind addressing the clippy error in this diff as well?

kevinlewi (approved):
LGTM

PR #427, opened 2024-03-16
Implemented AsyncInMemoryDatabase::set in terms of ::batch_set

cryo28:
The implementation of AsyncInMemoryDatabase::set and batch_set contain a lot of duplicate code.

Tested with:

```
cargo test test_in_memory_db
```


kevinlewi (approved):
Great, thanks!

PR #428, opened 2024-03-21
Add documentation around HistoryParams insecure options

kevinlewi:
The non-complete history options offered in the HistoryParams struct are insecure and should not be used for production purposes. This change adds these caveats to the docs.


slawlor (approved):
Lgrm

dillonrg (approved):

Hemmikuhschlemur (commented):
What would be a main purpose of this.

PR #429, opened 2024-03-26
Update tombstone API to minimize DB ops

afterdusk:
## Current
Our current tombstoning API `tombstone_value_states` accepts a list of `ValueState` keys that tombstoning should be performed for. This means that the client flow will, more often than not, involve:
1. looking up all `ValueState` objects for a given user in DB
2. figuring out which objects should be tombstoned based on their epochs (relative to the current epoch)
3. passing the keys of said objects to the tombstone API

Internally, the tombstone API then does another lookup of `ValueState` objects with the keys passed, modifies the objects to "tombstone" them, then another write to commit the now-tombstoned objects.

## Proposed
We can cut out the DB lookup pre-API call by having the API take in an epoch and a username, then tombstoning all `ValueState` objects for the user with an epoch less than or equal to the passed epoch. Clients will not need any knowledge of what `ValueState` objects a user has, and just need to inform the API which epoch to stop tombstoning at.

In addition, before we write the tombstoned records, we should filter out any objects that are already tombstoned, to avoid unnecessary writes.

### DIFF
--- a/akd/src/storage/manager/mod.rs
+++ b/akd/src/storage/manager/mod.rs
@@ -14,7 +14,6 @@ use crate::storage::transaction::Transaction;
 use crate::storage::types::DbRecord;
 use crate::storage::types::KeyData;
 use crate::storage::types::ValueState;
-use crate::storage::types::ValueStateKey;
 use crate::storage::Database;
 use crate::storage::DbSetState;
 use crate::storage::Storable;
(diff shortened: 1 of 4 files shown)


slawlor (approved):
Lgtm

akd/src/storage/manager/mod.rs
```diff
@@ -421,17 +420,16 @@ impl<Db: Database> StorageManager<Db> {
...
-
-        let data = self.batch_get::<ValueState>(keys).await?;
+    /// Tombstones all value states for a given AkdLabel, up to and including a given epoch
+    pub async fn tombstone_value_states(
+        &self,
+        username: &AkdLabel,
+        epoch: u64,
+    ) -> Result<(), StorageError> {
+        let key_data = self.get_user_data(username).await?;
         let mut new_data = vec![];
```
dillonrg, line 430: Nit: I think you can effectively write this a bit cleaner (without a mutable variable) by doing something like so:

```rust
    let new_data = key_data
        .states
        .into_iter()
        .filter_map(|vs| {
            if vs.epoch <= epoch && vs.value.0 != crate::TOMBSTONE {
                DbRecord::ValueState(ValueState {
                    epoch: value_state.epoch,
                    label: value_state.label,
                    value: crate::AkdValue(crate::TOMBSTONE.to_vec()),
                    username: value_state.username,
                    version: value_state.version,
                })
            } else {
                None
            }
        })
        .collect();
```

dillonrg (approved):

akd/src/storage/manager/mod.rs
afterdusk: Ah, the rest of this file uses for loops and mutable collections, I'll keep it the current way for consistency for now. Thanks for the suggestion!

PR #432, opened 2024-04-04
Prevent Any Storage Errors from Clobbering aZKS Entry

dillonrg:


dillonrg:
## Current
Prior to this patch, AKD would incorrectly assume that any error during initialization of
a `Directory` should result in an `aZKS` being created and stored in the underlying database
implementation. The problem here is that this behavior happened on every error, even those which
do not definitively indicate that the read for an `aZKS` passed but a record was not found.

For example, if a connection issue occurs when connecting to the storage layer, the current implementation
could wrongly overwrite an existing `aZKS`. In such a scenario, the original read for the `aZKS` fails
during `Directory` initialization, but a subsequent write succeeds. This can result in clobbered records
and unexpected behavior.

## Proposed/New
With this change, we update the initialization logic for the `Directory` type to only create an `aZKS` if
the returned `Err` indicates that the read was successful, but an `aZKS` was not found. In all other instances,
we propagate the `Err` back to the caller instead of swallowing it and wrongly creating an `aZKS`.

akd/src/directory.rs
```diff
@@ -65,11 +65,12 @@ where
...
-            // generate a new azks if one is not found
-            let azks = Azks::new::<TC, _>(&storage).await?;
-            // store it
-            storage.set(DbRecord::Azks(azks.clone())).await?;
+        if let Err(AkdError::Storage(StorageError::NotFound(_))) = azks {
+            // generate + store a new azks only if one is not found
+            let new_azks = Azks::new::<TC, _>(&storage).await?;
+            storage.set(DbRecord::Azks(new_azks.clone())).await?;
+        } else {
+            let _res = azks?;
```
kevinlewi, line 73: Maybe we can just do a single match statement on `azks?`:
```
match azks? {
  Ok(_) => Ok(Directory { ... }),
  Err(StorageError) => ... ,
  Err(e) => Err(e) ,
}
```

Thoughts?

dillonrg: Works for me! We mostly care about the `StorageError::NotFound` occurrence here, so felt an if-let was a bit more succinct. That said, I don't have any strong reservations around using a match.

afterdusk (approved):

akd/src/directory.rs
dillonrg: For posterity -- @kevinlewi  and I spoke offline and felt that the if-let approach was a bit cleaner after some more discussion.

akd/src/tests.rs
```diff
@@ -130,6 +131,39 @@ fn setup_mocked_db(db: &mut MockLocalDatabase, test_db: &AsyncInMemoryDatabase)
...
+        ..Default::default()
+    };
+    mock_db
+        .expect_get::<Azks>()
+        .returning(|_| Err(StorageError::Connection("Fire!".to_string())));
+    let storage = StorageManager::new_no_cache(mock_db);
+
+    let maybe_akd = Directory::<TC, _, _>::new(storage, vrf.clone()).await;
+    assert!(maybe_akd.is_err());
+
```
eozturk1, line 152: Can we also check here that `storage.get` doesn't return any data for `AZKS` and `root` `StorageKey`s to ensure they haven't been inserted into the database?

akd/src/tests.rs
```diff
@@ -130,6 +131,39 @@ fn setup_mocked_db(db: &mut MockLocalDatabase, test_db: &AsyncInMemoryDatabase)
...
+    let mut mock_db = MockLocalDatabase {
+        ..Default::default()
+    };
+    let test_db = AsyncInMemoryDatabase::new();
+    setup_mocked_db(&mut mock_db, &test_db);
+    let storage = StorageManager::new_no_cache(mock_db);
+
+    let maybe_akd = Directory::<TC, _, _>::new(storage, vrf).await;
+    assert!(maybe_akd.is_ok());
+
```
eozturk1, line 163: Similar to above, can we check the AZKS with epoch 0 and root with initial value has been inserted into the database when we successfully created the directory?

akd/src/directory.rs
```diff
@@ -65,11 +65,14 @@ where
     pub async fn new(storage: StorageManager<S>, vrf: V) -> Result<Self, AkdError> {
         let azks = Directory::<TC, S, V>::get_azks_from_storage(&storage, false).await;
 
-        if azks.is_err() {
-            // generate a new azks if one is not found
-            let azks = Azks::new::<TC, _>(&storage).await?;
-            // store it
-            storage.set(DbRecord::Azks(azks.clone())).await?;
+        if let Err(AkdError::Storage(StorageError::NotFound(_))) = azks {
+            // generate + store a new azks only if one is not found
```
eozturk1, line 69: Can we log here that AZKS not found (and the error string), and we are creating a new one along with a root?

eozturk1 (approved):

akd/src/tests.rs
dillonrg: The mock is configured to return an `Err` here, so I don't really think that's possible in this case.

akd/src/tests.rs
dillonrg: Yes, we can check here!

akd/src/directory.rs
dillonrg: Sure! I am not sure the error string will have much value, but it doesn't hurt to log it!

akd/src/directory.rs
```diff
@@ -65,11 +65,14 @@ where
...
+            let new_azks = Azks::new::<TC, _>(&storage).await?;
+            storage.set(DbRecord::Azks(new_azks.clone())).await?;
```
dillonrg, line 71: I don't think we need to clone the new azks here, as it'll be dropped once we're out of scope anyways and it isn't being used anywhere else.

akd/src/tests.rs
eozturk1: Can we confirm mock database's set function is not called perhaps? (If not I'm fine as is.)

dillonrg: @eozturk1 good call! Added.

akd/src/tests.rs
```diff
@@ -130,6 +131,44 @@ fn setup_mocked_db(db: &mut MockLocalDatabase, test_db: &AsyncInMemoryDatabase)
...
+    let akd = maybe_akd.expect("Failed to get create a Directory!");
+    let azks = akd.retrieve_azks().await.expect("Failed to get aZKS!");
+    assert_eq!(0, azks.get_latest_epoch());
```
eozturk1, line 167: Shall we check initial root is there too?

akd/src/tests.rs
```diff
@@ -130,6 +131,44 @@ fn setup_mocked_db(db: &mut MockLocalDatabase, test_db: &AsyncInMemoryDatabase)
...
+    let vrf = HardCodedAkdVRF {};
+
+    // Verify that a Storage error results in an error when attempting to create the Directory
+    let mut mock_db = MockLocalDatabase {
+        ..Default::default()
+    };
+    mock_db
+        .expect_get::<Azks>()
+        .returning(|_| Err(StorageError::Connection("Fire!".to_string())));
+    mock_db.expect_set().times(0);
```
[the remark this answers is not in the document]
dillonrg: Oh, no, it's on the `mock_db` itself. So we want to do that before we pass it to `StorageManager`. If you change the times call to 1, for example, this fails.

kevinlewi (approved):

PR #433, opened 2024-05-02
Adding support for HistoryProofV2 

haochenuw:
The main change in this PR is the introduction of HistoryProofV2 which improves security properties for limited history proofs. It is based upon https://github.com/facebook/akd/pull/422, with changes to maintain backward compatibility and support unit tests for both versions of history proof. 

- Added HistoryProofV2 struct and key_history_v2 and key_history_verify_v2 functions which generates and verifies HistoryProofV2. 
- Extended HistoryVerificationParams to include information about the HistoryParams, to be used during the proof verification. 
- Removes HistoryParams::SinceEpoch. The only way for specifying a non-default parameter now is with HistoryParams::MostRecent.
- Added a new get_marker_versions() utility function which determines the past and future version numbers to check as part of the HistoryProofV2 generation and verification
- Moved HistoryParams out from akd and into akd_core since it is also used by verification
- Added a new InvalidVersion error type
- Added tests for HistoryProofV2
- updated docs
- Bumping the version to 0.12.0-pre.1


haochenuw:
@dillonrg -- I saw that Kevin has renamed in PR428 the MostRecent to MostRecentInsecure in HistoryParams. Since I am reusing MostRecent for history V2 as well that created a merge conflict. I have thought there are two options 

1/ Rename it back to MostRecent. This is the easiest and since in this PR explicitly mark V1 of HistoryProof as deprecate, it feels like risk of misuse is small. 

2/ Leave HistoryParams untouched and add a new struct HistoryParamsV2 (similarly, HistoryVerificationParamsV2). This makes the message clearer but the unit tests macros that I have added in this PR will need to be modified to pass in those structs as parameters. 

Let me know what you think and I can make the changes! 

cc @kevinlewi for FYI.

haochenuw:
cc @eozturk1 @afterdusk for your thoughts on the choices above.

kevinlewi:
@haochenuw Thanks for working on this!

I think, rather than having a key_history() and key_history_v2() interface, it might be better to instead just have one key_history() function, but add an additional parameter that allows the caller to essentially specify the version. Then, the return of the function can be an enum, something like:
```
pub enum KeyHistoryResult {
  V1(V1Struct),
  V2(V2Struct),
}
```

which can be parsed by key_history_verify() as well. That way, we also avoid having to throw everything into macros for the tests that test both versions.

Let me know what you think about the suggested change... I am hoping it will lead to a smaller change overall and simpler code.

Also, no need to push each commit to the PR, you can just force-push to erase the history and have a single commit change for the PR.

README.md
```diff
@@ -56,8 +56,9 @@ Alex Chernyak ([@alexme22](https://github.com/alexme22)),
 Dillon George ([@dillonrg](https://github.com/dillonrg)),
 Sean Lawlor ([@slawlor](https://github.com/slawlor)),
 Kevin Lewi ([@kevinlewi](https://github.com/kevinlewi)),
-Jasleen Malvai ([@jasleen1](https://github.com/jasleen1)), and
-Ercan Ozturk ([@eozturk1](https://github.com/eozturk1)).
+Jasleen Malvai ([@jasleen1](https://github.com/jasleen1)),
+Ercan Ozturk ([@eozturk1](https://github.com/eozturk1)), and
+Hao Chen ([@haochenuw](https://github.com/haochenuw)).
```
kevinlewi, line 61: Let's revert this for now, this is the list of "original authors" and we can update this section to give you attribution at some point later, but at the very least in a separate PR.

akd/benches/directory.rs
```diff
@@ -79,7 +79,7 @@ fn history_generation<TC: NamedConfiguration>(c: &mut Criterion) {
 
                 // generate for the most recent 10 updates
                 let label = AkdLabel::from("User 1");
-                let params = akd::HistoryParams::MostRecentInsecure(5);
+                let params = akd::HistoryParams::MostRecent(5);
```
kevinlewi, line 82: I think we should actually keep the name with the "Insecure" tag. Adding deprecation tags doesn't necessarily imply that the functionality is insecure, as this could be confusing to callers.

akd/src/append_only_zks.rs
```diff
@@ -340,6 +340,7 @@ impl Azks {
     /// before it is written to storage. The is_new flag indicates whether the
     /// returned node is new or not.
     #[async_recursion]
+    #[allow(clippy::multiple_bound_locations)]
```
kevinlewi, line 343: What is this for?

akd_core/src/build.rs
```diff
@@ -18,11 +18,12 @@ const PROTOBUF_FILES: [&str; 1] = ["types"];
 const PROTOS_OUTPUT_DIR: &str = "protos";
 
 fn build_protobufs() {
+    println!("calling build protobufs...");
     let mut protobuf_files = Vec::with_capacity(PROTOBUF_FILES.len());
 
     for file in PROTOBUF_FILES.iter() {
         let proto_file = format!("{PROTOBUF_BASE_DIRECTORY}/{file}.proto");
-        println!("cargo:rerun-if-changed={proto_file}");
+        // println!("cargo:rerun-if-changed={proto_file}");
```
kevinlewi, line 26: Don't forget to revert this as well

kevinlewi (changes requested):

akd/src/directory.rs
```diff
@@ -580,6 +572,145 @@ where
...
+            return Err(AkdError::Directory(DirectoryError::InvalidVersion(
+                "Computed start version for the key history should be non-zero".to_string(),
+            )));
+        }
+
+        let (past_marker_versions, future_marker_versions) =
+            get_marker_versions(start_version, end_version, current_epoch);
+
+        let mut past_marker_vrf_proofs = Vec::<Vec<u8>>::new();
+        let mut existence_of_past_marker_proofs = Vec::<MembershipProof>::new();
```
dillonrg, line 656: Nit: Any reason we're using the turbofish notation here? Not correctly inferred without it?

akd_core/src/build.rs
```diff
@@ -18,11 +18,12 @@ const PROTOBUF_FILES: [&str; 1] = ["types"];
 const PROTOS_OUTPUT_DIR: &str = "protos";
 
 fn build_protobufs() {
+    println!("calling build protobufs...");
```
dillonrg, line 21: Probably another spot to revert :)

akd_core/src/utils.rs
```diff
@@ -16,6 +16,47 @@ pub fn get_marker_version_log2(version: u64) -> u64 {
...
+/// future marker versions is as described in SEEMless: the consecutively increasing set of versions
+/// from end_version until the next power of 2, and then all consecutive powers of 2 up until the
+/// epoch.
+///
+/// This will panic if start_version = 0
+pub fn get_marker_versions(
+    start_version: u64,
+    end_version: u64,
+    epoch: u64,
+) -> (Vec<u64>, Vec<u64>) {
```
dillonrg, line 40: Nit: I don't know if a tuple of `Vec` results feels super intuitive to me. I think I'd prefer named fields here to make it a bit more apparent to readers what these represent.

akd_core/src/verify/history.rs
```diff
@@ -193,6 +222,208 @@ pub fn key_history_verify<TC: Configuration>(
...
+    // Make sure the update proofs are non-empty
+    if num_proofs == 0 {
+        return Err(VerificationError::HistoryProof(format!(
+            "No update proofs included in the proof of user {akd_label:?} at epoch {current_epoch:?}!"
+        )));
+    }
+
+    // Check that the sent proofs are for a contiguous sequence of decreasing versions
+    for count in 0..num_proofs {
+        if count > 0 {
```
dillonrg, line 242: Can't we start looping from index 1 and avoid the if-check here?

akd_core/src/verify/history.rs
```diff
@@ -193,6 +222,208 @@ pub fn key_history_verify<TC: Configuration>(
...
+            if proof.update_proofs[count].version + 1 != proof.update_proofs[count - 1].version {
+                return Err(VerificationError::HistoryProof(format!(
+                    "Update proofs should be ordered consecutively and in decreasing order.
+                    Error detected with version {} = {}, followed by version {} = {}",
+                    count,
+                    proof.update_proofs[count].version,
+                    count - 1,
+                    proof.update_proofs[count - 1].version
+                )));
+            }
```
dillonrg, line 253: Since we're using the versions in a few spots (if-expression check and logging), I think we can just move these to variables to clean this up a bit:

```rust
            let prev_version = proof.update_proofs[count - 1].version;
            let curr_version = proof.update_proofs[count].version;

            if curr_version + 1 != prev_version {
                return Err(VerificationError::HistoryProof(format!(
                    "Update proofs should be ordered consecutively and in decreasing order.
                    Error detected with version {count} = {curr_version}, followed by version {} = {prev_version}",
                    count - 1,
                )));
            }
```

akd_core/src/verify/history.rs
```diff
@@ -193,6 +222,208 @@ pub fn key_history_verify<TC: Configuration>(
...
+            if start_version != 1 {
+                return Err(VerificationError::HistoryProof(format!(
+                    "Expected start version to be 1 given that it is a complete history, but got start_version = {}",
+                    start_version
+                )));
+            }
+        }
+        HistoryParams::MostRecent(recency) =>
+        {
+            #[allow(clippy::comparison_chain)]
```
dillonrg, line 292: Since this match block mostly seems to be aimed at surfacing errors which violate invariants, I think you can add all of these if-checks into arms in the match with corresponding guards to return an `Err`. That may help with the linter warning here, as I think you'd be forced to handle the "ok" case.

akd/src/append_only_zks.rs
dillonrg: Hmm, looks like cases where a [generic is defined in a bound predicate and a `where` clause](https://rust-lang.github.io/rust-clippy/master/index.html#/multiple_bound_locations). That said, I don't see that in this signature.

@haochenuw is this from a macro or something?

haochenuw: Yes, this is to address a clippy warning. It seems to show up only for the functions with #[async_recursion].

akd/src/directory.rs
haochenuw: not sure if I follow -- what is the alternative that you suggest?

akd/src/utils.rs
```diff
@@ -51,3 +51,24 @@ macro_rules! test_config {
         }
     };
 }
+
+/// Macro used for running tests with different configurations
+#[cfg(any(test, feature = "public_tests"))]
+#[macro_export]
+macro_rules! versioned_test_config {
```
kevinlewi, line 58: This is not used anywhere and can be deleted, right?

akd_core/src/verify/history.rs
haochenuw: Sounds good! 

on the error message, how about this "Error detected with version {curr_version} at index {count}, followed by version {prev_version} at index {count-1}"?

akd_core/src/utils.rs
haochenuw: sounds good!

akd/src/utils.rs
haochenuw: yes, will delete

akd/src/directory.rs
kevinlewi: I think Dillon's question is, can we simplify these lines to:

```
        let mut past_marker_vrf_proofs = vec![];
        let mut existence_of_past_marker_proofs = vec![];
```
and not explicitly specify the types, since presumably they are inferred

README.md
haochenuw: sounds good, reverted!

akd_core/src/verify/history.rs
haochenuw: I updated the message. Let me know what you think!

examples/src/fixture_generator/reader/mod.rs
```diff
@@ -15,12 +15,15 @@ use crate::fixture_generator::generator::{Delta, Metadata, State};
 /// Interface for reading output generated by the tool.
 pub trait Reader {
     /// Reads a metadata object.
+    #[allow(dead_code)]
```
kevinlewi, line 18: Is this actually dead code? Can we try removing this and seeing if clippy complains (or sorry if you already tried that).

Same for all of the instances below.

examples/src/mysql_demo/mysql_storables.rs
```diff
@@ -38,6 +38,7 @@ pub(crate) trait MySqlStorable {
 
     fn set_batch_params(items: &[DbRecord]) -> Result<mysql_async::Params>;
 
+    #[allow(dead_code)]
```
kevinlewi, line 41: Also this one

docker-compose.yml
```diff
@@ -6,7 +6,7 @@ services:
     container_name: akd-test-db
     platform: linux/x86_64
     image: mysql
-    command: --default-authentication-plugin=mysql_native_password
+    command: --mysql-native-password=ON
```
kevinlewi, line 9: Hmm, what's the reason for this change? Just curious

akd_core/src/utils.rs
```diff
@@ -10,12 +10,59 @@
...
 
+/// a list of past marker versions used by history proofs
+pub type PastMarkerVersions = Vec<u64>;
+
+/// a list of future marker versions used by history proofs
+pub type FutureMarkerVersions = Vec<u64>;
+
 /// Retrieve log_2 of the marker version, referring to the exponent
 /// of the largest power of two that is at most the input version
 pub fn get_marker_version_log2(version: u64) -> u64 {
```
kevinlewi, line 21: Can we add the following comment to this function:

```
///
/// Note: This will panic if called on version = 0
```

kevinlewi (approved):
Fantastic! Some nits in comments but overall LGTM!

akd/src/append_only_zks.rs
```diff
@@ -880,6 +881,7 @@ impl Azks {
 
     #[async_recursion]
     #[allow(clippy::type_complexity)]
+    #[allow(clippy::multiple_bound_locations)]
```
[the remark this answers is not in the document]
haochenuw: yes, this is added because clippy complains. I think there may have been an update to clippy so that this warning did not show up before this PR.

examples/src/fixture_generator/reader/mod.rs
haochenuw: yes, this is added because clippy complains. I think there may have been an update to clippy so that this warning did not show up before this PR.

docker-compose.yml
haochenuw: This is due to MySQL 8.4 docker image made a breaking change in the way it takes the command line parameters. Without this, the CI for mysql test will timeout due to hanging around this error 
```
ERROR 2002 (HY000): Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)
```

More context is https://stackoverflow.com/questions/78445419/unknown-variable-default-authentication-plugin-mysql-native-password

examples/src/mysql_demo/mysql_storables.rs
haochenuw: same as above. This is added specifically to address warnings that showed up during CI.

PR #435, opened 2024-06-04
Only instantiate storage metrics when runtime_metrics is enabled

afterdusk:
Only instantiate metrics attribute in storage manager when the `runtime_metrics` feature is enabled. This avoids an unnecessary import of atomic types, which not all architectures support.


eozturk1 (approved):
LGTM!

PR #436, opened 2024-06-20
[akd] add marker proof nodes to preload caching for v2 lookup proofs

artem-shu:
lookup v2 performs marker proof for past markers (membership proofs) and future parkers (non-membership proofs). To keep the number of DB reads low, we currently rely on `preload_nodes` to BFS down from the root to all nodes who share a prefix for target labels and their immediate children. This PR *should* include the marker proofs (past/future versions of the same akd_label and freshness) in the BFS traversal to take greater advantage of existing caching.

ATM, there do no appear to be specific tests for cache-traversal inclusion. Rust is not my forte, so please review carefully.

### DIFF
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -623,9 +623,10 @@ impl Azks {
         &self,
         storage: &StorageManager<S>,
         lookup_infos: &[LookupInfo],
+        marker_labels: Option<Vec<NodeLabel>>,
     ) -> Result<u64, AkdError> {
         // Collect lookup labels needed and convert them into Nodes for preloading.
-        let lookup_nodes: Vec<AzksElement> = lookup_infos
+        let mut lookup_nodes: Vec<AzksElement> = lookup_infos
             .iter()
             .flat_map(|li| vec![li.existent_label, li.marker_label, li.non_existent_label])
             .map(|l| AzksElement {
@@ -633,6 +634,13 @@ impl Azks {
                 value: AzksValue(EMPTY_DIGEST),
             })
             .collect();
+
+        if let Some(labels) = marker_labels {
+            lookup_nodes.extend(labels.iter().map(|l| AzksElement {
+                label: *l,
+                value: AzksValue(EMPTY_DIGEST),
+            }))
+        }
 
         // Load nodes.
         self.preload_nodes(storage, &AzksElementSet::from(lookup_nodes))
--- a/akd/src/directory.rs
+++ b/akd/src/directory.rs
@@ -298,7 +298,7 @@ where
             #[cfg(not(feature = "greedy_lookup_preload"))]
             {
                 current_azks
-                    .preload_lookup_nodes(&self.storage, &vec![lookup_info.clone()])
+                    .preload_lookup_nodes(&self.storage, &vec![lookup_info.clone()], None)
                     .await?;
             }
         }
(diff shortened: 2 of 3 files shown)


kevinlewi (commented):
> ATM, there do no appear to be specific tests for cache-traversal inclusion. Rust is not my forte, so please review carefully.

Can we add such a test for this diff specifically? And then we can hopefully verify before/after the change that it fails/succeeds.

akd/src/tests.rs
```diff
@@ -1467,6 +1460,29 @@ async fn test_key_history_verify_malformed<TC: Configuration>() -> Result<(), Ak
     Ok(())
 }
 
+// Test that BFS traversal captures both children of a prefix label for the sake of non-membership proofs
```
artem-shu, line 1463: temp changes
```
clear && cargo test test_preload_nodes
```

akd/src/tests.rs
```diff
@@ -110,10 +110,6 @@ fn setup_mocked_db(db: &mut MockLocalDatabase, test_db: &AsyncInMemoryDatabase)
             futures::executor::block_on(tmp_db.batch_get::<TreeNodeWithPreviousValue>(key))
         });
 
-    let tmp_db = test_db.clone();
-    db.expect_batch_get::<Azks>()
-        .returning(move |key| futures::executor::block_on(tmp_db.batch_get::<Azks>(key)));
-
```
kevinlewi, line 116: Hmm, any reason for removing this?

akd/src/append_only_zks.rs
```diff
@@ -634,6 +635,13 @@ impl Azks {
             })
             .collect();
 
+        if let Some(labels) = marker_labels {
+            lookup_nodes.extend(labels.iter().map(|l| AzksElement {
+                label: *l,
+                value: AzksValue(EMPTY_DIGEST),
+            }))
+        }
+
```
kevinlewi, line 644: nit: I feel like there should be a way to consolidate this so that we don't repeat the ```
.map(|l| AzksElement {
                label: *l,
                value: AzksValue(EMPTY_DIGEST),
            }))
            ```
 segment of code twice

akd/src/tests.rs
artem-shu: Screenshot 2024-06-21 at 11 27 24 AM
It appears to be a dupe/typo, and from my local test-runs has no bearing on test outcomes

akd/src/append_only_zks.rs
artem-shu: I had issues with references when I tried to extend the `lookup_infos` iterator after the flat-map. Perhaps there is a valid syntax to correctly reference them

kevinlewi: Let's create an issue out of this so that we remember to come back to it and also add a comment in the code

artem-shu: https://github.com/facebook/akd/issues/437

akd/src/directory.rs
```diff
@@ -479,27 +479,14 @@ where
...
-                if let Ok(lo) = self.build_lookup_info(ud).await {
-                    lookup_infos.push(lo);
-                }
-            }
-            current_azks
-                .preload_lookup_nodes(&self.storage, &lookup_infos)
-                .await?;
-        }
-
         let mut update_proofs = Vec::<UpdateProof>::new();
```
artem-shu, line 482: I think I created a bug here.

PR #439, opened 2024-06-25
Reorganizing test files

kevinlewi:
No functional changes, just moves all of the existing tests out of the 1500-line file `akd/src/tests.rs` into a new folder, split under:
- `akd/src/tests/mod.rs`
- `akd/src/tests/test_core_protocol.rs`
- `akd/src/tests/test_errors.rs`
- `akd/src/tests/test_preloads.rs`

### DIFF
--- a/akd/src/tests/mod.rs
+++ b/akd/src/tests/mod.rs
@@ -1,0 +1,126 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+//
+// This source code is dual-licensed under either the MIT license found in the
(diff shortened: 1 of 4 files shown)


dillonrg (approved):
Thanks for putting this together, @kevinlewi ! It makes parsing the tests a lot more friendly for readers. 

Not a blocker, but do you think it makes sense to drop the `test_` prefix on the individual files since they're already in the `test` directory/module? We'd then have something like `tests/core_protocol.rs`. Feels slightly redundant as is, but I'm fine with keeping it if you prefer it that way!

PR #441, opened 2024-06-26
[akd] fix prefetch ordering bug

artem-shu:
My original diff added a bug by moving the pre-fetch down to the code-block where version-range was in scope. This means that the update proof was run before the pre-load, resulting in extra fetches.

https://github.com/facebook/akd/pull/436

This PR should address that my moving the pre-fetch block back to its original position, and moving the range-selection to a separate earlier block.

From initial canary, the latency appears to be improved:

Screenshot 2024-06-26 at 4 01 21 PM

### DIFF
--- a/akd/src/directory.rs
+++ b/akd/src/directory.rs
@@ -479,16 +479,11 @@ where
             return Err(AkdError::Storage(StorageError::NotFound(msg)));
         }
 
-        let mut update_proofs = Vec::<UpdateProof>::new();
         let mut start_version = user_data[0].version;
         let mut end_version = 0;
         for user_state in &user_data {
             // Ignore states in storage that are ahead of current directory epoch
             if user_state.epoch <= current_epoch {
-                let proof = self
-                    .create_single_update_proof(akd_label, user_state)
-                    .await?;
-                update_proofs.push(proof);
                 start_version = std::cmp::min(user_state.version, start_version);
                 end_version = std::cmp::max(user_state.version, end_version);
             }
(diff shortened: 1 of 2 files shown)


akd/src/directory.rs
```diff
@@ -479,20 +479,14 @@ where
...
-                end_version = std::cmp::max(user_state.version, end_version);
-            }
-        }
+        let versions = &user_data
+            .iter()
+            .filter(|user_state| user_state.epoch <= current_epoch)
+            .map(|user_state| user_state.version)
+            .collect::<Vec<u64>>();
+        let first_version = user_data[0].version;
+        let start_version = versions.iter().min().copied().unwrap_or(first_version);
```
artem-shu, line 488: NOTE: this is slightly different than the original code as previously, it would find the min of `[first_version] ++ min_state_version`, whereas now it just uses `first_version` if no versions were found in `min_state_version`. Maybe will be best to match to old behavior exactly

kevinlewi: Hmm yeah I would say let's just match the old behavior exactly. The min / max computation won't be expensive, and  rather than using the .iter().filter() syntax I think it's fine to just have the explicit loop as before.

artem-shu: updated to chaining the elements to a list so now the logic should be in parity. I can switch back to an explicit loop if preferred.

artem-shu:
somehow latency is back to 4 seconds.

akd/src/directory.rs
kevinlewi: Yeah let's just do the explicit loop still... the presence of unwrap()s makes me uneasy anyway.

akd/src/directory.rs
```diff
@@ -529,6 +524,17 @@ where
                 .await?;
         }
 
+        let mut update_proofs = Vec::<UpdateProof>::new();
```
kevinlewi, line 527: Hmm, perhaps to avoid a future attempt to optimize this, can we add a comment to this codeblock saying that it should only happen after the preloading is finished?

```
// The creation of update proofs should happen only after the preload operation (to prevent cache misses).
```

kevinlewi (approved):
LGTM! One nit about a comment

akd/src/directory.rs
artem-shu: added comment

PR #443, opened 2024-07-08
Anchor mysql version to 8.4

afterdusk:
In #433, we updated our `docker-compose.yaml` to use a [startup argument](https://github.com/facebook/akd/pull/433/files#r1610960175) relevant to the then latest mysql version (8.4).

However this startup argument is deprecated in the current latest version of mysql (9.0). I have not had time to understand what changes need to be made to be compatible - for now, let's set the mysql version to 8.4 to fix CI and unblock local development. We will upgrade to a more recent version when we are ready.

I also drop the `version` attribute from the `docker-compose.yaml`, since it is [obsolete](https://docs.docker.com/compose/compose-file/04-version-and-name/#version-top-level-element-obsolete) and produces a warning when used.


eozturk1 (approved):

afterdusk:
Added one more commit to rename `.cargo/config` to `.cargo/config.toml`, to address the warning that runs on every cargo invocation:
```
warning: `akd/.cargo/config` is deprecated in favor of `config.toml`
```

PR #444, opened 2024-07-12
Adding lookup / history proof efficiency optimization

kevinlewi:
Implements an efficiency optimization that changes the contents of a key history proof (reducing the number of total past marker versions and future marker versions) without compromising on security.

From the doc comments of get_marker_versions() (which is basically the only function that is updated):

The past marker versions are determined as follows:

1. Include the largest power of 2 that is less than start_version.
2. Include the largest element of MARKER_VERSION_SKIPLIST that is less than start_version.
3. Include at most a log_2(start_version) number of versions between start_version and the
largest power of 2 less than start_version, determined as follows: For each bit position i
in start_version, if the bit is 1, include the value of start_version with the ith bit set
to 0 and followed by trailing zeros.

As a concrete example, if start_version = 85, the past marker versions would be [16, 64, 80, 84].
Since:
01010101 => 85
01010100 => 84
01010000 => 80
01000000 => 64

And 16 comes from MARKER_VERSION_SKIPLIST.

The future marker versions are determined as follows:

1. Include all powers of 2 that begin from start_version, up until the smallest element in
MARKER_VERSION_SKIPLIST that is greater than start_version.
2. Include all elements of MARKER_VERSION_SKIPLIST that are between start_version and epoch.
3. Include at most a log_2(start_version) number of versions between start_version and the
smallest power of 2 greater than start_version, determined as follows: For each bit position i
in start_version, if the bit is 0, include the value of start_version with the ith bit set
to 1 and followed by trailing zeros.

As a concrete example, if start_version = 85, the future marker versions would be
[86, 88, 96, 128, 256, 65536, 2^32] (potentially truncated depending on if any of these
numbers exceed epoch).

Since:
01010101 => 85
01010110 => 86
01011000 => 88
01100000 => 96
10000000 => 128

And the remainder of the list comes from MARKER_VERSION_SKIPLIST.

Note that the past marker versions do not contain start_version, as this would be redundant
in the history proof (since membership is already checked for start_version).

### DIFF
--- a/akd_core/src/utils.rs
+++ b/akd_core/src/utils.rs
@@ -10,6 +10,10 @@
 #[cfg(feature = "nostd")]
 use alloc::vec::Vec;
 
+/// This array is: [2, 4, 16, 256, 65536, 2^32] and is used in get_marker_versions() as
+/// an efficiency optimization
+const MARKER_VERSION_SKIPLIST: [u64; 7] = [1, 1 << 1, 1 << 2, 1 << 4, 1 << 8, 1 << 16, 1 << 32];
+
 /// a list of past marker versions used by history proofs
 pub type PastMarkerVersions = Vec<u64>;
 
@@ -19,8 +23,21 @@ pub type FutureMarkerVersions = Vec<u64>;
 /// Retrieve log_2 of the marker version, referring to the exponent
 /// of the largest power of two that is at most the input version
 /// Note: This will panic if called on version = 0
-pub fn get_marker_version_log2(version: u64) -> u64 {
+pub(crate) fn get_marker_version_log2(version: u64) -> u64 {
+    assert!(
+        version != 0,
+        "get_marker_version_log2 called with version = 0"
+    );
     64 - (version.leading_zeros() as u64) - 1
+}
+
+/// Returns the position of the first 1 in the binary representation of the input
+fn get_bit_length(input: u64) -> u64 {
+    let leading_zeros = input.leading_zeros() as u64;
+    if leading_zeros > 64 {
+        panic!("get_bit_length input has more than 64 leading zeros");
+    }
+    64 - leading_zeros
 }
 
 /// Return two (possibly empty) lists of marker versions, given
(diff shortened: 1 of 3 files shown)


akd_core/src/utils.rs
```diff
@@ -33,37 +42,140 @@ pub fn get_marker_version_log2(version: u64) -> u64 {
...
+                past_marker_versions.push(past);
+            }
+        }
     }
 
     // Compute future marker versions
+    let mut future_marker_versions: Vec<u64> = Vec::new();
+
+    let end_version_length = get_bit_length(end_version);
+    let mut future: u64 = end_version;
```
dillonrg, line 131: Very small nit pick: shall we consider renaming this to something that isn't easily mistaken for a future (in the async sense).

akd_core/src/utils.rs
```diff
@@ -33,37 +42,140 @@ pub fn get_marker_version_log2(version: u64) -> u64 {
...
 
     (past_marker_versions, future_marker_versions)
 }
 
+// Given an input u64 and a sorted array of u64s, find the largest index for which
+// the corresponding array element is less than the input.
+//
+// Note that if the input is less than the smallest element of the array, or the
+// array is of length 0, then this function will panic.
+fn find_max_index_in_array(input: u64, arr: &[u64]) -> usize {
```
dillonrg, line 168: To ensure I'm not missing anything, the arr in this case is usually a small sorted array; right? I ask because we can technically do a binary search on the arr (since it's sorted) to find the largest index. However, if the array is small, I imagine the linear search may actually be faster in practice. I don't think this matters too much in the grand scheme of things, but figured I'd mention it.

akd_core/src/utils.rs
```diff
@@ -164,32 +276,115 @@ macro_rules! test_config_sync {
...
+    }
+
+    fn gen_versions(
+        rng: &mut OsRng,
+        start_type: &RangeType,
+        end_type: &RangeType,
+        epoch_type: &RangeType,
+    ) -> (u64, u64, u64) {
+        let small_jump = 10;
+        let medium_jump = 1000;
```
dillonrg, line 324: Think it makes sense to tie the numeric values to the enum entries as part of the type, or as a conversion into an i32 or something? Might make it a bit more explicit rather than putting those details in this fn. Feel free to ignore if you prefer this!

dillonrg (approved):

akd_core/src/utils.rs
```diff
@@ -19,10 +23,15 @@ pub type FutureMarkerVersions = Vec<u64>;
 /// Retrieve log_2 of the marker version, referring to the exponent
 /// of the largest power of two that is at most the input version
 /// Note: This will panic if called on version = 0
-pub fn get_marker_version_log2(version: u64) -> u64 {
+pub(crate) fn get_marker_version_log2(version: u64) -> u64 {
```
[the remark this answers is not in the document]
kevinlewi: Good catch! Added an explicit assert! check.

akd_core/src/utils.rs
kevinlewi: Renamed to future_version (and the other variable from past -> past_version)

akd_core/src/utils.rs
kevinlewi: Good call! Hmm... so I changed this function to actually be called `find_max_index_in_skiplist()` since we only ever call it on MARKER_VERSION_SKIPLIST. And I added a comment that we could optimize this to be a binary search instead of a linear search in the future, but that at the moment it doesn't make too much of a difference since the array is only 7 elements long.

akd_core/src/utils.rs
kevinlewi: There was some annoying math / edge cases to cover with these ranges that I ended up hardcoding with these as integers, and since it's test code, I figured not to bother too much in terms of over-engineering it. I think this is ok for now.

kevinlewi:
@dillonrg thanks for the feedback!

dillonrg (approved):

akd_core/src/utils.rs
```diff
@@ -33,37 +46,145 @@ pub fn get_marker_version_log2(version: u64) -> u64 {
...
+// Given an input u64 and a sorted array of u64s, find the largest index for which
+// the corresponding array element is less than the input.
+//
+// This implementation performs a linear search over the MARKER_VERSION_SKIPLIST array,
+// but since it is sorted, it could be faster to do a binary search. However, given that
+// the array is small, there shouldn't be too much of a difference
+// between a binary search and a linear one. But if this ends up being problematic
+// in the future, it could certainly be optimized.
+//
+// Note that if the input is less than the smallest element of the array, then this
```
dillonrg, line 175: Hmm, this actually will not panic either. This is another underflow area!

kevinlewi: Ah, thanks! Fixed in the latest update by adding explicit checks that result in panics

dillonrg (approved):

dillonrg (approved):

PR #449, opened 2024-08-23
Fix Key History Dirty Read Bug

dillonrg:
**Bug**
In the `key_history` API exposed in `directory.rs`, it is possible that states read from storage are not fully committed yet (i.e. the transaction has not been committed), but the associated epoch in the states is higher than what is seen in the `aZKS` entry read from storage. In fact, we explicitly account for that scenario when inspecting versions across the states: https://github.com/facebook/akd/blob/main/akd/src/directory.rs#L484-L490.

Prior to this patch, we always initiated the `end_version` variable at 0. However, the `end_version` variable should **never** be 0 since it will panic in downstream API calls: https://github.com/facebook/akd/blob/main/akd_core/src/utils.rs#L182-L184. The bug in this case is simply that we default `end_version` to 0, but do not update it in the event that we have performed a dirty read for states.

To resolve the bug, we default `end_version` to the same value as `start_version`.

**Verification**
Prior to updating `directory.rs`, the added tests to repro cases where a dirty read is performed and a newer epoch is returned as part of the `KeyData` fails with the expected panic:

```
failures:

---- tests::test_errors::test_key_history_dirty_reads_experimental_config stdout ----
thread 'tests::test_errors::test_key_history_dirty_reads_experimental_config' panicked at akd_core/src/utils.rs:183:9:
find_max_index_in_skiplist called with input less than smallest element of MARKER_VERSION_SKIPLIST
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- tests::test_errors::test_key_history_dirty_reads_whatsapp_v1_config stdout ----
thread 'tests::test_errors::test_key_history_dirty_reads_whatsapp_v1_config' panicked at akd_core/src/utils.rs:183:9:
find_max_index_in_skiplist called with input less than smallest element of MARKER_VERSION_SKIPLIST

failures:
    tests::test_errors::test_key_history_dirty_reads_experimental_config
    tests::test_errors::test_key_history_dirty_reads_whatsapp_v1_config

test result: FAILED. 100 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 4.36s
```

After updating `directory.rs`, the panic no longer occurs.

### DIFF
--- a/akd/src/directory.rs
+++ b/akd/src/directory.rs
@@ -480,7 +480,7 @@ where
         }
 
         let mut start_version = user_data[0].version;
-        let mut end_version = 0;
+        let mut end_version = user_data[0].version;
         for user_state in &user_data {
             // Ignore states in storage that are ahead of current directory epoch
             if user_state.epoch <= current_epoch {
@@ -489,9 +489,10 @@ where
             }
         }
 
-        if start_version == 0 {
+        if start_version == 0 || end_version == 0 {
             return Err(AkdError::Directory(DirectoryError::InvalidVersion(
-                "Computed start version for the key history should be non-zero".to_string(),
+                "Computed start and end versions for the key history should be non-zero"
+                    .to_string(),
             )));
         }
 
--- a/akd/src/directory.rs
+++ b/akd/src/directory.rs
@@ -1034,7 +1034,7 @@ impl<TC: Configuration, S: Database + 'static, V: VRFKeyStorage> Directory<TC, S
             return Ok(EpochHash(current_epoch, root_hash));
         }
 
-        if let false = self.storage.begin_transaction() {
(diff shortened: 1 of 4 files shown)


akd/src/directory.rs
```diff
@@ -489,9 +489,10 @@ where
             }
         }
 
-        if start_version == 0 {
+        if start_version == 0 || end_version == 0 {
```
dillonrg, line 492: Realistically, `end_version` should never be 0 assuming `start_version` isn't (i.e. this will short circuit). I'm opting to be a bit more explicit here to err on the side of caution given the bug we're fixing :)

kevinlewi (approved):
LGTM!

PR #453, opened 2024-10-01
Add Support for Tracing

dillonrg:
**Context**
The [`tracing`](https://docs.rs/tracing/latest/tracing/index.html) crate enables more robust instrumentation of Rust programs compared to basic logging. With this patch, we introduce the `tracing` crate into AKD and only enable it when the `tracing` feature is specified. By default, the feature is disabled and we continue to leverage basic logging.

In addition to adding `tracing` based logging and instrumentation throughout the `akd` lib (i.e. not `akd_core`), various grammatical and organizational improvements were made in areas where tracing was being added. Notably, the `log_metrics` functions which exist throughout the storage layer were updated to simply log with `info` level instead of taking an argument to specify the level. Rationale being that the `log_metrics` functions are only called when the `runtime_metrics` feature is enabled and `info` is a fair median to assume.

**Testing**
Since no major functional changes were made, the changes were tested via existing automated tests with various different feature flags being passed to toggle `tracing` on and off.

### DIFF
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -9,6 +9,7 @@
 
 use crate::hash::EMPTY_DIGEST;
 use crate::helper_structs::LookupInfo;
+use crate::log::{debug, info};
 use crate::storage::manager::StorageManager;
 use crate::storage::types::StorageType;
 use crate::tree_node::{
@@ -22,8 +23,8 @@ use crate::{
     AppendOnlyProof, AzksElement, AzksValue, Digest, Direction, MembershipProof, NodeLabel,
     NonMembershipProof, PrefixOrdering, SiblingProof, SingleAppendOnlyProof, SizeOf, ARITY,
 };
+
 use async_recursion::async_recursion;
-use log::info;
 use std::cmp::Ordering;
 #[cfg(feature = "greedy_lookup_preload")]
 use std::collections::HashSet;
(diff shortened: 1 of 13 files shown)


akd/src/append_only_zks.rs
```diff
@@ -676,7 +677,7 @@ impl Azks {
                 .collect();
         }
 
-        info!("Preload of tree ({} nodes) completed", load_count);
+        debug!("Preload of tree ({} nodes) completed", load_count);
```
dillonrg, line 680: I'm moving this to `Debug`, as it's a bit noisy considering that it's in the hot path of read oriented APIs. Under load, this may be a source of spammy log messages which offer little value in terms of information.

slawlor (approved):
Lgtm thanks for the great addition!

kevinlewi (approved):
LGTM!

kevinlewi (approved):
LGTM!

akd/src/directory.rs
```diff
@@ -673,10 +713,12 @@ where
     }
 
     /// Retrieves the [Azks]
+    #[cfg_attr(feature = "tracing_instrument", tracing::instrument(skip_all))]
     pub(crate) async fn retrieve_azks(&self) -> Result<Azks, crate::errors::AkdError> {
         Directory::<TC, S, V>::get_azks_from_storage(&self.storage, false).await
     }
 
+    #[cfg_attr(feature = "tracing_instrument", tracing::instrument(skip_all, fields(ignore_cache = ignore_cache)))]
```
slawlor, line 721: nit: `fields(ignore_cache)`

akd/src/lib.rs
```diff
@@ -499,6 +504,16 @@ pub mod helper_structs;
 pub mod storage;
 pub mod tree_node;
 
+/// Shim module to group logging-related macros more easily
+/// when switching between [log](https://docs.rs/log/latest/log/)
+/// and [tracing](https://docs.rs/tracing/latest/tracing/).
+pub mod log {
```
slawlor, line 510: nit:` pub(crate)`

slawlor (commented):
couple of tiny changes, but overall lgtm

PR #454, opened 2024-10-13
Parallelize node preloads

afterdusk:
## Overview
This PR adds parallelization for node preloading. This primarily benefits publishing, which performs node preloading prior to node insertion and audit proof generation.

During node preloading, we perform a Breadth First Search on the tree starting from the root, fetching each level of the tree in sequence. This operation's latency is lower bounded since a certain minimum number of back-to-back round trips to DB is always required. However, the amount of work performed between each DB read is not trivial, and can contribute significantly to the overall preload latency when publishing hundreds of thousands of keys.

We can improve preload latency and bring it closer to the theoretical lower bound, by parallelizing the work performed between DB fetches. In this PR, I do so by having having each step in the BFS split the subsequent work for the next level into two chunks, having each chunk be processed concurrently. The storage layer might also benefit from performing DB operations in batches that are not overly large.

Note that node preloading is used in publishing and lookups, and parallelism is always disabled for the lookup path. This is because lookups are likely performed on a machine that services client requests. We do not want a single lookup to consume all the resources on the machine and starve other client requests.

## Benchmark

Ran the azks benchmark, on trunk ([fcd665a](https://github.com/facebook/akd/commit/fcd665aa20f829cd9e06cb3d70cbe0c32ffe6b67)) and on this PR:

```
cargo bench -p akd --bench azks -F bench
```

**Batch Insertion (1000 Initial Leaves, 1000 Inserted Leaves)**

Trunk:
image

This PR:
image

**Batch Insertion (200,000 Initial Leaves, 200,000 Inserted Leaves)**

Trunk:
image

This PR:
image

On my 10-core Macbook Pro, we get a decent amount of improvement (~27%).  This improvement should be larger on higher core count machines, and at higher load levels.

### DIFF
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -30,6 +30,7 @@ use std::collections::HashSet;
 use std::convert::TryFrom;
 use std::marker::Sync;
 use std::ops::Deref;
+use std::sync::Arc;
 
 /// The default azks key
 pub const DEFAULT_AZKS_KEY: u8 = 1u8;
@@ -304,7 +305,7 @@ impl Azks {
         let azks_element_set = AzksElementSet::from(nodes);
 
         // preload the nodes that we will visit during the insertion
-        let (_, time_s) = tic_toc(self.preload_nodes(storage, &azks_element_set)).await;
+        let (_, time_s) = tic_toc(self.preload_nodes(storage, &azks_element_set, false)).await;
         if let Some(time) = time_s {
             info!("Preload of tree took {} s", time,);
         }
@@ -619,7 +620,7 @@ impl Azks {
         Ok(count)
     }
 
-    pub(crate) async fn preload_lookup_nodes<S: Database + Send + Sync>(
+    pub(crate) async fn preload_lookup_nodes<S: Database + Send + Sync + 'static>(
         &self,
         storage: &StorageManager<S>,
         lookup_infos: &[LookupInfo],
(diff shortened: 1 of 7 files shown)


akd/src/append_only_zks.rs
```diff
@@ -637,47 +638,132 @@ impl Azks {
             })
             .collect();
 
-        // Load nodes.
-        self.preload_nodes(storage, &AzksElementSet::from(lookup_nodes))
+        // Load nodes without parallelism, since multiple lookups could be
+        // happening and parallelism might consume too many resources.
+        self.preload_nodes(storage, &AzksElementSet::from(lookup_nodes), true)
```
slawlor, line 643: nit: Shoulda-been-an-enum :p

akd/src/append_only_zks.rs
```diff
@@ -637,47 +638,132 @@ impl Azks {
...
+            let right_next_nodes = next_nodes.split_off(next_nodes.len() / 2);
+            let left_next_nodes = next_nodes;
+            let child_parallel_levels =
+                parallel_levels.and_then(|x| if x <= 1 { None } else { Some(x - 1) });
+
+            // Handle the left chunk in a different tokio task.
+            let storage_clone = storage.clone();
+            let azks_element_set_clone = azks_element_set.clone();
+            let left_future = async move {
+                Azks::recursive_preload_nodes(
```
slawlor, line 728: nit: 

```lang=rust
let left_future = move {
  Azks::recursive_preload_nodes(...)
};
```

I think... Pretty sure you don't need the await and wrapping it in another async block since the method returns a Future anyways.

slawlor (changes requested):
nifty! But I think we could do with making the bool arg an enum for safer usage (As I'm guessing @dillonrg would say 😛 )

akd/src/append_only_zks.rs
afterdusk: Ah we do need it, since `tokio::task::spawn` expects a future, and a closure without `async` is not a future (and we do need a closure to move the values).

akd/src/append_only_zks.rs
afterdusk: We talked about this, and personally I'm inclined to introduce the indirection if the use of the param is slightly more complex than what we have here (either having multiple variants or being used externally). But given that I've heard this feedback twice, I'll adopt it :)

akd/src/append_only_zks.rs
dillonrg: @afterdusk does something like this work:

```rust
let left_future = {
    let storage = storage.clone();
    let azks_element_set = azks_element_set.clone();
    move {
        Azks::recursive_preload_nodes(...)
    }
};
```

akd/src/lib.rs
```diff
@@ -459,7 +459,7 @@
 //!
 //! Performance optimizations:
 //! - `parallel_vrf`: Enables the VRF computations to be run in parallel
-//! - `parallel_insert`: Enables nodes to be inserted via multiple threads during a publish operation
+//! - `parallel_azks`: Enables nodes to be fetched and inserted via multiple threads during a publish operation
```
dillonrg, line 462: Nit: Can we maintain alphabetical ordering here? 😅

dillonrg (approved):

akd/src/append_only_zks.rs
afterdusk: Shadowing does seem like the cleaner way to do things, but we do use this pattern quite a bit in AKD. Let's clean this pattern up for the entire repo in a separate PR.

dillonrg (approved):

kevinlewi (approved):
LGTM!

slawlor (approved):

PR #457, opened 2024-11-20
Allow Azks parallelism to be configurable

afterdusk:
This PR is stacked on #456. It refactors Azks parallelism in the following ways:
- Simplifies parallelism logic in `get_append_only_proof_helper` 
- Removes the `parallel_azks` feature in favor of explicit configuration (see next point), to avoid interactions between multiple configuration methods.
- Creates an `AzksParallelismConfig` struct, allowing users to set parallelism for different operation types (`insertion` and `preload`). Each operation can have one of the parallelism settings:
  - No parallelism
  - Static parallelism
  - Dynamic parallelism (derived from core count)

### DIFF
--- a/akd/src/append_only_zks.rs
+++ b/akd/src/append_only_zks.rs
@@ -1101,58 +1101,33 @@ impl Azks {
             let maybe_task: Option<
                 tokio::task::JoinHandle<Result<(Vec<AzksElement>, Vec<AzksElement>), AkdError>>,
             > = if let Some(left_child) = node.left_child {
-                #[cfg(feature = "parallel_azks")]
-                {
-                    if parallel_levels.map(|p| p as u64 > level).unwrap_or(false) {
-                        // we can parallelise further!
-                        let storage_clone = storage.clone();
-                        let tsk: tokio::task::JoinHandle<Result<_, AkdError>> =
-                            tokio::spawn(async move {
-                                let my_storage = storage_clone;
-                                let child_node = TreeNode::get_from_storage(
-                                    &my_storage,
-                                    &NodeKey(left_child),
-                                    latest_epoch,
-                                )
-                                .await?;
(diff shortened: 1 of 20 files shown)


akd/src/append_only_zks.rs
```diff
@@ -243,6 +200,77 @@ impl AzksElementSet {
...
+            preload: AzksParallelismOption::Dynamic(Self::DEFAULT_FALLBACK_PARALLELISM),
+        }
+    }
+}
+
+/// Parallelism setting for a given field in [AzksParallelismConfig].
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash, PartialOrd, Ord)]
+pub enum AzksParallelismOption {
+    /// No parallelism.
+    None,
```
dillonrg, line 239: Shall we stick with `Disabled` like we used to? Just to avoid reusing `None` given its use in `Option`.  No super strong feelings on my part here, so please feel free to choose what you prefer.

examples/src/mysql_demo/mod.rs
```diff
@@ -144,9 +145,10 @@ pub(crate) async fn render_cli(args: CliArgs) -> Result<()> {
     if cli.memory_db {
         let db = akd::storage::memory::AsyncInMemoryDatabase::new();
         let storage_manager = StorageManager::new_no_cache(db);
-        let mut directory = Directory::<TC, _, _>::new(storage_manager, vrf)
-            .await
-            .unwrap();
+        let mut directory =
+            Directory::<TC, _, _>::new(storage_manager, vrf, AzksParallelismConfig::default())
+                .await
+                .unwrap();
```
dillonrg, line 151: We don't need to do this now, but I think some `expects` with an appropriate message might be better here (since it's a demo). Same goes for all uses of `unwrap` here.

PR #460, opened 2025-01-19
Adding individual epoch auditing in kt auditor example

kevinlewi:
This adds the ability for the whatsapp-kt-auditor example to audit a specific epoch, rather than loading all proofs first and then picking the epoch (which can be much slower)

### DIFF
--- a/examples/src/whatsapp_kt_auditor/mod.rs
+++ b/examples/src/whatsapp_kt_auditor/mod.rs
@@ -11,7 +11,7 @@ mod auditor;
 
 use akd::local_auditing::AuditBlobName;
 use anyhow::{anyhow, bail, Result};
-use clap::Parser;
+use clap::{Parser, Subcommand};
 use dialoguer::theme::ColorfulTheme;
 use dialoguer::{Input, Select};
 use indicatif::{ProgressBar, ProgressStyle};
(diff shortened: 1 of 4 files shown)


examples/src/whatsapp_kt_auditor/mod.rs
```diff
@@ -68,82 +80,97 @@ struct CliOption {
...
+                                    }
+                                })
+                                .interact_text()?;
+                            let epoch = epoch_input.parse::<u64>()?;
+                            let maybe_proof = proofs.iter().find(|proof| proof.name.epoch == epoch);
+                            if let Some(epoch_summary) = maybe_proof {
+                                do_epoch_audit(epoch_summary).await?;
                             } else {
-                                Err("Epoch is out of available range")
+                                bail!("Could not find epoch {}", epoch);
```
dillonrg, line 140: We can use let-else here to clean this up a bit:

```rust
let Some(epoch_summary) = maybe_proof else {
    bail!("Could not find epoch {epoch}");
};

do_epoch_audit(epoch_summary).await?
```

examples/src/whatsapp_kt_auditor/mod.rs
```diff
@@ -68,82 +80,97 @@ struct CliOption {
...
+
 pub(crate) async fn do_epoch_audit(epoch_summary: &EpochSummary) -> Result<()> {
     let pb1 = start_progress_bar("Downloading proof...");
     let proof = auditor::get_proof(WHATSAPP_KT_DOMAIN, epoch_summary).await?;
     finish_progress_bar(
         pb1,
         format!(
-            "Successfully downloaded proof for epoch {}.",
-            epoch_summary.name.epoch
+            "Successfully downloaded proof for epoch {}. ({})",
```
dillonrg, line 171: Super small nit: let's add a uom on the string we're printing to denote bytes.

dillonrg (approved):

PR #461, opened 2025-05-12
Removed redundant curr_node.get_child_node call in get_child_azks_element_in_dir

cryo28:
This code is duplicate and is not required

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # Changelog
 
-## 0.12.0-pre.12 (May 12, 2025)
+## 0.12.0-pre.12 (May 13, 2025)
 * Addressed clippy warnings and errors
+* Removed redundant function call in append_only_zks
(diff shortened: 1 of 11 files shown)


kevinlewi (approved):
Thanks, this looks good!

I wonder also if it is time to release a 0.12.0 (instead of the pre-releases).

Do you mind modifying the appropriate lines + consolidating the changelog? Or alternatively I would be happy to do this in a follow-up diff, but if you can undo your edits that mark the 0.12.0-pre.11 -> 0.12.0-pre.12 changes, that would be great!

cryo28:
I have another minor change pending at this moment. Dillon was going to canary it on today and see the impact. So, let's wait for the release a little bit more.

dillonrg (approved):

kevinlewi (approved):
LGTM! We can update the version in a separate PR

PR #470, opened 2025-07-22
Auditing improvements

kevinlewi:
(Depends on #460, which should be merged first)

The goal of this change is to improve the overall efficiency of the auditor function (`verify_consecutive_append_only()`) which is used to verify the validity of an append-only proof between two epochs. Note that this should be completely backwards-compatible with the previous way of verifying (and generating) audit proofs -- no actual correctness logic has changed.

As an example of the improvements, here is a before and after of the result of attempting to audit epoch 714064, a particularly large proof of size 291MB. The below examples are run on my laptop, with the previous performance taking 2 minutes and 9GB of RAM to audit, and the new performance taking 21 seconds and only 760MB to audit.

### Old behavior:

```
$ cargo run -p examples --release -- whatsapp-kt-auditor -e 714064
[00:00:17] Successfully downloaded proof for epoch 714064. (291.6 MB)
[00:02:01] Audit proof for epoch 714064 has verified successfully!
```
```
$ top
...
45321  akd-examples 280.4     02:19.56 11/3   1     55     **8959M+** 0B     7146M+ 45321 1201
...
```

### New behavior:

```
$ cargo run -p examples --release -- whatsapp-kt-auditor -e 714064
[00:00:18] Successfully downloaded proof for epoch 714064. (291.6 MB)
[00:00:21] Audit proof for epoch 714064 has verified successfully!
```
```
$ top
...
44251  akd-examples 297.8     00:38.32 11/3   1     56     **760M+**  0B     397M-  44251 1201
...
```

## How it works

The way auditing works is by creating two new Azks instances from scratch, building each one up by inserting nodes (the first one from the proof's unchanged nodes and the second one from the proof's unchanged + inserted nodes), and then checking that the root hash of each Azks tree matches the expected root hash (the start hash or the end hash). Note that during this process, we are building the tree just for the purpose of computing the root hash, and nothing else -- the trees are discarded afterwards.

In the previous behavior, we would use an `AsyncInMemoryDatabase` to host all of the storage for the nodes that we insert into the tree (along with the resulting intermediary tree nodes). This means that the Azks instance would keep all of the nodes that have been inserted into the tree in memory at all times, up until the final root hash is computed.

In the new behavior, we still use an `AsyncInMemoryDatabase`, but we enable a flag which more aggressively removes nodes that no longer need to be kept in memory in order to compute the root hash. In particular, whenever we attempt to add a parent node to storage, we remove the left and right children from storage, if they exist. This takes advantage of the fact that batch insertion works on a level-by-level basis, computing the children hashes before computing their parent hash. So, once a parent hash has been computed, we no longer need the intermediary computations (corresponding to the child hashes).

### DIFF
--- a/akd/src/auditor.rs
+++ b/akd/src/auditor.rs
@@ -55,6 +55,11 @@ pub async fn audit_verify<TC: Configuration>(
 }
 
 /// Helper for audit, verifies an append-only proof.
+///
+/// This function first creates a new AZKS instance with the unchanged nodes from the proof,
+/// then it verifies the start hash against the root hash of this AZKS instance.
+/// Next, it creates another AZKS instance with the unchanged nodes and inserted nodes,
+/// and verifies the end hash against the root hash of this second AZKS instance.
 #[cfg_attr(feature = "tracing_instrument", tracing::instrument(skip_all))]
 pub async fn verify_consecutive_append_only<TC: Configuration>(
     proof: &SingleAppendOnlyProof,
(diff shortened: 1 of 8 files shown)


akd/src/auditor.rs
```diff
@@ -55,47 +55,70 @@ pub async fn audit_verify<TC: Configuration>(
...
+    let computed_start_root_hash: Digest = azks1.get_root_hash::<TC, _>(&manager1).await?;
+    if computed_start_root_hash != start_hash {
+        return Err(AkdError::AzksErr(AzksError::VerifyAppendOnlyProof(
+            format!(
+                "Start hash {} does not match computed root hash {}",
+                hex::encode(start_hash),
+                hex::encode(computed_start_root_hash)
+            ),
+        )));
+    }
```
dillonrg, line 91: It seems like we're mostly doing the same thing in the `StorageManager` instances we're creating here (i.e. d `manager1` and `manager2`). That is: 

1. Creating the per-level cache + an akzs
2. Inserting some set of nodes into the azks
3. Asserting the resultant root hash is equal to an expected root hash

With the above in mind, do you think it might make sense to define a helper-like function which captures the common? E.g.

```rust
async fn verify_append_only_hash<TC:Configuration>(
    nodes: Vec<AzksElement>,
    expected_hash: Digest,
    latest_epoch: Option<u64>,
) -> Result<(), AkdError> {
    let manager = StorageManager::new_no_cache(
        AsyncInMemoryDatabase::new_with_remove_child_nodes_on_insertion(),
    );
    let mut azks = Azks::new::<TC, _>(&manager1).await?;
    if let Some(epoch) = latest_epoch {
        azks.latest_epoch = epoch;
    }
    azks
        .batch_insert_nodes::<TC, _>(
            &manager,
            nodes,
            InsertMode::Auditor,
            AzksParallelismConfig::default(),
        )
        .await?;
    let computed_root_hash: Digest = azks.get_root_hash::<TC, _>(&manager).await?;
    if computed_root_hash != expected_hash {
        return Err(AkdError::AzksErr(AzksError::VerifyAppendOnlyProof(
            format!(
                "Expected hash {} does not match computed root hash {}",
                hex::encode(expected_hash),
                hex::encode(computed_root_hash)
            ),
        )));
    }
}
```

If we do something like that, then I think this function essentially becomes:

```rust
pub async fn verify_consecutive_append_only<TC: Configuration>(
    proof: &SingleAppendOnlyProof,
    start_hash: Digest,
    end_hash: Digest,
    end_epoch: u64,
) -> Result<(), AkdError> {
    let _unchanged = verify_append_only_hash::<TC>(
        proof.unchanged_nodes.clone(),
        start_hash,
        None,
    ).await?;
    let unchanged_with_inserted_nodes = proof.unchanged_nodes.clone().extend(proof.inserted.iter().map(|x| {
        let mut y = *x;
        y.value = AzksValue(TC::hash_leaf_with_commitment(x.value, end_epoch).0);
        y
    }));
    let _changed = verify_append_only_hash::<TC>(
        unchanged_with_inserted_nodes,
        end_hash,
        Some(end_epoch - 1),
    ).await
}
```

Note: I didn't run, nor did I format, any of the code above. It's just meant to reflect an idea to reduce some duplication, but please feel free to ignore if you prefer what's here.

akd/src/storage/memory.rs
```diff
@@ -30,15 +31,42 @@ type UserValueMap = HashMap<Epoch, ValueState>;
...
+    /// (and aggressively) remove entries corresponding to left and right
+    /// children of a tree node when the node is inserted. The purpose behind this
+    /// is to reduce the size of the in-memory database by culling the child enries
+    /// once the parent node's hash has been calculated. The primary use case for this
+    /// is to improve auditing memory usage and time (since during auditing, we no longer
+    /// care about the child node hashes once its parent has been computed). Note that this
+    /// technique takes advantage of the way batch insertion of nodes into the tree works,
+    /// since we always process all of the children of a particular subtree before processing
+    /// the root of that subtree.
+    remove_child_nodes_on_insertion: bool,
```
dillonrg, line 44: I'm generally not the biggest fan of using a boolean to differentiate behavior, as I'd like to use something like a different type to reflect that the in-memory store we're using doesn't store everything, but I think that's a bit more of a rework than what we have now. 

As such, I think what we have here is sufficient since we need to influence how inner workings of `batch_set` and using something like a newtype isn't necessarily going to make that easy given that we're not calling something before or after existing functionality. Additionally, you've commented this really well so it's pretty clear + the associated function to instantiate is super clear 👍

PR #474, opened 2025-08-26
Fix Clippy Issues

dillongeorge:
As the title states, this fixes some clippy issues which were impacting CI.

For example: https://github.com/facebook/akd/actions/runs/16900068262/job/47877320987.


slawlor (approved):
lint all the things!

PR #489, opened 2026-05-30
Add epoch-range storage query: get_user_data_in_range

### LINKED ISSUE #165: Storage contract for values between epoch range
With adding the `limited_key_history` call in `directory.rs` we now need the ability to select a series of user values given a minimum epoch (`start_epoch`). This should be a new call on the storage contract which doesn't exist today.

For usage, see `directory.rs:L366-368`.

albonec:
## Summary
- Adds `get_user_data_in_range(username, start_epoch, end_epoch)` to the `Database` trait
- Implements it in `AsyncInMemoryDatabase`, `Transaction`, `StorageManager`, and `AsyncMySqlDatabase`
- MySQL implementation pushes the range filter to the query layer to avoid over-fetching

Closes #165

### DIFF
--- a/akd/src/storage/manager/mod.rs
+++ b/akd/src/storage/manager/mod.rs
@@ -540,6 +540,58 @@ impl<Db: Database> StorageManager<Db> {
         }
     }
 
+    /// Retrieve all value states for a given user within the epoch range [start_epoch, end_epoch].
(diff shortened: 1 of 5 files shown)


albonec:
@slawlor Pretty late to this issue but saw it was open so giving it a shot! Please do respond when you can!

